PR #2, opened 2022-02-18
Fix defaults handling in modules

robin-tribe29:
Defining defaults in modules does not need fancy if statements, you can just do this in the arguments spec.


lgetwan (commented):
This implicates, that the user also has to hand over the folder name, even if he wants to change the attributes, only, for example. Otherwise, the host will be moved to the main folder.
`folder=dict(type="str", default='/'),`

PR #18, opened 2022-03-18
Fixed the galaxy import errors.

lgetwan:


lgetwan:
The two lines reported in the bug could simply be deleted.

robin-tribe29 (approved):

PR #21, opened 2022-03-22
consistent use of base_url for API calls

Edgxxar:
the modules `activation.py` and `discovery.py` don't put a slash (`/`) at the end of the `server_url` for API calls. This is inconsistent with the `host.py` and `folder.py` modules and requires the user to specify the server URL with a trailing slash in one case, and without a trailing slash in the other. This PR fixes the issue.

Maybe it would be better to create a utility class that is responsible for creating the API calls and handling response codes, as this seems very similar between all the modules. But I'm not familiar with the recommended structure for ansible collections which is why I've not done this here.


robin-tribe29:
Thanks for your contribution and suggestion! We will look into it and update here as soon as we have something.

lgetwan (approved):
Good point. Thanks for putting some more consistency to the code!

PR #23, opened 2022-03-22
Update task names to be more descriptive in the example playbooks

diademiemi:
This PR changes the name of a few tasks in the example playbooks to be more descriptive.
In my opinion the current names don't really describe what the playbook is doing, making understanding this collection more complicated.
In particular the title "Run activation module" doesn't describe at all what it's doing, so I've changed it here to "Activate changes on site". 
The discovery example just has a wrong name, the name says "Discover hosts" when that's not what it's doing, it's discovering services on an existing host.
I also changed the "Create hosts" to "create host" since it's just adding a single host here, not multiple.

### DIFF
--- a/playbooks/demo.yml
+++ b/playbooks/demo.yml
@@ -4,7 +4,7 @@
   vars_files:
     - ./vars/config.yml
   tasks:
-  - name: "Run activation module - Showcase no changes."
+  - name: "Activate changes on site - Showcase no changes."
     activation:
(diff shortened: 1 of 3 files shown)


robin-tribe29 (approved):

robin-tribe29:
Thanks for the contribution!
Names are just names, but I understand, why you made the changes and I feel like it can help people get a better idea of what is going on.

PR #38, opened 2022-04-29
Improved the error messages for non-OK HTTP returncodes.

lgetwan:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

See Bug #34 

## What is the new behavior?

- In case of an API error, not only the HTTP return code is printed out, but also the details of the response


robin-tribe29 (commented):
Looking good!

PR #55, opened 2022-06-02
Feature: Role Server

robin-tribe29:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

None.

## What is the new behavior?

A role installing Checkmk on servers and managing sites is introduced.


lgetwan:
Great work, Robin!
I found only two small cosmetic issues:
1. Importing the GPG key is always running into "changed".
2. The Site password is always updated. I know it's hard to check that beforehand, but probably, we should only set the password, if the key "admin_pw" is set for a particular site.

robin-tribe29:
Thanks!
Regarding your points:
1. This can probably be checked somehow in advance, I am just unsure if this is worth the effort.
2. This is close to impossible to check without a lot of effort. Additionally, I intended this as a feature, that the password will be managed completely, so you always know what the password will be.

But I will take another look at both topics before merging. :+1:

robin-tribe29:
I decided to not change the behaviors as I see more advantages than disadvantages here.

Issue #51, opened 2022-05-30
[BUG] Discovery got same problem with body variable


elwood218:
Verify first that your issue is not already reported [here](https://github.com/tribe29/ansible-collection-tribe29.checkmk/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc).
Where possible also test if the latest release and main branch are affected too.
Complete all sections as described!

**Describe the bug**

There are 2 errors but one is related to the module and the other is imho a problem of the API.
The error with the module is again that "body" is used but not defined. (If I put that in a loop over "play_hosts" then it works)
The general problem is that first hosts get "service discovered" and then there is a 500 error or so..

```
TASK [cmk_host_registration : Add/update/remove host] ****************************************************************
changed: [Host1 -> localhost]
changed: [Host2 -> localhost]
changed: [Host3 -> localhost]
changed: [Host4 -> localhost]
changed: [Host5 -> localhost]

RUNNING HANDLER [cmk_host_registration : service discovery] **********************************************************
changed: [Host4 -> localhost]
changed: [Host1 -> localhost]
changed: [Host3 -> localhost]
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: KeyError: 'body'
[...]

PR #57, opened 2022-06-03
Fix 51 exceptionhandling

lgetwan:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: `#51`

## What is the new behavior?

- In some rare cases, the fetch_url call to the API returns -1. In that cases, an error message is also printed out


robin-tribe29:
We should merge into devel first, can you please change the PR towards that?

PR #66, opened 2022-06-30
Install xinetd when systemd <220

diademiemi:
The new Checkmk 2.1 Linux agent requires Systemd version 220 or higher to function. If systemd is not version 220 or higher, the install script will instead deploy it for use with xinetd.

This PR adds a task to ensure xinetd is installed if the Systemd version is below 220. The agent installer will register with xinetd instead and use this to run the agent service.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

If xinetd is not installed, this can result in the agent not working on older system. CentOS 7 has Systemd 219 so it is included in this.

## What is the new behavior?
Xinetd is installed and enabled on systems where the Systemd version is below 220. There is also a variable `checmmk_agent_prep_legacy` which is trrue by default. This can be set to false if you don't want the role to do this, for example if you want to use your own system for this.

### DIFF
--- a/roles/agent/defaults/main.yml
+++ b/roles/agent/defaults/main.yml
@@ -6,3 +6,4 @@ checkmk_agent_server: localhost
 checkmk_agent_site: my_site
 checkmk_agent_update: 'false'
 checkmk_agent_tls: 'false'
+checmmk_agent_prep_legacy: 'true'
--- a/roles/agent/tasks/main.yml
+++ b/roles/agent/tasks/main.yml
@@ -2,6 +2,14 @@
 - name: "Include Derivate specific Variables."
   ansible.builtin.include_vars: "{{ ansible_os_family }}.yml"
 
+- name: "Get RPM or APT package facts."
+  ansible.builtin.package_facts:
+    manager: "auto"
+
+- name: "Import Legacy agent tasks."
+  ansible.builtin.include_tasks: "legacy.yml"
+  when: ansible_facts.packages['systemd'][0]['version'] | regex_search('\d{3,}') | int < 220 and checmmk_agent_prep_legacy | bool
+
 - name: "Download Checkmk CRE Agent."
   ansible.builtin.get_url:
     url: "{{ checkmk_agent_agent.url.cre }}"
--- a/roles/agent/tasks/legacy.yml
+++ b/roles/agent/tasks/legacy.yml
@@ -1,0 +1,11 @@
+---
+- name: "Install xinetd"
+  ansible.builtin.package:
+    name: xinetd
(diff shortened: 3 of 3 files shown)


roles/agent/defaults/main.yml
```diff
@@ -6,3 +6,4 @@ checkmk_agent_server: localhost
 checkmk_agent_site: my_site
 checkmk_agent_update: 'false'
 checkmk_agent_tls: 'false'
+checmmk_agent_prep_legacy: 'true'
```
robin-tribe29, line 9: There is a typo here, please fix that.
Also, the default should be `false` from my point of view, as you typically do not want a role to install stuff unexpectedly.

roles/agent/tasks/main.yml
```diff
@@ -2,6 +2,14 @@
 - name: "Include Derivate specific Variables."
   ansible.builtin.include_vars: "{{ ansible_os_family }}.yml"
 
+- name: "Get RPM or APT package facts."
+  ansible.builtin.package_facts:
+    manager: "auto"
+
+- name: "Import Legacy agent tasks."
+  ansible.builtin.include_tasks: "legacy.yml"
+  when: ansible_facts.packages['systemd'][0]['version'] | regex_search('\d{3,}') | int < 220 and checmmk_agent_prep_legacy | bool
```
robin-tribe29, line 11: Referring to my earlier comment, this task should fail, if the systemd version is unsupported and `checmmk_agent_prep_legacy` is set to `false`.

robin-tribe29 (changes requested):

roles/agent/defaults/main.yml
diademiemi: Right, my bad, fill fix. And yeah, makes sense

diademiemi:
Okay fixed the typo, idk how I didn't catch that earlier. Set it to false by default. I also made the regex for the version match 1 digits or more, for if you're somehow running systemd <100.

roles/agent/tasks/main.yml
diademiemi: Do you mean by this that it should just hard fail and prevent the playbook from preceding? I imagine if you have it set to false you'll be using your own inet or have already configured it manually, I think still doing the installation would make sense. Maybe it could output a message that further configuration might be required for that system?

robin-tribe29: Yes, you are making sense here. Setting `checkmk_agent_prep_legacy: false` will always skip the installation, so we are good I think.

robin-tribe29 (approved):

PR #68, opened 2022-07-01
Add option to add the host and discover services with the host and discovery modules

diademiemi:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The role does not automatically register the machine as a host in Checkmk. If this is not done manually before hand this will also result in the TLS step failing. This adds a feature to automatically register it, this way an entire server with added hosts can be rolled out completely non-interactively. 

## What is the new behavior?
The `tribe29.checkmk.host` and `tribe.checkmk.discovery` modules are used to add and discover the host.
The new booleans `checkmk_agent_add_host` and `checkmk_agent_discover` determine whether the role should add the host to the Checkmk server and whether to discover services and labels on this host. Both variables default to false.

The `checkmk_agent_delegate_api_calls` variable is used to determine which host the API calls will be sent from, defaults to `localhost`, but if the server isn't reachable from here, you can set another host that can.

The `checkmk_agent_host_ip` is set to the Ansible default IPv4 address, but can be overwritten manually, same goes for the `checkmk_agent_host_name` but for the FQDN.

The `checkmk_agent_folder` can be used to put this host in a given folder, if not given it will be omitted.

### DIFF
--- a/roles/agent/defaults/main.yml
+++ b/roles/agent/defaults/main.yml
@@ -4,5 +4,10 @@ checkmk_agent_edition: cre
 checkmk_agent_protocol: http
 checkmk_agent_server: localhost
 checkmk_agent_site: my_site
+checkmk_agent_add_host: 'false'
 checkmk_agent_update: 'false'
 checkmk_agent_tls: 'false'
+checkmk_agent_discover: 'false'
+checkmk_agent_delegate_api_calls: localhost
+checkmk_agent_host_ip: "{{ hostvars[inventory_hostname]['ansible_default_ipv4']['address'] }}"
+checkmk_agent_host_name: "{{ hostvars[inventory_hostname]['ansible_fqdn'] }}" 
--- a/roles/agent/tasks/main.yml
+++ b/roles/agent/tasks/main.yml
@@ -46,7 +46,7 @@
   register: checkmk_host_agent_tls_state
   when: (checkmk_agent_edition == "cee") and checkmk_agent_tls | bool
 
-- name: Discover services
+- name: "Discover services and labels on host."
   tribe29.checkmk.discovery:
     server_url: "{{ checkmk_agent_protocol }}://{{ checkmk_agent_server }}/"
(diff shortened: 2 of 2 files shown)


diademiemi:
I see the README doesn't include the "checkmk_agent_folder" variable yet, can you add that? It's not a required var and if omitted the host will be placed in the root folder

robin-tribe29:
Thanks for that catch!
The `devel` branch has all my changes, if you fork from there, you make the change yourself without risking a conflict. :v:

diademiemi:
Yes, will do!

I have a few more things we'll be adding before we bring Checkmk 2.1 into production:

* Update Checkmk site if the version in the vars differs from the current version
* Simple Role to set (group of) hosts in downtime
* Role / tasks in server role to use a list of hosts which it will add to its own site. (Same thing as the agent registering itself in that role, but also for existing hosts and SNMP). This is so we can define all hosts in code and have a disaster recovery solution to quickly roll out the server again
* Automated way to export rules as JSON with API and reimport 
		Will be trying to do this through Ansible, probably a module. If the API supports this we could create a module that can interface with the API to get these and enable or disable them, push JSON to update/create them and get the running JSON (register)

Which of these would you be interested in seeing PRd if we make them?

robin-tribe29:
That sounds very good!
Although I would only care for the downtime role (maybe just a playbook?) and the rules module I think.
Do let us know early if you start on the module, so we do not develop the module both. ;)

Regarding the host management, I definitely want to take a look at that! But I would have to see, what you do, before I think about including it here.

You want to be careful with the site upgrade stuff though, keep in mind that you could break something unintentionally.

Thanks for sharing all this! :hugs:

diademiemi:
Yes, will do!

I'll likely start on the rules module next sprint (which is next monday).

I'll make something that works for us and then I'll PR it and you can see what you think of it and how you'd like it changed and whether to accept it.

And yeah the site upgrade is potentially dangerous so I'll need to test that a lot, for minor releases it's probably fine though

PR #69, opened 2022-07-01
Add firewall tasks to agent and adjust server tasks

diademiemi:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The agent role does not adjust the firewall, The server role opens the firewall on port 80 if Firewalld is running.

## What is the new behavior?
This PR adds a boolean that makes the agent host open the firewall on RHEL hosts, but only to the Checkmk host using a rich rule. If the `checkmk_agent_server` variable is an IP and not a DNS name, it will use this. If it isn't an IP, `checkmk_agent_server_ip` will need to be set.
For the server role, I've also made it use a boolean. Instead of skipping the task if firewalld is not running.
The server role uses a list of ports instead of statically opening 80/tcp. I've also included port 8000 since agents use this to register with TLS. This being a list means it can also be used to open the Livestatus port.


diademiemi:
Resolved the merge conflicts for if you do decide to want to include this

roles/server/tasks/RedHat.yml
```diff
@@ -18,15 +18,14 @@
...
   ansible.posix.firewalld:
     service: http
     permanent: true
     immediate: true
     state: enabled
-  when: services_state.ansible_facts.services['firewalld.service']['state'] == 'running'
+  with_items:
+    - "{{ checkmk_server_ports }}"
+  when: checkmk_server_change_firewall | bool
+  ignore_errors: true
```
robin-tribe29, line 31: This feels dangerous, why is this necessary?

roles/agent/defaults/main.yml
```diff
@@ -7,6 +7,7 @@ checkmk_agent_site: my_site
 checkmk_agent_add_host: 'false'
 checkmk_agent_update: 'false'
 checkmk_agent_tls: 'false'
+checkmk_agent_change_firewall: 'false'
```
robin-tribe29, line 10: I like the approach. I think  following your arguments, it is fine to enable this by default. Someone with a firewall management can then disable this. Also, I would like to name the variable `checkmk_agent_configure_firewall`.

robin-tribe29 (changes requested):
Overall, after thinking about our arguments for some time I agree with you. Implement the requested changes, and we are good to merge. :)

roles/server/tasks/RedHat.yml
diademiemi: I did this so it wouldn't halt the playbook for the host if firewalld isn't running on the host. And I did this instead of the previous check since you're more likely to notice a fail rather than it skipping the host. But the firewall not being configured doesn't really block any other tasks, so it doesn't really need to stop. But if you think it should just stop the run after that failing and you should just know to set it to false for hosts where it's not running/failed before that seems okay too, I can remove the ignore errors if you want.

roles/agent/defaults/main.yml
diademiemi: Yes, will do

robin-tribe29: Yes, I missed that.

roles/server/tasks/RedHat.yml
robin-tribe29: Jup, I rather have a task fail once too often than the other way around.

robin-tribe29:
One more comment: Please keep an eye on the READMEs for the roles, if you introduce new variables. I added the changes to the role's READMEs in the devel branch, but I appreciate it, if you keep an eye on that too. Changelogs would be nice too, but I can take care of that for now.

diademiemi:
Yes, will do from now on! I'll commit the requested changes in a bit

diademiemi:
Made the requested changes :)
Want me to add it to the READMEs too?

robin-tribe29:
> Made the requested changes :) Want me to add it to the READMEs too?

Let me do that afterwards, as I already updated the READMEs.

robin-tribe29 (approved):

PR #88, opened 2022-07-29
Add task to update site

diademiemi:
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

This PR adds a task to update a site to a given version of Checkmk. 

It uses a small bash if statement to check if the version given is higher than the currently used version, then it uses the omd update command to update the site.
I'm using the force flag since otherwise it would create an interactive dialogue and hang.

The task will not run unless a method for conflict resolution is given in the `update_conflict_resolution` variable of the site


diademiemi:
I've tested it to upgrade between different versions in 2.1. I made the conflict resolution a required variable instead of setting a default so that it would force you to read the documentation on updating servers before using this

robin-tribe29:
Hi @diademiemi! We already discussed this feature internally and the result of that discussion was, that we want to stay careful here. The reason is, that this feature might suggest, that updating can be done without further user attention, which is not true in the end.
Let me raise this topic with our product management again and see where we can get.

diademiemi:
I see. I believe this is a safe way to do since it requires giving a variable out of a few choices for it to not skip the task, so you would have to read the documentation and know the risks

diademiemi:
Gonna work on implementing these changes right now

diademiemi:
Doing all the logic through Ansible now instead of a bash if statement, much cleaner this way, took a while to get right though.

If the site version differs from what is given in the variables. it imports a new tasks file `update-site.yml` which checks if the required variables are given and if the new version is installed.

After that it does a 5 second pause for patch updates with a notice. For major updates, it does 60 seconds.

It then creates a backup of the site and updates the site.

roles/server/tasks/update-site.yml
```diff
@@ -0,0 +1,47 @@
...
+    prompt: |
+      Proceeding with updating site {{ item.item.name }} from version {{ item.stdout }} to version {{ item.item.version }}.{{ checkmk_server_edition }}.
+      This is a major update. This can carry risks.{% if checkmk_server_backup_on_update %} A backup will be created in {{ checkmk_server_backup_dir }}. {% endif %}
+      This can take a while! The site will be down during the update
+  when: item.item.version | regex_replace('p.*', '') != item.stdout | regex_replace('p.*', '')
+
+- name: "Create backup of site."
+  become: true
+  ansible.builtin.shell: |
+      omd backup {{ item.item.name }} {{ checkmk_server_backup_dir }}/{{ item.item.name }}-backup-$(date +%Y%m%d-%H%M%S).zip
```
robin-tribe29, line 35: We should do this default backup without history and graphs. This is just for disaster recovery. Users should have proper backups in advance.

roles/server/defaults/main.yml
```diff
@@ -20,3 +20,6 @@ checkmk_server_sites: []
 #   state: started
 #   admin_pw: test
 checkmk_server_configure_firewall: 'true'
+
+checkmk_server_backup_on_update: 'true' # Not recommended to disable this option
+checkmk_server_backup_dir: /opt/omd/sites/
```
robin-tribe29, line 25: I do not like the path, I would rather go with `/tmp/`.
I know it might be small in size, but the backup should be too (see comment in `update-site.yml`).
That way the backup will be cleaned up upon reboot, and we do not need to worry about automated backups filling up disk space.
I am open for suggestions though.

roles/server/tasks/update-site.yml
robin-tribe29: Another option would be to carve the backup job out completely and make it fully configurable in its own file.
Then we could include it here with the settings mentioned above.

robin-tribe29 (changes requested):

roles/server/tasks/update-site.yml
diademiemi: That sounds good yeah. What files should be backed up then?

roles/server/defaults/main.yml
diademiemi: Knowing the files will b smaller I think this is preferable too. I didn't go for /tmp at first since with the size of these files that wouldn't even work in my test environment. But if the files are smaller, I think that's fine

roles/server/tasks/update-site.yml
robin-tribe29: From the help:
```
Possible options for this command:
 --no-rrds            do not copy RRD files (performance data)
 --no-logs            do not copy the monitoring history and log files
 -N,--no-past         do not copy RRD files, the monitoring history and log files
 --no-compression      do not compress tar archive
```
I would go with `--no-past`.

diademiemi: I see, I'll add that. Also didn't realise it was a tar file and not a zip, I'll change the filename too

roles/server/README.md
```diff
@@ -49,10 +49,21 @@ web interface to be accessible.
...
-This feature does not carry out updates though. If a site already exists,
-changing the version here does not change a thing.
-The version is only used during creation.
+If a higher version is specified for an existing site, a config update resolution method must first be given to update it.
+Valid choices include `install`, `keepold` and `abort`.
 
+    checkmk_server_sites:
+      - name: test
+        version: "{{ checkmk_server_version }}"
+        update_conflict_resolution: install
```
robin-tribe29, line 58: We should include this variable in `defaults/main.yml` but set it to an empty value. That way we are consistent, while still avoiding undesired updates.
Also, the variable should abide by the naming scheme (checkmk_server_foo).

roles/server/defaults/main.yml
```diff
@@ -20,3 +20,7 @@ checkmk_server_sites: []
 #   state: started
 #   admin_pw: test
 checkmk_server_configure_firewall: 'true'
+
+checkmk_server_backup_on_update: 'true' # Not recommended to disable this option
+checkmk_server_backup_dir: /tmp
+checkmk_server_backup_opts: '--no-past'
```
robin-tribe29, line 26: We are missing a newline at the end of the file. :upside_down_face:

roles/server/tasks/update-site.yml
```diff
@@ -0,0 +1,47 @@
+---
+- name: "Fail if update conflict resolution method is not given."
+  ansible.builtin.fail:
+    msg: |
+      variable `update_conflict_resolution` is not given. See `omd update --help` for more information.
+  when: item.item.update_conflict_resolution is undefined
```
robin-tribe29, line 6: This check fails for me, even if I set the variable. I also do not fully understand, what the `item.item` does here.

roles/server/tasks/update-site.yml
```diff
@@ -0,0 +1,47 @@
+---
```
robin-tribe29, line 1: Try to run `ansible-lint` and `yamllint` against your changes, there will be some findings.
Our CI will do this at a later stage too, and at some point, we will also run this checks against PRs.

robin-tribe29 (changes requested):

roles/server/tasks/update-site.yml
diademiemi: Yes will fix and will do from now on

roles/server/tasks/update-site.yml
diademiemi: `update_conflict_resolution` is a variable I have set per site. like this:

```yaml
checkmk_server_sites:
  - name: site
    version: "2.1.0p9"
    state: started
    admin_pw: testing
    update_conflict_resolution: install
```
If you'd prefer it to just be a variable like `checkmk_server_update_conflict_resolution` in the root of the variables I can adjust it to that.


The `item.item` is an artifact of how looping over multiple items is done with `include_tasks`. The root of the `item` contains the results of the comparison of the versions in the task from `sites.yml`, the `item.item` includes the variable it used for it, which in this case is the site definition in the variables. I don't believe there's a better way to loop through a list of tasks per item.. I wish `block:` had that functionality, for now doing an `include_tasks` is seemingly the best way to do it.

robin-tribe29: Well, that my moment of dumbness for this week. :facepalm: 
Forget I said anything, your implementation makes perfect sense.

roles/server/defaults/main.yml
diademiemi: Fixed!

roles/server/README.md
diademiemi: You said in the other comment that this implementation is good, so should this stay like this? I don't think having this as a default is possible since it's part of a list.

Also, I changed the example from install to abort, I feel like that's safer to have in a README since people will interpret it as a "default".

robin-tribe29: Yeah, as said in the other comment, I don't know where my mind was.
What you did makes perfect sense.

I like the new default. :+1:

roles/server/tasks/update-site.yml
```diff
@@ -0,0 +1,49 @@
...
+  when: item.item.version | regex_replace('p.*', '') == item.stdout | regex_replace('p.*', '')
+
+- name: "Long pause for major update."
+  ansible.builtin.pause:
+    seconds: 60
+    prompt: |
+      Proceeding with updating site {{ item.item.name }} from version {{ item.stdout }} to version {{ item.item.version }}.{{ checkmk_server_edition }}.
+      This is a major update. This can carry risks.{% if checkmk_server_backup_on_update %} A backup will be created in {{ checkmk_server_backup_dir }}. {% endif %}
+      This can take a while! The site will be down during the update
+  when: item.item.version | regex_replace('p.*', '') != item.stdout | regex_replace('p.*', '')
```
robin-tribe29, line 32: We need to think about downgrades here. While not being officially supported, they are often conducted.
I have no opinion yet, just want to leave a note, so we do not forget.

diademiemi: I don't think the playbook prevents a downgrade like this. I could add a task to compare the values and check if it's an upgrade or downgrade (Though parsing this might be a bit difficult but I think Ansible has some lookups you can use for version strings like these). Then have a boolean that defaults to false on whether to allow downgrades.

robin-tribe29: That sounds like a sound approach. :+1:

robin-tribe29:
No idea, why the sanity checks suddenly fail, but the error is not related to the changes in here.

diademiemi:
Yes I see, looks good!

roles/server/tasks/update-site.yml
diademiemi: Alright I'll work on that in a little bit

lgetwan:
Automatic updating is a risky step, but it looks like @diademiemi has built in some safety features that reduce this risk. :-)
I'll give it a go!

roles/server/tasks/update-site.yml
diademiemi: My bad this took so long, other things got in the way. Just pushed a commit that adds a boolean `checkmk_server_allow_downgrades` that defaults to false. If this is set to false there's a task that uses Ansibles `version` test to compare the versions.

diademiemi:
Any updates on this? @robin-tribe29, did you see comment on the downgrade check, does this resolve it in your opinion?
I tested some cases by comparing some fake versions with varying amounts of decimals to make sure it doesn't error (for example detect `p21` as a downgrade to `p3`) and it seems like Ansible parses this correctly.

robin-tribe29:
Sorry for the delay @diademiemi! I was meaning to merge and release this, but I was on a business trip and this week starts loaded with appointments. But I will get back to this within the week, pinky promise! :)

diademiemi:
I see, no worries!

roles/server/README.md
```diff
@@ -39,6 +39,10 @@ Cryptographically verify the downloaded setup file.
 
     checkmk_server_configure_firewall: 'true'
 
+Whether to allow downgrading a sites version.
```
robin-tribe29, line 42: We should note here, that this creates an unsupported setup for enterprise customers I think.

roles/server/README.md
```diff
@@ -49,10 +53,21 @@ web interface to be accessible.
...
+
+    checkmk_server_sites:
+      - name: test
+        version: "{{ checkmk_server_version }}"
+        update_conflict_resolution: abort
+        state: started
+        admin_pw: test
+
+Directory to backup sites to when updating between versions.
+    checkmk_server_backup_dir: /opt/omd/sites
```
robin-tribe29, line 67: I think this is out of sync with `defaults/main.yml` right?

roles/server/tasks/update-site.yml
robin-tribe29: Looks great!
I am testing this right now.

roles/server/README.md
diademiemi: Yes correct, I'll edit this

roles/server/README.md
diademiemi: I'll add a note for this!

roles/server/tasks/update-site.yml
robin-tribe29: When testing this with versions 2.1.0p9 and 2.1.0p1, the downgrade check does not seem to work.
I might be doing something here though.

diademiemi: I'll have a look and see what's up

diademiemi: Looks to be caused by the variable in the defaults being `'false`', which when not casted to a boolean will equal true. Will push a commit to fix this

diademiemi: I must've been testing in my vars with just `true` and `false`

robin-tribe29:
I just realized, that we might run into merge conflicts, or overwrite some stuff, as the firewalls task in this branch was fixed in another PR. Do you know a good way how to tackle this @diademiemi?

diademiemi:
Once the change requests are accepted GitHub should show if merging can be done automatically or if manual intervention is required. If manual intervention is required it'll open an editor when you attempt to merge which shows conflicts. It'll then show proposalsand you can choose between them or manually insert your own. It's likely it'll have a proposal which is what the devel branch has right now (The fixed version) and you can just click on that.

diademiemi:
Actually it says in the sub test merging can be performed automatically. So I think it should be fine (Which makes sense, as the modified files in that PR were not modified in this PR, and the defaults.yml is very easy for Git to automatically correct)

robin-tribe29 (approved):

PR #106, opened 2022-08-18
Update RedHat firewalld handling

mto79:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Just open port by service in firewall but doesn't do anything possible because of boolean

## What is the new behavior?

Just implement service and nog with_items
Deleted with_items directive because no {{ item }} and the subject is firewall service en not ports

### DIFF
--- a/roles/server/tasks/RedHat.yml
+++ b/roles/server/tasks/RedHat.yml
@@ -25,6 +25,4 @@
     permanent: true
     immediate: true
     state: enabled
-  with_items:
-    - "{{ checkmk_server_ports }}"
   when: checkmk_server_configure_firewall | bool


robin-tribe29:
Hi @mto79. We do appreciate every contribution, but there are two issues with this one:
1. Your pull request is poorly formatted, I can barely understand what it says.
2. If you take a look at `vars/RedHat.yml`, you will realize, that the loop actually makes sense.
If you want to rename the task to reflect, what it does, I would support that.
If you want to extend the port list, I would also support that.

But in this state, I will have to close the PR.

mto79:
Hi Robin, with a loop , you expect an item . The directive is only on the httpd service .. not to open the port.

mto79:
My apologies, it is of course the Python module that is handling the loop under the hood in the xml file of the service. So the with_items than Will make sense.. forgetting how dodgy ansible can be. So used to

- ansible.posix.firewalld:
    port: "{{ item }}"
    permanent: yes
    state: enabled
  loop: "{{ ports_to_open }}"

Sorry again my bad.

robin-tribe29:
Your solution in this PR does not make sense, but as far as I understand, you uncovered an actuall issue here.
The task does not respect the variable it loops over.
@diademiemi I think you have deeper insight into firewalld. Does the code in the last comment from @mto79 make sense, or are we missing something? Just want to make sure, we get the right grip on this.

diademiemi:
Yes, the code in this comment does make sense. The initial pull request reverts it to wht it was before I made it a loop.
I see the current code is indeed broken, if I had to guess something might have gone wrong while merging? Not quite sure how this could have happened but it does look like half the code is from when it only opened port 80 that and the other half is for when I made it a loop.

I think the best code for this would be:
```yaml
- name: "Open firewall ports."
  become: true
  ansible.posix.firewalld:
    port: "{{ item }}"
    permanent: true
    immediate: true
    state: enabled
  loop: "{{ checkmk_server_ports }}"
  when: checkmk_server_configure_firewall | bool
```

robin-tribe29:
I messed up pushing to this PR, but I pushed the change to devel: 111092bedf3ed01e5667cfec2424d63514562e8c
Closing this now. Thanks everyone!

PR #108, opened 2022-08-19
Fix ansible-lint findings.

robin-tribe29:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

ansible-lint fails with several issues.

## What is the new behavior?

- Fix indentation
- Fix FQCN
- Fix ignore_errors usage

### DIFF
--- a/.github/workflows/ansible-lint.yaml
+++ b/.github/workflows/ansible-lint.yaml
@@ -59,6 +59,9 @@ jobs:
       - name: Install ansible-lint
         run: pip install ansible-lint yamllint
 
+      - name: Install Checkmk Collection
+        run: ansible-galaxy collection install tribe29.checkmk
+
       # Run the linting
       - name: Run yamllint on roles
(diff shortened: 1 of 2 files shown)


robin-tribe29:
@diademiemi can you take a look at the removal of `ignore_errors` and maybe figure out a way to achieve whatever you tried to achieve with `changed_when` or `failed_when`?

diademiemi:
Sure, I think it can just be removed. I think it shouldn't fail unless firewalld isn't running, but if that's the case you should just disable that boolean on the host.

robin-tribe29:
Perfect. If this comes up again, we can still implement a detection for firewalld.

diademiemi:
Yes, I think I had one in the server task before it became a boolean, could probably get that from past commits if wanted.

robin-tribe29:
Yeah, I have the logic in mind, but little time. :sweat_smile: 
I think we are good for now.

PR #111, opened 2022-08-24
Adjustment Preflight Check for Redhat based systems and diff approach including RedHat specific vars 

mto79:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

1. The preflight check in the roles/server/tasks/main.yml checks for RHEL 7 and RHEL 8 with ansible_distribution.
    According to https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#ansible-facts-distribution it has to check against RedHat 7 and RedHat 8.
 
2. Also the use of spaces is not nice if you are using without quotes.
 
3. The checks with for e.g. ansible_distribution are out dated.
    image
    See also: https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html 

4. Including of vars only one time and using array variable for setting the checkmk requirements.
5. The loop has to be like below otherwise an array...
```
- name: "Install Checkmk Prerequisites."
  ansible.builtin.package:
    name: "{{ item }}"
    state: present
  loop: "{{ checkmk_server_prerequisites }}" 
```
6. Added dnf-plugin-cores and yum-plugins-core in list variable  checkmk_server_prerequisites.. for later PR suggestion when looking at powertools/epel-release

## What is the new behavior?

- Check now on RedHat-7 and RedHat-8
- Use of '-' instead of space in the concat
- Use of ansible_facts['distribution'] instead of ansible_distribution etc.
- One inclusion of vars file
- Extension of checkmk_server_prequisities var

## Other information
https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#ansible-facts-distribution
https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html

### DIFF
--- a/changelogs/fragments/agent_role.yml
+++ b/changelogs/fragments/agent_role.yml
@@ -2,6 +2,7 @@
 
 minor_changes:
   - Agent role - Host attributes can be fully customized now.
+  - Agent role - Check for agent updater and controller binaries. Skip registration if respective binary is missing.
 
 bugfixes:
   - Agent role - Support CFE properly.
--- a/roles/agent/tasks/main.yml
+++ b/roles/agent/tasks/main.yml
@@ -46,6 +46,16 @@
   delegate_to: "{{ checkmk_agent_delegate_api_calls }}"
   when: checkmk_agent_add_host | bool
 
+- name: "Check for Agent Updater Binary."
+  ansible.builtin.stat:
+    path: /usr/bin/cmk-update-agent
+  register: checkmk_agent_updater_binary
+
+- name: "Check for Agent Controller Binary."
+  ansible.builtin.stat:
+    path: /usr/bin/cmk-agent-ctl
+  register: checkmk_agent_controller_binary
+
 - name: "Register Agent for automatic Upates."
   become: true
   ansible.builtin.shell: |
@@ -55,6 +65,7 @@
   register: checkmk_agent_update_state
   when: |
     (checkmk_agent_edition == "cee" or checkmk_agent_edition == "cfe")
+    and checkmk_agent_updater_binary.stat.exists | bool
     and checkmk_agent_update | bool
 
 - name: "Register Agent for TLS."
@@ -66,6 +77,7 @@
   register: checkmk_agent_tls_state
   when: |
     (checkmk_agent_edition == "cee" or checkmk_agent_edition == "cfe")
+    and checkmk_agent_controller_binary.stat.exists | bool
     and checkmk_agent_tls | bool
 
 - name: "Discover services and labels on host."


mto79:
Should have make feature-branch for the second suggestion of including vars ...

robin-tribe29:
@mto79 thanks for the fresh PR. Can you look at the checks and fix the findings?
I will take a closer look into this later.

mto79:
> 

Thanks Robin, saw the trailing spaces indeed. Just fixed but will look closer tonight.

Also today i have been working on RedHat/Centos implementation of installation Checkmk server role further and i am in the phase of testing myself on DigitalOcean.)

Below a codefragment of what i am trying to test with RedHat en CentOS.
It is the implementation of powertools and Epel for the different distribution cases based on:

- https://docs.checkmk.com/latest/en/install_packages_redhat.html
- https://docs.fedoraproject.org/en-US/epel/#_el9

The current below code from (tasks/RedHat.yml) the role is only for CentOS 8 version i suppose. 
```
  ansible.builtin.shell: dnf config-manager --set-enabled powertools  # noqa command-instead-of-shell
  when: ansible_distribution_major_version == "8"
```

```
- name: "{{ ansible_facts.os_family|lower }} | install | enable powertools"
  ansible.builtin.command: "{{ item }}"
  loop:
    - "yum config-manager --enable rhel-7-server-optional-rpms"
    - "yum-config-manager --enable rhel-7-server-extras-rpms"
  when:
    - (ansible_facts.distribution == "CentOS" and ansible_facts.distribution_major_version == "7") or
      (ansible_facts.os_family == "RedHat" and ansible_facts.distribution_major_version == "7")

- name: "{{ ansible_facts.os_family|lower }} | install | Enable all repositories "
  community.general.rhsm_repository:
    name: "{{ item }}"
    state: enabled
  loop:
    - "rhel-7-server-optional-rpms"
    - "rhel-7-server-extras-rpms"
  when:
    - (ansible_facts.distribution == "CentOS" and ansible_facts.distribution_major_version == "7") or
      (ansible_facts.os_family == "RedHat" and ansible_facts.distribution_major_version == "7")

- name: "{{ ansible_facts.os_family|lower }} | install | enable powertools"
  ansible.builtin.command: dnf config-manager --set-enabled powertools
  when:
    - ansible_facts.distribution == "CentOS" and ansible_facts.distribution_major_version == "8"
  tags: ['checkmk-server-install']

- name: "{{ ansible_facts.os_family|lower }} | install | enable powertools"
  ansible.builtin.command: dnf config-manager --set-enabled crb
  when:
    - ansible_facts.distribution == "CentOS" and ansible_facts.distribution_major_version == "9"

- name: "{{ ansible_facts.os_family|lower }} | install | Enable all repositories "
  community.general.rhsm_repository:
    name: "codeready-builder-for-rhel-{{ ansible_facts.distribution_major_version }}-x86_64-rpms"
    state: enabled
  when:
    - ansible_facts.os_family == "RedHat" and ansible_facts.distribution_major_version in [ "8", "9" ])

- name: "{{ ansible_facts.os_family|lower }} | install | set epel-release from fedora download"
  block:
    - name: "{{ ansible_facts.os_family|lower }} | install | epel gpg key"
      ansible.builtin.rpm_key:
        key: "{{ checkmk_server_epel_gpg_key }}"
        state: present

    - name: "{{ ansible_facts.os_family|lower }} | install | epel-release"
      ansible.builtin.yum:
        name: "{{ checkmk_server_epel_url }}"
        state: present
      notify:
        - yum update cache
  when: checkmk_server_epel_dl_fedora|bool     

- name: "{{ ansible_facts.os_family|lower }} | install | epel-release "
  block:
    - name: "{{ ansible_facts.os_family|lower }} | install | epel-release"
      ansible.builtin.yum:
        name: "epel-release"
        state: present
      notify:
        - yum update cache
  when: not checkmk_server_epel_dl_fedora|bool    
```

If my findings are good i will open a PR if you are interested.

diademiemi:
How come some variables have an underscore prefixed to them, like `_checkmk_server_prerequisites`. Is this best practice when it comes to variables included from a role? (Which would be hard coded afaik and can not be overrules by passing variables to the run, correct me if I'm wrong)

roles/server/tasks/main.yml
```diff
@@ -1,46 +1,40 @@
...
 - name: "Include OS Family specific Variables."
-  ansible.builtin.include_vars: "{{ ansible_os_family }}.yml"
-
-- name: "Include RHEL Version specific Variables."
-  ansible.builtin.include_vars: "{{ ansible_os_family }}-{{ ansible_distribution_major_version }}.yml"
-  when: ansible_os_family == "RedHat"
+  ansible.builtin.include_vars: "{{ ansible_facts['os_family'] }}.yml"
 
 - name: "Install Checkmk Prerequisites."
-  become: true
```
robin-tribe29, line 17: You cannot remove this. We need higher privileges here for package installations.

roles/server/tasks/main.yml
```diff
@@ -1,46 +1,40 @@
...
 - name: "Download Checkmk Server Setup."
   ansible.builtin.get_url:
     url: "{{ checkmk_server_download_url }}"
     dest: "/tmp/{{ checkmk_server_setup_file }}"
-    mode: 0640
-    url_username: "{{ checkmk_server_download_user | default(omit) }}"
-    url_password: "{{ checkmk_server_download_pass | default(omit) }}"
+    mode: "0640"
+    url_username: "{{ checkmk_server_download_user|default(omit) }}"
+    url_password: "{{ checkmk_server_download_pass|default(omit) }}"
```
robin-tribe29, line 24: Jinja encourages spaces aroud pipes.

roles/server/tasks/main.yml
```diff
@@ -1,46 +1,40 @@
...
+  ansible.builtin.include_vars: "{{ ansible_facts['os_family'] }}.yml"
 
 - name: "Install Checkmk Prerequisites."
-  become: true
   ansible.builtin.package:
     name: "{{ item }}"
     state: present
-  loop:
-    - "{{ checkmk_server_prerequisites }}"
+  loop: "{{ checkmk_server_prerequisites }}"
```
robin-tribe29, line 16: I am uncertain, if I understand how this loop works with the new variable structure.

robin-tribe29 (changes requested):

roles/server/tasks/main.yml
mto79: Ah Will do!

roles/server/tasks/main.yml
mto79: Well know what you mean at first glance. But if you look in vars/redhat.yml You see you do not need the precedence and the inclusion of var files on red hat version Numbers. 

The prerequisit packages variabele is adjusted special for that. (Array like)

roles/server/tasks/main.yml
mto79: It is somewhat different than the with_items. When running the server role i had to change it inline. Otherwise it saw an array

diademiemi: Should be exactly the same, if yamllint and ansible-lint don't complain about this then I think it's better. The variable is a list anyway, so there's no functional difference, this just makes more sense.

mto79:
Will make changes tonight.

mto79:
> How come some variables have an underscore prefixed to them, like `_checkmk_server_prerequisites`? Is this best practice when it comes to variables included from a role? (Which would be hard coded afaik and can not be overrules by passing variables to the run, correct me if I'm wrong)

It is merely a temp variabele like locals in other languages , like in you can use with register for than debug massages.. it is merely style and to determine checkmk_server_prerequisities

roles/server/tasks/main.yml
mto79: Will put this back

mto79:
> > How come some variables have an underscore prefixed to them, like `_checkmk_server_prerequisites`? Is this best practice when it comes to variables included from a role? (Which would be hard coded afaik and can not be overrules by passing variables to the run, correct me if I'm wrong)
> 
> It is merely a temp variabele like locals in other languages , like in you can use with register for than debug massages.. it is merely style and to determine checkmk_server_prerequisities

@diademiemi found some elaboration:

Variable names can begin with an underscore. In many programming languages, variables that begin with an underscore are private. This is not true in Ansible. Variables that begin with an underscore are treated exactly the same as any other variable.

It iss frequently used as a dummy variable (a variable to receive a value from a function where the value is useless and never used).

But never use only the underscore  _ instead it must than be dummy.
https://docs.ansible.com/ansible/latest/dev_guide/testing/sanity/no-underscore-variable.html
(nice never knew that)

roles/server/tasks/main.yml
robin-tribe29: My uncertainty targets the logic behind the loop, not the syntax.

robin-tribe29:
I never cared too much for variable names prefixes with underscores in Ansible. The geerlingguy does it too, I think, but I never saw the need. I would rather stick with the current naming convention, as long as there are no good technical reasons to change it.

mto79:
> I never cared too much for variable names prefixes with underscores in Ansible. The geerlingguy does it too, I think, but I never saw the need. I would rather stick with the current naming convention, as long as there are no good technical reasons to change it.


Indeed big fan of geerlingguy :) It is as mentioned not technical .. merely style.  I will rename _checkmk_server_prerequisites to other variable.

mto79:
I will take some time for ansible linting here local before committing!

PR #114, opened 2022-08-24
[FEATURE] Add tags to tasks in roles

diademiemi:
This adds tags to all tasks in the role that aren't already controlled with a boolean.
With the exception when multiple tasks are controlled by a single boolean and one of these is not required for the functioning of the tasks grouped by this boolean (For example, importing the GPG key could be done prior to the package verification tasks).

## Pull request type
Feature addition

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

This feature request is described in https://github.com/tribe29/ansible-collection-tribe29.checkmk/issues/112


diademiemi:
One moment please, realised a potential usecase which could benefit from some more tags, will add this

diademiemi:
Added some tags to the include tasks and vars tasks.

In combination with https://github.com/tribe29/ansible-collection-tribe29.checkmk/pull/115 this could even allow adding SNMP devices in an Ansible inventory to a Checkmk site. (Since these include tasks would fail there).

I also changed the order of the `when` statement on the legacy import task, since like this it would check the systemd version first and fail if the tags were set to skip package facts. Tested this and it has the same behavior as before.

mto79:
@diademiemi maybe it is a good idea to describe in the readme.md som use case with tags ? The problem is that in case of to many tags people loose sight and are going to use them wrongly. So i would not use to many butI like very much your Download use case for example!

robin-tribe29:
I agree with @mto79 here, we should somehow outline in the README, how tags are intended to be used. Not sure how to exactly, though.
With `-` instead of `_` I have no strong opinion, and I am not aware of best practices. I tend to prefer `-`, but that is merely my personal aesthetic preference.

diademiemi:
I can add a table to the README of both roles explaining each tag and what it's intended use is.

As for `-` or `_`, I personally prefer snake case with `_`. I don't believe there's a best practice for this, and Ansible Lint and yamllint don't seem to do anything with this either. I think using underscores would make using tags from the command line a bit easier to read (`--skip-tags=download_package` compared to `--skip-tags=download-package`. But that's really getting into nit picking territories, so just let me know which you'd prefer I go with.

robin-tribe29:
Okay, how about this for the final argument on `-` vs. `_`: We name variables snake case, so we go with the dash for tags as a distinction. Sound reasonable? Then go for it. :v:

diademiemi:
Yep, I'm fine with that! I'll modify the tags and rebase my other PR that depends on this (maybe try to make it independent from this).

Should I update the README with a table?

robin-tribe29:
Great!

Yes, please do.

diademiemi:
Fixed merge conflicts in README from the agent role

roles/agent/README.md
```diff
@@ -79,13 +79,26 @@ Define the hostname which will be used to add the host to Checkmk.
...
+| ---- | ------- |
+| `download-package` | Download agent package. |
+| `install-package` | Install agent package with package manager. |
+| `install-prerequisites` | Install packages that are required for the role or agent to work. |
+| `include-os-family-vars` | Include OS family specific variables. |
+| `include-os-family-tasks` | Include OS family specific tasks. |
+| `get-package-facts` | Get package facts, used in the role. |
+| `enable-xinetd` | Enable xinetd on hosts with systemd prior to version 220. |
+
+You can use Ansible to skip tasks, or only run certain tasks by using these tags.
```
robin-tribe29, line 100: Maybe we should add a hint, that by default all tasks are run? Or do you think this should be common knowledge?

diademiemi: I think it should be common knowledge but I'll add a notice

robin-tribe29 (approved):

mto79:
Very Nice work @diademiemi !!

PR #115, opened 2022-08-24
[FEATURE] Don't hardcode host attributes

diademiemi:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The attributes of a host are hardcoded to set the IP and checkmk agent only. Having the attributes as a variable allows for more customisation when it comes to adding hosts.

## What is the new behavior?
The default behavior is the same. But now that the attributes aren't hardcoded, this can now be used to add hosts in a more customisable manner. Could also possibly used to add SNMP devices, though getting those attributes correct is a bit of a hassle.. But with this simple change making this a variable, a lot is possible.

One example of the possibilities this adds is the option to add labels when creating a host. a `pre_task` could be used to generate labels based on Ansible Facts and these can be added by defining this variable as this on the host:
```yaml
  tag_agent: 'cmk-agent'
  labels: "{{ dict_of_labels_for_this_host }}"
```
I did however notice that this would not update it if the host already exists, so this might be a bug in the module or API that it doesn't update labels if the host already exists, but that is outside the scope of this PR.

### DIFF
--- a/roles/agent/README.md
+++ b/roles/agent/README.md
@@ -79,6 +79,13 @@ Define the hostname which will be used to add the host to Checkmk.
 
 Define an IP address which will be added to the host in Checkmk. This is optional, as long as the hostname is DNS-resolvable.
 
+
+    checkmk_agent_host_attributes:
+        ipaddress: "{{ checkmk_agent_host_ip | default(omit) }}"
+        tag_agent: 'cmk-agent'
+
+Define attributes with which the host will be added to Checkmk.
+
 ## Dependencies
 
 <!-- A list of other roles hosted on Galaxy should go here, plus any details in regards to parameters that may need to be set for other roles, or variables that are used from other roles. -->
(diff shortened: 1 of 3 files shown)


robin-tribe29 (approved):
Good thought and elegant implementation, thanks!
I approve this, but will keep this on hold for the next release.
Will also test a bit if possible.

diademiemi:
Would that be 0.6.0 or later?

robin-tribe29:
Later. I want the site upgrade feature in that release.
Bug fixes are fine for that release, but with new features, I want to stay as clean and distinct as possible.

diademiemi:
Ah right okay. Could these be merged into devel after the site upgrade feature is merged into main and released? It would make it so these PRs wouldn't drift from the current devel if they are left open and cause merge conflicts.

robin-tribe29:
That's the plan, Padawan. ;)

PR #116, opened 2022-08-24
[BUGFIX] Fix legacy agent tasks not deploying agent in xinetd

diademiemi:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The legacy tasks omit `become: true` on tasks that do need it. This is fixed in this PR.

The xinetd script is also not always deployed. I was under the impression that the package will automatically run this deploy script on installation, but it doesn't always work.

For example, this is after running the role against a CentOS 7 host:
```
[root@server2 vagrant]# netstat -tulpn
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name    
tcp        0      0 0.0.0.0:22              0.0.0.0:*               LISTEN      1038/sshd           
tcp        0      0 127.0.0.1:25            0.0.0.0:*               LISTEN      1132/master         
tcp6       0      0 :::22                   :::*                    LISTEN      1038/sshd           
udp        0      0 0.0.0.0:68              0.0.0.0:*                           2780/dhclient       
udp        0      0 127.0.0.1:323           0.0.0.0:*                           645/chronyd         
udp6       0      0 ::1:323                 :::*                                645/chronyd         
```
The following files are present in `/etc/xinetd.d/`:
`chargen-dgram   chargen-stream  daytime-dgram   daytime-stream  discard-dgram   discard-stream  echo-dgram      echo-stream     tcpmux-server   time-dgram      time-stream`

There is no Checkmk agent deployed.
I'm not sure why the package doesn't always do this, I could've sworn it did this automatically. But at best it seems inconsistent, so having the legacy tasks trigger this will guarantee it's deployed.

## What is the new behavior?
A new task is added that triggers the script `/var/lib/cmk-agent/scripts/super-server/setup deploy`.
After this script is ran, the `check-mk-agent` file is added to `/etc/xinetd.d/`.

This will also now restart `xinetd`, for if it was already installed and enabled.
The agent is now listening:
```
[root@server2 vagrant]# netstat -tulpn
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name    
tcp        0      0 0.0.0.0:22              0.0.0.0:*               LISTEN      1038/sshd           
tcp        0      0 127.0.0.1:25            0.0.0.0:*               LISTEN      1132/master         
tcp6       0      0 :::22                   :::*                    LISTEN      1038/sshd           
tcp6       0      0 :::6556                 :::*                    LISTEN      6970/xinetd         
udp        0      0 0.0.0.0:68              0.0.0.0:*                           2780/dhclient       
udp        0      0 127.0.0.1:323           0.0.0.0:*                           645/chronyd         
udp6       0      0 ::1:323                 :::*                                645/chronyd         
```

This PR includes tags for this new task, which were added in https://github.com/tribe29/ansible-collection-tribe29.checkmk/pull/114, this should be put on hold until that PR is merged! 
Since those commits make the changes hard to read, this is the new `legacy.yml`:
```yaml
---
- name: "Install xinetd"
  become: true
  ansible.builtin.package:
    name: xinetd
    state: present
  tags:
    - install_package
    - install_prerequisites

- name: "Deploy xinetd agent."
  become: true
  ansible.builtin.shell: |
    /var/lib/cmk-agent/scripts/super-server/setup deploy
  args:
    creates: /etc/xinetd.d/check-mk-agent
  tags:
    - deploy_agent

- name: "Restart and enable xinetd"
  become: true
  ansible.builtin.service:
    name: xinetd
    state: restarted
    enabled: true
  tags:
    - enable_service
```

### DIFF
--- a/roles/agent/tasks/legacy.yml
+++ b/roles/agent/tasks/legacy.yml
@@ -1,10 +1,12 @@
 ---
 - name: "Install xinetd"
+  become: true
   ansible.builtin.package:
     name: xinetd
     state: present
 
 - name: "Enable xinetd"
+  become: true
   ansible.builtin.service:
     name: xinetd
     state: started


diademiemi:
Guess that means my ansible-lint is broken, mine didn't show that. I'll fix it

roles/agent/tasks/legacy.yml
```diff
@@ -1,11 +1,27 @@
...
 
-- name: "Enable xinetd"
+- name: "Deploy xinetd agent."
+  become: true
+  ansible.builtin.shell: |
+    /var/lib/cmk-agent/scripts/super-server/setup deploy
+  args:
+    creates: /etc/xinetd.d/check-mk-agent
+  tags:
+    - deploy_agent
```
robin-tribe29, line 18: Without testing, it might be expected behavior, that no xinetd configuration is set.
The Checkmk agent defaults to systemd, xinetd is the fallback.
IIRC, on CentOS 7 systemd is too old, so we configure systemd but do not enable it, because IP whitelisting will not work.
Could you please check your agent configuration and test this with an agent, that has xinetd explicitly enabled?
I would rather not have the role override bakery logic. Also, this could cause problems down the road.

diademiemi: Correct yeah CentOS 7 is not supported by the new agent. I don't believe it configures the systemd service, so you can't manually enable it either. 

I always test my PRs on a fresh install of Checkmk in Vagrant deployed with this collection, so not deploying the xinetd configuration is the default.

I'll test if the same happens if I explicitly enable xinetd in the bakery.

I do believe this is a sane default to do, since with the default configuration in Checkmk this would mean the role runs with no errors and everything looks okay, but the agent isn't actually configured.

robin-tribe29: I hear you. Still I am uncertain how to tackle this properly.

diademiemi: Just noticed I was testing this on hosts that already had it installed.. I'll add this as a handler, that should work better

diademiemi: Actually, the order of these tasks is complicated, for example.

If we run it where it's ran now, it means this file won't exist on fresh hosts.

If we run it later after installation, that means there's a chance xinetd isn't installed yet and the bakery will not respect this host.

So I'm thinking this file will need to be merged into main.yml itself. If it's a legacy host, install xinetd, then install the agent.

diademiemi: I'm not sure what's going on here, but I skipped this task and it still deployed. 

Like I said it's incredibly inconsistent whether it deploys or not. I was testing with the agent rule to fall back to xinetd, and that worked. But then I disabled the rule (redeployed the site to be sure it was gone) and it still worked.. 

Let me test this on 2.1.0p9, perhaps it was a bug?

diademiemi: In any case, the tasks here do still need `become: true` to be added. If I conclude this is a <=2.1.0p9 default and a bakery rule can fix it I think it'd be best to just remove this task

robin-tribe29: Yeah, this can be a mess at times, I do agree.

One more thing: How would we handle a host, that has the agent running via xinetd, where a user sets this role to deploy legacy? Especially with enabled automatic agent updates, there would be a continuous ping pong of configuration. Furthermire I think reimplementing detection logic for this here does not scale well too.

We want to make sure the unmodified agent works as good as possible, but I think it should stop there.
If one needs to handle e.g., CentOS 7, one could use a `post_tasks` task or similar.

diademiemi: I have no idea what's up here, but I now can't seem to replicate the issue, even on 2.1.0p9.. I think just leaving this be is best. I think CentOS 7 is one of the few OS's still supported that would be an issue here. And you could always just set a bakery rule for this.

I propose I rebase this PR to just include the `become: true` and leave it there. Just using a post_task if this doesn't deploy the agent properly I think is fine, since this has the potential to override bakery logic like you said.

diademiemi:
Force pushed to this branch, only adds the become true now.

Anyone searching for issues here that the agent isn't properly deployed should see this PR and can use the task as a post_task.
```yaml
- name: "Deploy xinetd agent."
  become: true
  ansible.builtin.shell: |
    /var/lib/cmk-agent/scripts/super-server/setup deploy
  args:
    creates: /etc/xinetd.d/check-mk-agent
  tags:
    - deploy_agent```

Issue #119, opened 2022-08-25
[BUG] CFE not properly supported in roles


robin-tribe29:
Verify first that your issue is not already reported [here](https://github.com/tribe29/ansible-collection-tribe29.checkmk/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc).
Where possible also test if the latest release and main branch are affected too.
Complete all sections as described!

**Describe the bug**

**Component Name**

`role agent`
`role server`?

**Ansible Version**

```console
$ ansible --version

```

**Checkmk Version**

```console
2.1.0p10.cfe
```

**Collection Version**

```console
$ ansible-galaxy collection list

```
**Environment**

**To Reproduce**
1. Set `checkmk_edition` to `cfe`.
2. Run the `agent` role.
3. Nothing happens, due to conditionals not respecting `cfe`.

**Expected behavior**

CFE should work.

**Actual behavior**

**Screenshots**

**Additional context**

PR #121, opened 2022-08-25
Fix cfe support

robin-tribe29:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

#119 

## What is the new behavior?

CFE is properly supported by the agent role.

### DIFF
--- a/changelogs/fragments/agent_role.yml
+++ b/changelogs/fragments/agent_role.yml
@@ -3,6 +3,9 @@
 minor_changes:
   - Agent role - Host attributes can be fully customized now.
 
+bugfixes:
+  - Agent role - Support CFE properly.
+
 # known_issues:
 #   - This release is still in development and a heavy work in progress.
 #   - Discovery module is not feature complete yet.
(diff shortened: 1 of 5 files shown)


robin-tribe29:
@diademiemi would you take a quick look?

diademiemi:
Just to be sure, CFE = Free / Enterprise trail edition?

robin-tribe29:
Affirmative.

diademiemi:
If I'm understanding correctly, the issue before was that CFE wouldn't count to run the enterprise edition tasks since those had the when statements set to only allow "cee"? But CFE is a trial of the enterprise edition so it does have those features available, so they should run. 
If that was the only issue before then I think this is a good approach to solving that. 

The integration test fails look unrelated, is that correct?

robin-tribe29:
Yes, your understanding is correct. :+1: 
I was thinking about possible better approaches from an Ansible point of view.
Checkmk related we are good here.

Yes, the integration tests are unrelated, I am already investigating.

diademiemi:
I think the logic is good, although I would suggest splitting the when statements, they are getting a bit long

For example I'd suggest 
```yml
  when: (checkmk_agent_edition == "cee" or checkmk_agent_edition == "cfe") and checkmk_agent_download_state.status != 200
```

```yml
  when: |
    (checkmk_agent_edition == "cee" or checkmk_agent_edition == "cfe")
    and checkmk_agent_download_state.status != 200
```

This makes the line shorter and easier to separate the part of the logic checking for the edition and which part is related to the download state. 
Same idea for the when statements for the TLS, updating, etc

PR #132, opened 2022-08-30
Make SSL verification a boolean

diademiemi:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

You can not use the agent role if your Checkmk server uses HTTPS and does not have a certificate trusted by all hosts. Downloading the agent package will fail, as will any tasks that interact with the API.

## What is the new behavior?
A new boolean `checkmk_agent_server_validate_certs` is added, it defaults to true. This is used in every `ansible.builtin.uri` task and decides whether SSL certificates should be trusted.
An option `validate_certs` is also added to all modules in this collection. This does not add any code besides the option and the documentation for it. 

## Other information
The `validate_certs` option in the modules is built into the `ansible.module_utils.urls` Python module. You can see the code that handles this here: https://github.com/ansible/ansible/blob/0c57734819853d410038d175afbe3f8529690fd4/lib/ansible/module_utils/urls.py#L1860

### DIFF
--- a/roles/agent/tasks/Debian.yml
+++ b/roles/agent/tasks/Debian.yml
@@ -2,6 +2,7 @@
 - name: "Debian Derivates: Download host-specific Checkmk {{ checkmk_agent_edition | upper }} Agent."
   ansible.builtin.uri:
     url: "{{ checkmk_agent_agent.url.cee }}?host_name={{ inventory_hostname }}&os_type=linux_deb&agent_type=host_name"
+    validate_certs: "{{ checkmk_agent_server_validate_certs | bool }}"
     dest: "{{ checkmk_agent_agent.file.host }}"
     method: GET
     headers:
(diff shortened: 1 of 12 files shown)


roles/agent/tasks/Debian.yml
```diff
@@ -2,6 +2,7 @@
 - name: "Debian Derivates: Download host-specific Checkmk {{ checkmk_agent_edition | upper }} Agent."
   ansible.builtin.uri:
     url: "{{ checkmk_agent_agent.url.cee }}?host_name={{ inventory_hostname }}&os_type=linux_deb&agent_type=host_name"
+    validate_certs: "{{ checkmk_agent_server_validate_certs | bool }}"
```
robin-tribe29, line 5: Should we add a default filter here too, in case the variable is undefined?
Applies to all files in the role.

diademiemi: I don't see why that's needed since the variable is set to true in the defaults/main.yml

robin-tribe29: I hear you, I just thought, if someone tampered with the defaults, that would ensure the certs are validated.
But might be overcautious.

diademiemi: There's no reason to ever tamper with the defaults file since they can be overriden. If you're tampering with the defaults, that's your problem imo

robin-tribe29:
@diademiemi if you add a changelog here too, I will merge this. :)

robin-tribe29:
You were quicker on the agent related changelog @diademiemi, so added the changelog for the modules.
Once the tests are green, I will merge.

PR #133, opened 2022-08-30
Add a variable for the server port

diademiemi:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The port 80 is always used for HTTP and port 443 is always used for HTTPS.

## What is the new behavior?
Just like my other PR https://github.com/tribe29/ansible-collection-tribe29.checkmk/pull/132, this makes the agent role more flexible in certain situations. 
This adds a variable to the agent role `checkmk_agent_port`. It defaults to 80 when `checkmk_agent_protocol` is `http`, and defaults to `443` when `checkmk_agent_protocol` is `https`. If you leave this variable unchanged, it shouldn't affect the role.

### DIFF
--- a/roles/agent/tasks/main.yml
+++ b/roles/agent/tasks/main.yml
@@ -32,7 +32,7 @@
 
 - name: "Create host on server."
   tribe29.checkmk.host:
-    server_url: "{{ checkmk_agent_protocol }}://{{ checkmk_agent_server }}/"
+    server_url: "{{ checkmk_agent_protocol }}://{{ checkmk_agent_server }}:{{ checkmk_agent_port }}/"
     site: "{{ checkmk_agent_site }}"
     automation_user: "{{ checkmk_agent_user }}"
     automation_secret: "{{ checkmk_agent_pass }}"
@@ -110,7 +110,7 @@
 
 - name: "Discover services and labels on host."
   tribe29.checkmk.discovery:
-    server_url: "{{ checkmk_agent_protocol }}://{{ checkmk_agent_server }}/"
+    server_url: "{{ checkmk_agent_protocol }}://{{ checkmk_agent_server }}:{{ checkmk_agent_port }}/"
     site: "{{ checkmk_agent_site }}"
     automation_user: "{{ checkmk_agent_user }}"
     automation_secret: "{{ checkmk_agent_pass }}"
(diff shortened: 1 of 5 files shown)


roles/agent/tasks/main.yml
```diff
@@ -32,7 +32,7 @@
 
 - name: "Create host on server."
   tribe29.checkmk.host:
-    server_url: "{{ checkmk_agent_protocol }}://{{ checkmk_agent_server }}/"
+    server_url: "{{ checkmk_agent_protocol }}://{{ checkmk_agent_server }}:{{ checkmk_agent_port }}/"
```
robin-tribe29, line 35: Should we include the fallback to port 80 here too, just in case?

roles/agent/tasks/main.yml
```diff
@@ -110,7 +110,7 @@
 
 - name: "Discover services and labels on host."
   tribe29.checkmk.discovery:
-    server_url: "{{ checkmk_agent_protocol }}://{{ checkmk_agent_server }}/"
+    server_url: "{{ checkmk_agent_protocol }}://{{ checkmk_agent_server }}:{{ checkmk_agent_port }}/"
```
robin-tribe29, line 113: Should we include the fallback to port 80 here too, just in case?

roles/agent/vars/main.yml
```diff
@@ -1,2 +1,2 @@
 ---
-checkmk_agent_site_url: "{{ checkmk_agent_protocol }}://{{ checkmk_agent_server }}/{{ checkmk_agent_site }}"
+checkmk_agent_site_url: "{{ checkmk_agent_protocol }}://{{ checkmk_agent_server }}:{{ checkmk_agent_port }}/{{ checkmk_agent_site }}"
```
robin-tribe29, line 2: Should we include the fallback to port 80 here too, just in case?

roles/agent/tasks/main.yml
diademiemi: That would break it if it was HTTPS, and I don't see why that's needed since it already has defaults

robin-tribe29: Right, I missed that while quickly thinking about it. Good point.

robin-tribe29:
Oh, one more thing @diademiemi: I always wrote the changelog for your changes. Would you start doing yourself, pretty please?
Take a look at `CONTRIBUTING`, that should get you started. If there are questions, do let me know.

diademiemi:
Yes, will do from now on, didn't realise I was supposed to do that, my bad!

robin-tribe29:
I took the liberty to add a changelog, as I want to merge this. :v:
No worries about the missing changelog, that info is a bit hidden in `CONTRIBUTING`. :slightly_smiling_face:

PR #126, opened 2022-08-26
Adjustment Preflight Check for Redhat based systems and diff approach including RedHat specific vars

mto79:
## Pull request type

Please check the type of change your PR introduces:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

1. The preflight check in the roles/server/tasks/main.yml checks for RHEL 7 and RHEL 8 with ansible_distribution.
    According to https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#ansible-facts-distribution it has to check against RedHat 7 and RedHat 8.
 
2. Also the use of spaces is not nice if you are using without quotes.
 
3. The checks with for e.g. ansible_distribution are out dated.
    image
    See also: https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html 

4. Including of vars only one time and using array variable for setting the checkmk requirements.
5. The loop has to be like below otherwise an array... must be list
```
- name: "Install Checkmk Prerequisites."
  ansible.builtin.package:
    name: "{{ item }}"
    state: present
  loop: "{{ checkmk_server_prerequisites }}" 
```
6. Added dnf-plugin-cores and yum-plugins-core in list variable  checkmk_server_prerequisites.. for later PR suggestion when looking at powertools/epel-release

## What is the new behavior?

- Check now on RedHat-7 and RedHat-8
- Use of '-' instead of space in the concat
- One inclusion of vars file
- Extension of checkmk_server_prequisities var

## Other information
https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#ansible-facts-distribution
https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html

PR #137, opened 2022-09-02
Prepare release 0.9.0

robin-tribe29:
## Pull request type

- [x] Bugfix
- [x] Feature
- [x] Build related changes

See PRs in next section.

## What is the new behavior?

- #140 
- #138
- #126 
- #146 
- #147


roles/server/defaults/main.yml
```diff
@@ -1,12 +1,12 @@
...
-  - CentOS 8
-  - RHEL 7
-  - RHEL 8
+  - Debian-11
+  - Ubuntu-20
+  - Ubuntu-22
+  - CentOS-7
+  - CentOS-8
+  - RedHat-7
+  - RedHat-8
```
robin-tribe29, line 9: @diademiemi can you tell, if replacing `RHEL` with `RedHat` will break the role for you?

roles/server/vars/RedHat.yml
```diff
@@ -1,11 +1,28 @@
...
+  RedHat-7:
+    - firewalld
+    - epel-release
+    - libsemanage-python
+    - yum-plugins-core
+  RedHat-8:
+    - firewalld
+    - epel-release
+    - python3-libsemanage
+    - dnf-plugins-core
```
robin-tribe29, line 25: @diademiemi can you check, if this works for you?

roles/server/tasks/main.yml
```diff
@@ -4,35 +4,18 @@
...
-- name: "Install Checkmk Prerequisites."
-  become: true
-  ansible.builtin.package:
-    name: "{{ item }}"
-    state: present
-  loop:
-    - "{{ checkmk_server_prerequisites }}"
-  tags:
-    - install-package
-    - install-prerequisites
```
robin-tribe29, line 29: @mto79 can you explain, why you moved the tasks to the OS playbooks? I missed this in your PR I think.

roles/server/defaults/main.yml
diademiemi: I'll spin up a RHEL VM and will try this out

diademiemi: RHEL doesn't appear anywhere, it does seem to be `RedHat`. So the `RHEL` lines can be removed

roles/server/tasks/main.yml
robin-tribe29: Actually this breaks the role. I will revert it, if no light is shed on the change @mto79.

mto79: Sorry Robin on vacation at the moment, have an alternative for this in mine codebase. This has to be reverted indeed!

roles/server/vars/RedHat.yml
diademiemi: Doesn't seem like epel-release is available in RHEL. I'll look at mto79s codebase and see how he did this differently. I'll create a PR with the changes to try to get it working on RHEL 7 and RHEL 8.

robin-tribe29: I think `epel` should be available for RHEL, just not sure if you enable it just like with CentOS or similar.

diademiemi: It should be yeah. But regardless I've heard the role breaks at others points on RHEL systems so I'll take a look at that today and PR

roles/server/tasks/main.yml
robin-tribe29: I reverted the change, feel free to have another look. Will probably merge this next week.

roles/server/vars/RedHat.yml
diademiemi: https://github.com/tribe29/ansible-collection-tribe29.checkmk/pull/146 includes necessary changes

PR #142, opened 2022-09-06
Functional workaround for creating an existing host

lupusAnimus:
This is a functional workaround for the case, that you want to create an already existing host

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

### DIFF
--- a/roles/agent/tasks/main.yml
+++ b/roles/agent/tasks/main.yml
@@ -43,8 +43,8 @@
     attributes: "{{ checkmk_agent_host_attributes }}"
   register: checkmk_agent_create_result
   failed_when: |
-    checkmk_agent_create_result.failed is true
-    and "The host is already part of the specified target folder" not in checkmk_agent_create_result.msg
+    (checkmk_agent_create_result.failed == true) and
+    ("The host is already part of the specified target folder" not in checkmk_agent_create_result.msg)
   delegate_to: "{{ checkmk_agent_delegate_api_calls }}"
   when: checkmk_agent_add_host | bool
 


robin-tribe29:
@lupusAnimus can you please explain your change? I see no improvement here, the current implementation works just fine.

lupusAnimus:
Hi Robin, as far as I know, it is a difference if you use 'is' or '==' as operator. In my case, I always got the error "template error while templating string: no test named 'true'." and after replacing 'is' with '==' it just worked

robin-tribe29:
> Hi Robin, as far as I know, it is a difference if you use 'is' or '==' as operator. In my case, I always got the error "template error while templating string: no test named 'true'." and after replacing 'is' with '==' it just worked

Interesting, which Ansible version are you using?

lupusAnimus:
Right at the moment, we are using ansible-playbook 2.9.27

robin-tribe29:
> Right at the moment, we are using ansible-playbook 2.9.27

That might explain it. Probably something changed in a more recent version.
As we support Ansible from Version 2.11 with this collection, I am compelled to close this PR.
Unless you have anything else to add here.

lupusAnimus:
Ok, this may explain it. So you can close this PR and I try to update our environment. Thanks

robin-tribe29:
As this came up internally too, I will investigate once more and see, if I can find a way to make this more resilient.

robin-tribe29:
This PR seem valid after all, so I will merge it.

robin-tribe29:
One more update: It seems like this issue does not occur, then Ansible is installed through `pip`.
@lupusAnimus can you confirm, that your Ansible was installed through the OS package manager?
That still does not explain the situation, but at least we have symptoms.

lupusAnimus:
Right, my Ansible was installed via the systems repositories on a CentOS 7 and another one on CentOS 8

PR #154, opened 2022-09-19
Feature host groups 1

msekania:
Additional plugin that manages host groups. Created in close analogy with host.py and partly folders.py. In connection with rule module one can create a monitoring groups assigned to chosen subfolder.

## Pull request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

-

## What is the new behavior?

- manage host groups in Checkmk (create/modify/delete)
- in connections with rule allows to monitor all hosts in the given subfolder

## Other information

For examle:

```yaml
- name: "Create a single host group."
  tribe29.checkmk.host_group:
    server_url: "http://localhost/"
    site: "my_site"
    automation_user: "automation"
    automation_secret: "$SECRET"
    host_group_name: "my_host_group"
    title: "My Host Group"
    state: "present"

- name: "Check_MK: create/register rules, which assign all servers in the given folder to the corresponding host group"
  tribe29.checkmk.rule:
    server_url: "http://localhost/"
    site: "my_site"
    automation_user: "automation"
    automation_secret: "$SECRET"
    ruleset: "host_groups"
    rule:
        folder: "/My/Host_Groups"
        conditions: {
            "host_labels": [],
            "host_tags": [],
            "service_labels": []
        }
        properties: {
            "disabled": false,
        }
        value_raw: "'my_host_group'"
    state: "present"
  run_once: 'true'
  delegate_to: localhost
```

PR #160, opened 2022-09-23
Feature bulk host group

msekania:
Module for bulk host group management. Intentionally created as a stand alone module and not mixed with host_group one.

Stand alone module for bulk host group management

## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Issue Number: `e.g. #1`

## What is the new behavior?

- bulk manage host groups in Checkmk by providing list of host groups to be created/modified/deleted
- with state "present", non existing host group will be created, existing host group with different title (alias) will be modified
- with state "absent", existing host groups with matched names will be deleted.


msekania:
it's really unfortunate that nested host_groups definition
```python
    module_args = dict(
        server_url=dict(type="str", required=True),
        site=dict(type="str", required=True),
        host_groups=dict(
            type="list",
            elements="dict",
            required=True,
            options=dict(
                name=dict(type="str", required=True),
                title=dict(type="str", required=False, default=""),
            ),
        ),
        state=dict(type="str", default="present", choices=["present", "absent"]),
    )
```
is not accepted by sanity check. changed to type="raw"

tests/integration/targets/bulk_host_group/tasks/test.yml
```diff
@@ -0,0 +1,62 @@
...
+- name: "{{ outer_item.version }} - Create host groups."
+  bulk_host_group:
+    server_url: "{{ server_url }}"
+    site: "{{ outer_item.site }}"
+    automation_user: "{{ automation_user }}"
+    automation_secret: "{{ automation_secret }}"
+    host_groups: "{{ checkmk_host_groups_create }}"
+    attributes:
+      site: "{{ outer_item.site }}"
+      ipaddress: 127.0.0.1
```
lgetwan, line 11: Please remove

tests/integration/targets/bulk_host_group/tasks/test.yml
```diff
@@ -0,0 +1,62 @@
...
+- name: "{{ outer_item.version }} - Modify part of host groups (also checks for idempotency!)."
+  bulk_host_group:
+    server_url: "{{ server_url }}"
+    site: "{{ outer_item.site }}"
+    automation_user: "{{ automation_user }}"
+    automation_secret: "{{ automation_secret }}"
+    host_groups: "{{ checkmk_host_groups_modify }}"
+    attributes:
+      site: "{{ outer_item.site }}"
+      ipaddress: 127.0.0.1
```
lgetwan, line 25: Please remove

lgetwan (changes requested):
Hi Michael.
besides the two little changes in the test playbook, your module runs fine.
Thanks a lot for contributing! :-D
Bet regards
Lars

tests/integration/targets/bulk_host_group/tasks/test.yml
msekania: removed

tests/integration/targets/bulk_host_group/tasks/test.yml
msekania: removed

msekania:
Hi @robin-tribe29,

Question: do contact group and host group have a the same namespace?
I get an error messages when I try to create a contact group with the name that already exists in host groups.

Best regards,
Michael

robin-tribe29:
Hi @msekania,

I am quite certain, that they do not share a namespace. I would rather think of a bug in your programming, maybe a classical copy/paste error?

Anyway, here is something more important: We discussed this feature proposal internally, and we see no value here.
You already have #154 in the pipeline and nearly ready to merge. The Ansible way would be to just use that module with a loop, rather than having an additional module, which also means possible additional maintenance.
So if you have no strong case in favor of this module, I have to reject it.

msekania:
Hi @robin-tribe29,

> I am quite certain, that they do not share a namespace. I would rather think of a bug in your programming, maybe a classical copy/paste error?

Yes, :-D, unfortunately in my test suite, again... should get more consistent (not with bugs though).

> The Ansible way would be to just use that module with a loop, rather than having an additional module, which also means possible additional maintenance.
> So if you have no strong case in favor of this module, I have to reject it.

Sure, it more Ansible way and you guys also know it better in which direction modules should be developed!
From my side it's actually the speed or better said how many times module contacts the server.
One bulk request vs. several single ones, let's say in case of more than 10 host groups.

As a proposal, one can also directly combine two in one. If one then provides host_groups, it will run bulk version and for host_group_name, title conventional (single) one.

I can prepare contact_group module which will combine bulk and conventional features. On the level what REST API allows.
And I can merge this module with comprehensive tests in the current pull request #160 or in a separate one.

Contact groups are essentially similar to host groups at least from REST API point view and we are already using it.

Best regards,
Michael

robin-tribe29:
> Sure, it more Ansible way and you guys also know it better in which direction modules should be developed! From my side it's actually the speed or better said how many times module contacts the server. One bulk request vs. several single ones, let's say in case of more than 10 host groups.

I get that and your thoughts make perfect sense here!
 
> As a proposal, one can also directly combine two in one. If one then provides host_groups, it will run bulk version and for host_group_name, title conventional (single) one.

I opt for on module, that can handle both. Maybe it is possible to somehow make the module aware, whether it creates one or more groups, to switch between bulk and single mode?
But let us first finish #154, so we have a clean state. Then we build upon that. Makes sense?
 
> I can prepare contact_group module which will combine bulk and conventional features. On the level what REST API allows. And I can merge this module with comprehensive tests in the current pull request #160 or in a separate one.
> Contact groups are essentially similar to host groups at least from REST API point view and we are already using it.

For the contact groups module feel free to  create a dedicated PR, we discuss everything further there.

PR #169, opened 2022-10-02
Avoids odd timeouts when running tasks delegated_to localhost

thorian93:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When using the agent role, there are odd timeouts, when Ansible tries to become root without having a proper password.

## What is the new behavior?

With the explicit `become: false`, those tasks will try to become root, hence avoiding the timeout.

### DIFF
--- a/roles/agent/tasks/Debian.yml
+++ b/roles/agent/tasks/Debian.yml
@@ -8,6 +8,7 @@
     headers:
       Authorization: "Bearer {{ checkmk_agent_user }} {{ checkmk_agent_auth }}"
       Accept: "application/octet-stream"
+  become: false
   when: checkmk_agent_edition == "cee" or checkmk_agent_edition == "cfe"
   register: checkmk_agent_download_state
   # This task may fail, as we fall back to the generic agent in that case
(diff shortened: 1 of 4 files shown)


robin-tribe29 (approved):

robin-tribe29:
This will be merged for version 0.12.0.

flybyray:
@robin-tribe29 , @thorian93 

Is there some reference to those timeouts? Maybe in the automatic testings?
I do not see such timeouts when I switch certain tasks to the default ansible become behaviour.

I have an issue with the logic here: 

https://github.com/tribe29/ansible-collection-tribe29.checkmk/blob/c09eb94ef2831270b0fbe0efa7a856912374898b/roles/agent/tasks/main.yml#L22-L44

Those 2 tasks combined should suffice that a downloaded package is on the managed node.
The first task may download it to a different "managed node" - depending on `checkmk_agent_delegate_download` - , but actually it would only work when this variable is set to `localhost`. Otherwise the second task will most probably fail.

In case `checkmk_agent_delegate_download` is set the package will be owned by the become user. Otherwise it would be owned by the `ansible_user` on the managed node.

But may real problem is that `become: false` on the first task and not setting `checkmk_agent_delegate_download`.
The `ansible_user` may not be allowed to download files on the managed node. Then the default behaviour of this role will fail.

Sure I can switch to the usage of `checkmk_agent_delegate_download` but as this actually only works if this is `localhost` it is not useful to me. a lot of copy operations slow down massive. 

From my point of view the usage of `checkmk_agent_delegate_download` should be extended to allow to download to a dedicated managed node ( e.g. in the locked down environment ) to distribute the copy operation later from.

PR #178, opened 2022-10-13
Feature service groups

msekania:
Additional plugin that manages service groups.

## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Issue Number: `e.g. #1`

## What is the new behavior?

- manage service groups in Checkmk (create/modify/delete)
- bulk and single mode for contact group management

## Other information

```yaml
- name: "Create a single service group."
  tribe29.checkmk.service_group:
    server_url: "http://localhost/"
    site: "my_site"
    automation_user: "automation"
    automation_secret: "$SECRET"
    service_group_name: "my_service_group"
    title: "My Service Group"
    state: "present"
  delegate_to: localhost

- name: "Create service groups."
  tribe29.checkmk.service_group:
    server_url: "http://localhost/"
    site: "my_site"
    automation_user: "automation"
    automation_secret: "$SECRET"
    service_groups: "my_service_groups"
    title: "My Service Group"
    state: "present"
  vars:
    my_service_groups:
      - name: "service_one"
        title: "Service One"
      - name: "mail"
        title: "IMAP"
  delegate_to: localhost
```


plugins/modules/service_group.py
```diff
@@ -0,0 +1,511 @@
...
+# i.e. the version is of the form "2.5.0" and not "2.4".
+version_added: "0.12.0"
+
+description:
+- Manage service groups in Checkmk.
+
+extends_documentation_fragment: [tribe29.checkmk.common]
+
+options:
+    service_group_name:
```
robin-tribe29, line 27: @msekania I just realized this: In all three 'group' modules including this one, you use `foo_group_name`. I feel like this should simply be `name`, as we get the context from the module name. Is there a reason you are doing this? Otherwise, I would ask you fix it in all three modules, before it gets used too much out there. Thanks!

P.S.: This is already getting inconsistent, so I will create a ticket to discuss how to deal with it. But for your modules, we should keep it simple with `name` for now.

msekania: @robin-tribe29 

> @msekania I just realized this: In all three 'group' modules including this one, you use `foo_group_name`. I feel like this should simply be `name`, as we get the context from the module name. Is there a reason you are doing this? Otherwise, I would ask you fix it in all three modules, before it gets used too much out there. Thanks!

There is no particular reason behind it, I was just following tag_group example.

I have made changes in both in Feature service group and Feature contact groups.

How shall we proceed with host groups (in which branch I should adjust it)?

Best,
Misha

robin-tribe29: I think you should use the current `devel` branch and really just fix the name. As this is a breaking change, I need to evaluate how to release it.

msekania: > As this is a breaking change, I need to evaluate how to release it.

What about keeping 'host_group_name' (marking it as deprecated) and adding 'name' as an alternative.
I know how to adjust the code for it, if it's a preferred option for the moment.

msekania: Dear @robin-tribe29,

> @msekania I just realized this: In all three 'group' modules including this one, you use `foo_group_name`.

There are foo_groups also in the same modules.
Shall we change them to just groups?

Best,
Michael

robin-tribe29: > What about keeping 'host_group_name' (marking it as deprecated) and adding 'name' as an alternative.
> I know how to adjust the code for it, if it's a preferred option for the moment.
Hm. I like the idea of deprecation in general, but I feel like this time, I would rather have a breaking change, that would affect only a few users.

> There are foo_groups also in the same modules.
> Shall we change them to just groups?
Another good point. I feel like both ways would be fine here, in contrast to the `name` attribute. Do you have a strong opinion on this?

robin-tribe29: @msekania I made up my mind about the `foo_groups` option: I want to go with `groups` for the two open PRs.

msekania: Fine, then style will be also consistent.

I'll make corresponding changes in corresponding pull requests for service and contact groups today.
I also prepare changes for the host group,  for devel branch, but I do not know how to push them then

msekania: @robin-tribe29 service and contact groups are modified and pushed.

host_groups is also ready. I do not know where to push it

robin-tribe29: We really do appreciate your efforts @msekania, thanks! :heart: 

I will merge as soon as I find a few moments of peace.

Regarding the `host_groups` changes: You need to create a dedicated PR from your branch, where you did the change. You cannot push directly to this repository ever.

robin-tribe29 (approved):

PR #183, opened 2022-10-27
Local download

tonk:
In our site only the Ansible Control node is allowed to contact the Checkmk server on port 443. So the download of the agent
and the API calls need to go through the Ansible control node. 

I have added an `checkmk_agent_delegate_download` option to delegate the download.

## Pull request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [x] Documentation content changes
- [x] Other (please describe):
   For Red Hat and Suse based systems the agent was installed multiple times. Removed this duplicate code.

## What is the current behavior?

It is now possible to let the Ansible Control node download the Checkmk agent and transfer it to the remote node.

## What is the new behavior?

- Download the Checkmk agent on the control and transfer it when needed

PR #185, opened 2022-10-28
Update feature host groups

msekania:
Changed host_group_name and host_groups with name and groups, respectively, also to match two other groups, service and contact.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [x] Build related changes

Issue Number: `e.g. #1`

## What is the new behavior?

- `host_group_name` is changed with simple `name`
- `host_groups` is changed with simple `groups`


msekania:
@robin-tribe29  what I see issues are with "Local download #183" branch and not with "Update feature host groups #185"!

lgetwan (approved):
I very much appreciate this change. :-)

robin-tribe29:
Still finishing some details, but this will merge rather soon. Will keep you posted.

robin-tribe29:
I will merge this after my vacation, I want to make sure to inform users properly about the breaking change.

robin-tribe29:
Hi @msekania! Sorry to keep you waiting here and even more apologies for the following:
We discussed this change internally, and we came to the conclusion, that your idea of aliases makes a lot of sense.
Can you tell, if it is a lot off effort to enable the old `host_group_name` and `host_groups` as additional aliases? That way, existing configuration does not break, while we can still go for the new naming and remove the old option names at a later time.

msekania:
Hi @robin-tribe29,
I need to recall how I planned it, but latest by Monday I will have it ready.

robin-tribe29:
Awesome dude, thanks! :pray:

plugins/modules/host_group.py
```diff
@@ -24,18 +24,20 @@
 extends_documentation_fragment: [tribe29.checkmk.common]
 
 options:
-    host_group_name:
+    name:
         description: The name of the host group to be created/modified/deleted.
         type: str
+    host_group_name: depricated alias of name
```
robin-tribe29: "depricated" should be "deprecated".
Also, let's put it like this:
"Deprecated alias of I(name). This will be removed in a future release. Make sure to update your configuration to use I(name)!"

robin-tribe29 (changes requested):
Just the little nit pic. If did not miss something, we are good to merge this for the next release.

plugins/modules/host_group.py
msekania: Unfortunately, does not work like that, one should add the field as alias in documentation also.

I think it's done!

msekania: It should be done differently.

robin-tribe29: I will review this hopefully within this week, thanks!

plugins/modules/host_group.py
```diff
@@ -334,19 +336,42 @@ def run_module():
         validate_certs=dict(type="bool", required=False, default=True),
         automation_user=dict(type="str", required=True),
         automation_secret=dict(type="str", required=True, no_log=True),
-        host_group_name=dict(type="str", required=False),
+        name=dict(
+            type="str",
+            required=False,
```
robin-tribe29, line 341: I just saw this: I think this has to be `required=True`, as the Checkmk UI also declares it as "required".

plugins/modules/host_group.py
```diff
@@ -334,19 +336,42 @@ def run_module():
...
+            aliases=["host_group_name"],
+            deprecated_aliases=[
+                {
+                    "name": "host_group_name",
+                    "date": "2024-01-01",
+                    "collection_name": "tribe29.checkmk",
+                }
+            ],
+        ),
         title=dict(type="str", required=False),
```
robin-tribe29, line 351: I just saw this: I think this has to be `required=True`, as the Checkmk UI also declares it as "required".

plugins/modules/host_group.py
```diff
@@ -334,19 +336,42 @@ def run_module():
...
-        host_groups=dict(type="raw", required=False, default=[]),
+        groups=dict(
+            type="raw",
+            required=False,
+            default=[],
+            aliases=["host_groups"],
+            deprecated_aliases=[
+                {
+                    "name": "host_groups",
+                    "date": "2024-01-01",
```
robin-tribe29, line 360: Let's make this "2023-03-01". There is little reason to make the impression, that this option might stay longer than necessary.

plugins/modules/host_group.py
```diff
@@ -334,19 +336,42 @@ def run_module():
...
         automation_secret=dict(type="str", required=True, no_log=True),
-        host_group_name=dict(type="str", required=False),
+        name=dict(
+            type="str",
+            required=False,
+            aliases=["host_group_name"],
+            deprecated_aliases=[
+                {
+                    "name": "host_group_name",
+                    "date": "2024-01-01",
```
robin-tribe29, line 346: Let's make this "2023-03-01". There is little reason to make the impression, that this option might stay longer than necessary.

robin-tribe29 (changes requested):

plugins/modules/host_group.py
msekania: `required=True` will not work since there is an alternative parameter to it which is `groups`

plugins/modules/host_group.py
msekania: `required=True` will not work since there is an alternative parameter to it which is `name`

plugins/modules/host_group.py
msekania: changed in both places

robin-tribe29 (approved):

PR #156, opened 2022-09-21
Feature agent windows support

robin-checkmk:
## Pull request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

There is no support for Windows clients in the agent role.

## What is the new behavior?

We want to introduce Windows support for the agent role.

## Other information

Currently, there are no resources to start developing this. If you have the time and possibilities, have at it. :slightly_smiling_face: 
There are files from another project here, but you should be able to get started from the existing tasks for Linux.
If you have questions, feel free to drop them here.

PR #187, opened 2022-11-01
Feature agent windows support

muehlings:
## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

## What is the new behavior?
This is my contribution to PR #156

- Changes in linux agent roles
- Addition in windows agent roles
- Preparation files for WinRM to work


muehlings:
Sorry, but I have no idea how to change the behavior of github concerning these checks:
1. Running sanity test "line-endings"
```
ERROR: Found 3 line-endings issue(s) which need to be resolved:
ERROR: preparation/ansible-winrm/00-ansible.cmd:0:0: use "\n" for line endings instead of "\r\n"
```
Line ending of \r\n is correct, because those are windows cmd files. I tried to set the line ending in .gitattributes to LF or CRLF but the check always fails.

2. Running sanity test "pslint"
```
ERROR: Found 6 pslint issue(s) which need to be resolved:
ERROR: preparation/ansible-winrm/ConfigureRemotingForAnsible.ps1:281:58: PSCustomUseLiteralPath: Use the explicit -LiteralPath parameter name instead of -Path
```
This script is from [ansible ](https://github.com/ansible/ansible/blob/devel/examples/scripts/ConfigureRemotingForAnsible.ps1) and I don't want to change anything in it.

I would prefer to skip the tests on both types of files (.cmd and .ps1) at the moment.

robin-tribe29:
I agree, that we can ignore those tests right now.
Let me review your changes and try to understand them, before we merge them.

playbooks/example-linux-inventory.yml
```diff
@@ -0,0 +1,23 @@
+---
```
robin-tribe29, line 1: Do we need this playbook, or is it for your local testing only? If the latter is true, please remove it from the PR.

playbooks/example-windows-inventory.yml
```diff
@@ -0,0 +1,46 @@
+---
```
robin-tribe29, line 1: Do we need this playbook, or is it for your local testing only? If the latter is true, please remove it from the PR.

roles/agent/tasks/Debian.yml
```diff
@@ -1,62 +1,71 @@
 ---
-- name: "Debian Derivates: Download host-specific Checkmk {{ checkmk_agent_edition | upper }} Agent."
+- name: "({{ansible_os_family}}) Derivates: Download host-specific Checkmk {{ checkmk_agent_edition | upper }} Agent."
```
robin-tribe29, line 2: Why are you doing this? We are clearly in Debian.yml, so we can - and in my opinion should - hard code it in the task name. Unless you can convince me, why this makes sense, I would ask you to revert it. Especially to keep the PR as clean as possible.

roles/agent/tasks/Linux.yml
```diff
@@ -0,0 +1,68 @@
+---
```
robin-tribe29, line 1: Is this new file structure necessary for your PR to work, or can we do changes one step at a time?

roles/agent/tasks/Win32NT.yml
```diff
@@ -0,0 +1,15 @@
+---
```
robin-tribe29, line 1: Why this distinction with Win32NT? Is this related to the overall change in the file structure?

robin-tribe29 (changes requested):
Overall I see a lot of potential here, but I think we need to make sure we focus on the task at hand. I would like to keep the PR as small and simple as possible, before doing major reorganization of the role.

roles/agent/tasks/Debian.yml
muehlings: I had to move a lot of tasks from the new generic "ansible_system" (Win32NT/Linux) to the "ansible_os_family" and back to see, if they were running like expected and in the right context. The download agent part is for example identical on all linux derivates because it is an builtin ansible method. I left the names with the variables because it doesn't hurt to have it generalized across the includes.

roles/agent/tasks/Linux.yml
muehlings: main.yml had to be split up because it was Linux-only. Linux.yml is the generic linux branch with ansible-related tasks and nothing derivate-specific.

roles/agent/tasks/Win32NT.yml
muehlings: main.yml had to be split up because it was Linux-only. Win32NT.yml is the generic Windows branch with ansible-related tasks and nothing derivate-specific.

The role structure now looks like:
```
main
+-Linux (ansible_system)
|  +- Debian (ansible_os_family)
|  +- RedHat (ansible_os_family)
|  +- SuSE (ansible_os_family)
|  +- ...
+- Win32NT (ansible_system)
   +- Windows (ansible_os_family)
```

ansible_system contains the following tags/commands (mostly shell):
- get-package-facts
- download-package
- include-os-family-tasks
- cmk-update-agent
- cmk-agent-ctl

ansible_os_family contains the following tags/commands (mostly package manager):
- download-package (host specific)
- install-package (apt, yum, ..., win_package)
- configure_firewall

playbooks/example-linux-inventory.yml
muehlings: They were for local testing. I deleted them.

playbooks/example-windows-inventory.yml
muehlings: They were for local testing. I deleted them.

PR #208, opened 2022-12-14
allow to specify the position of created rules

geof77:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

No way to specify the position of the rule in the rule folder, so it is always created at the bottom.

## What is the new behavior?

There is a new "position" parameter to specify the position of the rule.

My use case was to create service discovery rules before the generic, built-in, two-hours rule.

I've added two functions to rules.py

get_rule_etag : gets the rule ETag needed for the next operation (could be useful elsewhere)
move_rule: move the rule to the specified location as defined in the API documentation

I also added a new parameter I've called "position", it's a dict with the position and a rule_id (only for moves that need it: before_specific_rule and after_specific_rule).

Please review the PR. Suggestions/requests are welcome.

## Other information

Tested with CheckMK 2.1p14

### DIFF
--- a/plugins/modules/rule.py
+++ b/plugins/modules/rule.py
@@ -453,8 +453,8 @@ def run_module():
                         module, "Position 'folder' is not valid with 'state=present'"
                     )
                 move["folder"] = rule.get("folder")
-            if move.get("rule_id") is not None:
-                exit_failed(module, "Position does not support 'rule_id'")
+                if move.get("rule_id") is not None:
+                    exit_failed(module, "Position does not support 'rule_id'")
 
     # Get ID of rule that is the same as the given options
     rule_id = get_existing_rule(module, base_url, headers, ruleset, rule)
--- a/plugins/modules/rule.py
+++ b/plugins/modules/rule.py
@@ -109,10 +109,14 @@ EXAMPLES = r"""
 
 RETURN = r"""
 msg:
-    description: The output message that the module generates. Contains the API response details in case of an error.
+    description: The output message that the module generates. Contains the API status details in case of an error.
     type: str
     returned: always
     sample: 'Rule created.'
+content:
+    description: The Response body from the API when creating or moving a rule, or when a rule already exists.
+    type: dict
+    returned: when rule exists, is created or moved
 """
 
 import json
@@ -126,18 +130,18 @@ except ImportError:  # For Python 3
     from urllib.parse import urlencode
 
 
-def exit_failed(module, msg):
-    result = {"msg": msg, "changed": False, "failed": True}
+def exit_failed(module, msg, content={}):
+    result = {"msg": msg, "content": content, "changed": False, "failed": True}
     module.fail_json(**result)
 
 
-def exit_changed(module, msg):
-    result = {"msg": msg, "changed": True, "failed": False}
+def exit_changed(module, msg, content={}):
+    result = {"msg": msg, "content": content, "changed": True, "failed": False}
     module.exit_json(**result)
 
 
-def exit_ok(module, msg):
-    result = {"msg": msg, "changed": False, "failed": False}
+def exit_ok(module, msg, content={}):
+    result = {"msg": msg, "content": content, "changed": False, "failed": False}
     module.exit_json(**result)
 
 
@@ -176,7 +180,7 @@ def get_existing_rule(module, base_url, headers, ruleset, rule):
                 and sorted(r["extensions"]["value_raw"]) == sorted(rule["value_raw"])
             ):
                 # If they are the same, return the ID
-                return r["id"]
+                return r
     return None
 
 
@@ -204,8 +208,7 @@ def create_rule(module, base_url, headers, ruleset, rule):
             % (info["status"], info["body"]),
         )
 
-    rule_id = json.loads(response.read().decode("utf-8"))["id"]
-    return rule_id
+    return json.loads(response.read().decode("utf-8"))
 
 
 def get_rule_etag(module, base_url, headers, rule_id):
@@ -256,6 +259,8 @@ def move_rule(module, base_url, headers, rule_id, position):
             "Error calling API. HTTP code %d. Details: %s, "
             % (info["status"], info["body"]),
         )
+
+    return json.loads(response.read().decode("utf-8"))
 
 
 def delete_rule(module, base_url, headers, rule_id):
@@ -323,8 +328,11 @@ def run_module():
             "host_labels": [],
             "service_labels": [],
         }
+
     # Get ID of rule that is the same as the given options
-    rule_id = get_existing_rule(module, base_url, headers, ruleset, rule)
+    content = get_existing_rule(module, base_url, headers, ruleset, rule)
+    rule_id = content.get("id")
+
     # If rule exists
     if rule_id is not None:
         # If state is absent, delete the rule
@@ -333,18 +341,19 @@ def run_module():
             exit_changed(module, "Deleted rule")
         # If state is present, do nothing
         else:
-            exit_ok(module, "Rule already exists")
+            exit_ok(module, "Rule already exists", content)
     # If rule does not exist
     else:
         # If state is present, create the rule
         if module.params.get("state") == "present":
-            rule_id = create_rule(module, base_url, headers, ruleset, rule)
+            content = create_rule(module, base_url, headers, ruleset, rule)
             # move the rule into position if specified
             if module.params.get("position") is not None:
+                rule_id = content.get("id")
                 position = module.params.get("position")
                 position["folder"] = rule.get("folder")
-                move_rule(module, base_url, headers, rule_id, position)
-            exit_changed(module, "Created rule")
+                content = move_rule(module, base_url, headers, rule_id, position)
+            exit_changed(module, "Created rule", content)
         else:
             # If state is absent, do nothing
             exit_ok(module, "Rule did not exist")
--- a/plugins/modules/rule.py
+++ b/plugins/modules/rule.py
@@ -36,11 +36,6 @@ options:
         description: Position of the rule in the folder
         required: false
         type: dict
-        choices:
-            - {"position": "bottom_of_folder"} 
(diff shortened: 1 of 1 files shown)


robin-tribe29:
The integration tests are flaky, but please try to fix the findings of the other tests. If you need help, do let me know.

robin-tribe29:
Hi @geof77 and thanks for the update. We are waiting for a review of your code before moving forward with the merge.
We will keep you posted here. Thanks again for your contribution!

plugins/modules/rule.py
```diff
@@ -199,6 +204,58 @@ def create_rule(module, base_url, headers, ruleset, rule):
...
+
+
+def move_rule(module, base_url, headers, rule, rule_id, position):
+    api_endpoint = "/objects/rule/" + rule_id + "/actions/move/invoke"
+
+    if position.get("position") not in [
+        "bottom_of_folder",
+        "top_of_folder",
+        "after_specific_rule",
+        "before_specific_rule",
```
lgetwan, line 234: Can you add code examples for "before_specific_rule" and "bottom_of folder", as well?

plugins/modules/rule.py
```diff
@@ -199,6 +204,58 @@ def create_rule(module, base_url, headers, ruleset, rule):
...
+    ] or (
+        position.get("position") in ["after_specific_rule", "before_specific_rule"]
+        and (position.get("rule_id") is None or position.get("rule_id") == "")
```
lgetwan, line 237: The rule_id of the existing rule (for the "before_" or "after_") has to be provided in the ansible task, correct?
As far as I can see, there's no way to query the rule_id of an existing rule with our collection, yet.
Have you thought of returning the ID of a freshly created rule back to the ansible caller?

Having an Ansible lookup plugin for that, one day, would be great, too. :-)

plugins/modules/rule.py
```diff
@@ -199,6 +204,58 @@ def create_rule(module, base_url, headers, ruleset, rule):
             % (info["status"], info["body"]),
         )
 
+    rule_id = json.loads(response.read().decode("utf-8"))["id"]
```
lgetwan, line 207: What if the response cannot be decoded or "id" is not found in the resulting dict?

lgetwan (changes requested):
Great idea to make the rules position aware!
I haven't tested your code yet, I just read it, and I added some little ideas on how to further improve it and make it bulletproof.

Thanks for your work! :-)

plugins/modules/rule.py
geof77: Yes I already have code for returning the rule_id so in can be reused in subsequent tasks.  But I'm wondering if it would be useful to return the whole response from the API instead of the ID only. Maybe there are useful values other than the ID in there...

plugins/modules/rule.py
geof77: I assumed the HTTP status would be different from 200 and this part of the code would never be reached. But, ok, let's make this better.

plugins/modules/rule.py
lgetwan: That's correct. Hopefully, I'll have a bit of time for that in Q1 next year. If you're faster than that and you have a good idea how to solve that, go for it. :-D

geof77: Now returns the response from the API.

plugins/modules/rule.py
geof77: Done.

plugins/modules/rule.py
geof77: I had a look at the other modules in the collection, none of them check for decode exceptions, so the module would just fail and we would see the cause in the ansible debug output...

plugins/modules/rule.py
msekania: Just a small question, @geof77 @lgetwan,

would not it be more appropriate to use  "bottom_of_ruleset" and "top_of_ruleset" instead of  "bottom_of_folder", and "top_of_folder", since these terms address positions within the ruleset?

geof77: Sure, however these are the keywords used by the API. I don't like this, but that's the way it is....

geof77: What about simply having "top", "bottom", "before" and "after". That would be easier.
Also, having to write "position" twice is not pretty.

`    position: {"position":"top_of_folder"}`

So I would rather have something like:

```
    move:
       positiion: "top"
```

Or 

```
    move:
      position: "after"
      rule_id: "123456789abcdef"
```

Is that better?

geof77: To better understand what "move" does, this is what I came up with:

```
    move:
        description:
          - Move the rule at the specified location
          - Mutually exclusive with state: absent
        type: dict
        suboptions:
            position:
                description: Position of the rule in the folder
                required: true
                type: str
                choices:
                    - "top"
                    - "bottom"
                    - "before"
                    - "after"
            rule_id:
                description:
                    - Move the rule "before" or "after" this rule_id
                    - Required with position: "before" or "after"
                    - Mutually exclusive with position: "top" or "bottom"
                type: str
            folder:
                description:
                    - Folder the rule should be moved to
                    - Required with position: "top" or "bottom"
                    - Mutually exclusive with position: "before" and "after"
                default: "/"
```

I already have this in my fork but still polishing the code.

robin-tribe29: From a non-developer perspective this looks great to me!
Looking forward to the finalized version.

geof77:
Now we have:
* the new "move" param
* the API response returned so we can get the rule id out of it

To keep this PR simple, the move param only works with rule creation for now. Next I'd like to introduce a new "rule_id" parameter. With this new parameter we will be able to use rule IDs to move or delete existing rules. I left some examples in the EXAMPLES section.

geof77:
It's been a few days now and before going further I'd like your advice on a few design questions:
* Do you prefer the module returns only the rule id or the entire response from the API?  Do you think there is anything useful in the response other than the ID?
*  The response from the API is currently stored in a dict named "content" because this is what the uri module uses. But it might be better/prettier to call it something else, maybe "response" ? OK with "content"?
* Next I will add the rule_id param so we can move or delete existing rules by providing their rule id like in the examples I have left in the code (search for "TODO"). What do you think of this? Do I proceed here or should I make another PR for this?
Comments/suggestions are welcome...

geof77:
Well I finally managed to get ansible-test running on my setup. Hopefully this passes the tests now. Sorry for the errors.

robin-tribe29:
@geof77 no need to apologize, we do definitely appreciate your work! :pray:

I am no developer though, and we are always short on resources, so please bear with us until someone can answer your questions and review the code properly.

geof77:
@lgetwan : Indeed I don't see anything useful other than the ID in the response. So I'm going to make the change and just call it "id". That's straightforward and prettier :) We can always add a field with the full content in the future, if we ever need it.

plugins/modules/rule.py
```diff
@@ -32,6 +32,31 @@
...
+                choices:
+                    - "top"
+                    - "bottom"
+                    - "before"
+                    - "after"
+            rule_id:
+                description:
+                    - Put the created/moved rule C(before) or C(after) this rule_id.
+                    - Required when I(position) is C(before) or C(after).
+                    - Mutually exclusive with I(position=before) and I(after).
```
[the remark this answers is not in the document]
robin-tribe29: I for my part do agree with @lgetwan. :upside_down_face:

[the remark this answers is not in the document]
geof77: Ok for me to rename it "location". 

I wonder if "location" should be a suboption of the "rule" param rather than a top-level option. Maybe it could be used instead of the existing "folder" suboption (which is redundant and would be deprecated). Currently, "folder" cannot be specified under 'location' and the folder is simply copied from the other "folder" option.

I.E.:

```
rule:
  location:
     folder: "/tst" 
     position: "top"
```
looks better than
```
rule:
  folder: "/tst"
location:
   position: "top"
```

Here's what I also have in the queue for now :

* better parameter checking, moved to run_module instead of inside the move_rule function
* do not return a diffrent message if the rule is moved
* documentation correction for the errors you spotted and some others
* ability to use a rule_id instead of a full rule description
* Idempotency is a bit tricky. Currently, when the rule is already created and the task is re-run, the location will not change, so if we create a rule at the top of a fodler, than a bunch of others on top of that, we might not want the first rule to be moved to the top again. So maybe introduce a 'force_move' option...
* moving existing rules
* checking if the rule has moved so we return the correct status ("ok" vs. "changed")
* ...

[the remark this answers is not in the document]
lgetwan: This looks indeed better, I agree:

```
rule:
  location:
     folder: "/tst" 
     position: "top"
```
Regarding your to-do queue: Please only include _those_ of them into this PR, that are necessary to have a working version. Then, we can have this released, and people can use it. :-)
The rest of your queue can then result in (a) new PR(s).

Thanks again for your contribution!

geof77:
To be clear, what this PR does for end users is:
* return the ID of the rule when it is created or when it already exists so it can be used in subsequent tasks
* add a new "location" sub-option to the "rule" parameter for precise positioning of created rules
* deprecate the "folder" sub-option in favor of the new "location" sub-option

Internally, this means:
* a new get_rule_etag function to get the etag header required when moving a rule
* a new move_rule function to move the rule at the specified location, called after the rule is created
* modification of the create_rule function so it returns the ID of the created rule
* modification of the exit_* functions to make the rule ID available in ansible tasks
* arguments validation and default settings for the rule option and the location sub-option
* documentation of the rule option and its location sub-option
* examples using the new location option

lgetwan:
That sounds excellent!
Today, I will do some testing and then, most likely, merge your PR.

geof77:
@lgetwan : great! I have just added the deprecation warning if forgot to include so users are warned.

lgetwan (approved):
I did some tests, and they were all successful. :-) 
Your PR makes the rule module even better.
Thanks for your contribution!

Issue #180, opened 2022-10-24
[FEED] Harmonize module parameters


robin-tribe29:
Please verify first that your feedback is not already reported [here](https://github.com/tribe29/ansible-collection-tribe29.checkmk/issues?q=is%3Aissue+sort%3Aupdated-desc).
If possible and useful provide additional information below. This is optional however.

**Provide your feedback**

We already digressed with the naming of module parameters, namely the `name` of the managed object in the below-mentioned modules. This issue serves as a ground for discussion and work tracker towards harmonization.

This issue is updated regularly.

**Component Name**

- [x] `service_group`
- [x] `tag_group`
- [x] `contact_group`
- [x] `discovery` -> Not applicable here. `host_name` makes sense, just `name` does not.
- [x] `host`
- [x] `folder`

**Ansible Version**

```console
$ ansible --version
ansible [core 2.13.4]
  python version = 3.10.6 (main, Aug 10 2022, 11:40:04) [GCC 11.3.0]
  jinja version = 3.1.2
  libyaml = True
```

**Collection Version**

```console
$ ansible-galaxy collection list
Collection        Version
----------------- -------
tribe29.checkmk   0.11.0 
```
**Environment**

**Screenshots**

**Additional context**

msekania:
Dear @robin-tribe29 

There is a "host_name" field in host module.
To be consistent, then this should be also changed to just 'name'

It's a design issue, which one should decide once and carry on through all modules then.
One can also make this style decision for all group modules only. It will be then also consistent.

I do not mind to make changes if we all agree on name conventions.

Best,
Michael

robin-tribe29:
@msekania thanks for the input. I will investigate some more and discuss this internally. This will take some time though, but I feel like it is not that urgent.
For the open PRs, I am uncertain right now, if we should wait or not. I will have more to share at the end of next week.

robin-tribe29:
See the initial issue description for the current state of affairs.
The group modules are clean since the last release.
Currently looking into other modules, which might be inconsistent.
I also added some development guidelines for this, currently to be found in the devel branch.

PR #213, opened 2022-12-21
Make module names more consistent using aliases.

robin-tribe29:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number:  #180 

## What is the new behavior?

- The host module now uses `name` instead of `host_name`. The latter is retained as an alias but will be removed with a future release.
- The folder module now uses `name` instead of `title`. The latter is retained as an alias until further notice.


lgetwan (commented):
Looks good!

Issue #219, opened 2023-01-09
[FEED] Support Agent registration on remote site


robin-tribe29:
Please verify first that your feedback is not already reported [here](https://github.com/tribe29/ansible-collection-tribe29.checkmk/issues?q=is%3Aissue+sort%3Aupdated-desc).
If possible and useful provide additional information below. This is optional however.

**Provide your feedback**

Currently, the agent role can only talk to one site, which has to be the central site for most things to work.
This works in a lot of environments and makes sense, but it brings along limitations, which we need to address.

The goal of this issue is to:
- [ ] Enable agent download through remote site
- [ ] Enable agent registration for TLS and agent updates against remote site

**Component Name**

`agent role`

**Ansible Version**

```console
$ ansible --version

```

**Collection Version**

```console
$ ansible-galaxy collection list

```
**Environment**

**Screenshots**

**Additional context**

robin-tribe29:
Remote registration is now possible in [0.16.0](https://github.com/tribe29/ansible-collection-tribe29.checkmk/releases/tag/v0.16.0).
Remote download is pending on upstream.

DominikBauer1:
I think the remote registration does not work in combination with `checkmk_agent_add_host:true` and `checkmk_agent_auto_activate:true`.

In this case the new hosts are created on the central site and only activated on the central site, but not on the remote site where the registration takes place. This leads to an error in `checkmk.general.agent : Linux: Register Agent for TLS`. 

A fix would be to introduce a second handler wich only runs when `checkmk_agent_site != checkmk_agent_registration_site`

robin-checkmk:
I have not tested this, only refreshed my memory on how we designed this.

Your solution is to provide `checkmk_agent_host_attributes` or a `checkmk_agent_folder`, that has the host monitored on the remote site. The handler activates changes globally. Your issue sounds like the host is created on the central site (which will happen, if you stick with the defaults) and then registration fails against the remote site.

I am of course happy to update/improve documentation or even be proven wrong with my diagnosis. :)

DominikBauer1:
Hi, 

yes. in my case i set the `checkmk_agent_registration_server` and `checkmk_agent_registration_site` site explicitly, but the issue should be the same as with setting it via `checkmk_agent_folder`

This is config dump:  

```yaml
checkmk_agent_protocol: https
checkmk_agent_server: cmk_server1
checkmk_agent_server_protocol: https
[...]

PR #223, opened 2023-01-13
Enable agent registration on remote sites

robin-tribe29:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

#219 

## What is the new behavior?

- Enable agent registration on remote sites
- Download is not possible yet


roles/agent/README.md
```diff
@@ -46,6 +46,14 @@ The port of the web interface of your Checkmk server. Defaults to port 80 for ht
 
 The name of your Checkmk site.
 
+    checkmk_agent_registration_server: "{{ checkmk_agent_server }}"
+
+The server you want to use for registration tasks (Agent updates and TLS encryption). Defaults to {{ checkmk_agent_server }}.
+
+    checkmk_agent_registration_site: "{{ checkmk_agent_site }}"
+
+The server you want to use for registration tasks (Agent updates and TLS encryption). Defaults to {{ checkmk_agent_site }}.
```
lgetwan, line 55: I guess, you meant "site".

roles/agent/tasks/main.yml
```diff
@@ -47,7 +50,7 @@
 
 - name: "Create host on server."
   tribe29.checkmk.host:
-    server_url: "{{ checkmk_agent_protocol }}://{{ checkmk_agent_server }}:{{ checkmk_agent_port }}/"
+    server_url: "{{ checkmk_agent_protocol }}://{{ checkmk_agent_registration_server }}:{{ checkmk_agent_port }}/"
```
lgetwan, line 53: Shouldn't the host always be created on the central site? And the API to talk to should also be the central one, shouldn't it?

robin-tribe29: Interesting. I am almost certain, that this worked in my tests. I think I will need to reiterate.
Can you test this on your end too? Just to have a separate view.

roles/agent/README.md
robin-tribe29: Jup, copa-paste error. Already fixed. :v:

roles/agent/tasks/main.yml
lgetwan: It's even complaining that WATO is disabled on the remote site if I use checkmk_agent_registration_server, here.
To configure where a host should reside, the host attributes should contain the site. E.g.
`checkmk_agent_host_attributes:  
    site: "multi_remote_1"`

roles/agent/tasks/main.yml
```diff
@@ -78,7 +81,7 @@
   become: true
   ansible.builtin.shell: |
     cmk-update-agent register -H {{ checkmk_agent_host_name }} \
-    -s {{ checkmk_agent_server }} -i {{ checkmk_agent_site }} -p {{ checkmk_agent_protocol }} \
+    -s {{ checkmk_agent_registration_server }} -i {{ checkmk_agent_site }} -p {{ checkmk_agent_protocol }} \
```
lgetwan, line 84: Should be 
`{{ checkmk_agent_registration_site }}`

roles/agent/tasks/main.yml
```diff
@@ -94,7 +97,7 @@
...
```
lgetwan, line 100: Should be 
`{{ checkmk_agent_registration_site }}`

lgetwan (commented):
We also need an "Activate Changes" after creating the host. Otherwise, the TLS-Registration is complaining "Host xyz does not exist."

roles/agent/tasks/main.yml
DominikBauer1: I can confirm this. If WATO is enabled it creates an inconsistency where the host is available on the remote site but not on central.

DominikBauer1:
solves #232

roles/agent/tasks/main.yml
robin-tribe29: Fixes are pushed, along with automatic change activation.
Please re-test and do let me know. :V:

Issue #195, opened 2022-11-22
[BUG]  rule module does not respect parameter folder


msekania:
**rule module does not respect parameter folder**

module `rule` does not respect the parameter `folder`, only the parameters `conditions`, `properties`, `value_raw`.

It does not take into account that `folder` might be also different.

As a result one cannot assign the rule with the same `conditions`, `properties`, `value_raw` to two different folders.

With REST API it is possible!

snippet from the rule module
```python
def get_existing_rule(module, base_url, headers, ruleset, rule):
    # Get rules in ruleset
    rules = get_rules_in_ruleset(module, base_url, headers, ruleset)
    if rules is not None:
        # Loop through all rules
        for r in rules.get("value"):
            # Check if conditions, properties and values are the same
            if (
                sorted(r["extensions"]["conditions"]) == sorted(rule["conditions"])
                and sorted(r["extensions"]["properties"]) == sorted(rule["properties"])
                and sorted(r["extensions"]["value_raw"]) == sorted(rule["value_raw"])
            ):
                # If they are the same, return the ID
                return r["id"]
    return None
```
There is no check for `r["extensions"]["folder"]`

**Component Name**

`rule`

**Ansible Version**

any

**Checkmk Version**

any

**Collection Version**

any with `rule` module

**Environment**

**To Reproduce**

Example, only `folder` is different

```yaml
- name: "Create a rule in checkgroup_parameters:memory_percentage_used for production/proxies"
  tribe29.checkmk.rule:
    server_url: "http://localhost/"
    site: "my_site"
    automation_user: "automation"
    automation_secret: "$SECRET"
    ruleset: "checkgroup_parameters:memory_percentage_used"
    rule:
        folder: "/production/proxies"
        conditions: {
            "host_labels": [],
            "host_tags": [],
            "service_labels": []
        }
        properties: {
            "comment": "Warning at 80%\nCritical at 90%\n",
            "description": "Allow higher memory usage",
            "disabled": false,
            "documentation_url": ""
        }
        value_raw: "{'levels': (80.0, 90.0)}"
    state: "present"

- name: "Create a rule in checkgroup_parameters:memory_percentage_used for stging/proxies"
  tribe29.checkmk.rule:
    server_url: "http://localhost/"
    site: "my_site"
    automation_user: "automation"
    automation_secret: "$SECRET"
    ruleset: "checkgroup_parameters:memory_percentage_used"
    rule:
        folder: "/staging/proxies"
[...]

Issue #186, opened 2022-10-31
[BUG] value_raw set to a variable


clagio:
**Describe the bug**
I'm trying to set the value of value_raw to a variable instead of hardcoding it in the task, so I can reuse the same task for multiple rules. Unfortunately I get the error "Unsupported type. Field must be string". The same task with the value_raw explicitly set works fine. Seems to be an escaping issue, and I'm not 100% the issue is with the collection honestly.. but I'm banging my head on this so any help is appreciated. 

**Component Name**
tribe29.checkmk.rule

**Ansible Version**

```console
$ ansible --version
ansible 2.10.8
python version = 3.9.2 (default, Feb 28 2021, 17:03:44) [GCC 10.2.1 20210110]
```

**Checkmk Version**
```
Checkmk Raw Edition 2.1.0p13
```

**Collection Version**
```console
$ ansible-galaxy collection list
tribe29.checkmk   0.11.0
```

**To Reproduce**
Execute this task:
```
    - name: "Create a rule."
      tribe29.checkmk.rule:
        server_url: "{{ hostvars['checkmk'].checkmk_url }}"
        site: "{{checkmk_site}}"
        automation_user: "{{ hostvars['checkmk'].checkmk_user }}"
        automation_secret: "{{ hostvars['checkmk'].checkmk_secret }}"
        ruleset: "checkgroup_parameters:memory_percentage_used"
        rule:
            properties: {
                "comment": "Warning at 80%\nCritical at 90%\n",
                "description": "Allow higher memory usage",
                "disabled": false
            }
            value_raw: "{{ myvariable }}"
        state: "present"
      vars:
        - myvariable: "{'levels': (80.0, 90.0)}"
```
If I replace ``` value_raw: "{{ myvariable }}" ``` to ``` value_raw: "{'levels': (80.0, 90.0)}" ``` the task works fine.

**Expected behavior**
Rule to be created using the value in "myvariable"

**Actual behavior**
The following error is printed:
```
"msg": "Error calling API. HTTP code 400. Details: b'{\"title\": \"Bad Request\", \"status\": 400, \"detail\": \"These fields have problems: value_raw\", \"fields\": {\"value_raw\": [\"Unsupported type. Field must be string.\"]}}', "
```

**Additional context**
I suspect the issue is caused by the curly bracket required by the API which make ansible interpret the variable's value as a dictionary instead of a string. But I couldn't find a way to escape that and make it works with the API.

[...]

PR #224, opened 2023-01-16
Bugfix rule module folder handling

lgetwan:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

#195 

## What is the new behavior?

- When checking if the rule already exists, we now also take care of the rule's folder.
- Creating equal rules in multiple folders is now possible.

### DIFF
--- a/changelogs/fragments/bugfix_rule_module.yml
+++ b/changelogs/fragments/bugfix_rule_module.yml
@@ -1,0 +1,50 @@
+# https://docs.ansible.com/ansible/latest/community/development_process.html#changelogs-how-to
+bugfixes:
+    - Rule module - Now take the folder into account when checking for existing rules.
+
+# minor_changes:
+#   - Add agent role. Currently supports the vanilla agent.
+
+# known_issues:
+#   - This release is still in development and a heavy work in progress.
+#   - Discovery module is not feature complete yet.
+#   - Downtime module is not fully idempotent yet. This affects service downtimes and deletions.
+
+## Line Format
+# When writing a changelog entry, use the following format:
+
+# - scope - description starting with a lowercase letter and ending with a period at the very end. Multiple sentences are allowed (https://github.com/reference/to/an/issue or, if there is no issue, reference to a pull request itself).
+
+# The scope is usually a module or plugin name or group of modules or plugins, for example, lookup plugins. While module names can (and should) be mentioned directly (foo_module), plugin names should always be followed by the type (foo inventory plugin).
+
+# For changes that are not really scoped (for example, which affect a whole collection), use the following format:
+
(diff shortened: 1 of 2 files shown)


robin-tribe29 (approved):

geof77:
Does not work very well.

I have tried a lot of different solutions. Here is my version that is working most of the time (as far as I've tested), although there might be security issues using eval ?:

```
def get_existing_rule(module, base_url, headers, ruleset, rule):
    # Get rules in ruleset
    rules = get_rules_in_ruleset(module, base_url, headers, ruleset)
    if rules is not None:
        # Loop through all rules
        for r in rules.get("value"):
            # Check if conditions, properties and values are the same
            if (
                # these are both strings so == works
                r["extensions"]["folder"] == rule["folder"]
                # == also works for dicts 
                and r["extensions"]["conditions"] == rule["conditions"]
                and r["extensions"]["properties"] == rule["properties"]
               # per API documentation value_raw must be a valid python object
               # so just eval() both and compare
                and eval(r["extensions"]["value_raw"]) == eval(rule["value_raw"])
            ):
                # If they are the same, return the ID
                return r["id"]
    return None
```

geof77:
Thinking a bit further, what should this function really compare?
In properties, "description" and "comment" are not "functional", only "disabled" is significant.
So a rule with exactly the same folder, conditions and value_raw would have the same effect, regardless of the description and comment. I would exclude these two from the comparison.
What do you think?

geof77:
So this would give us something like:

```
def get_existing_rule(module, base_url, headers, ruleset, rule):
    # Get rules in ruleset
    rules = get_rules_in_ruleset(module, base_url, headers, ruleset)
    # Remove comment and description from comparison
    rule["properties"].pop("comment")
    rule["properties"].pop("description")
    if rules is not None:
        # Loop through all rules
        for r in rules.get("value"):
            r["extensions"]["properties"].pop("comment")
            r["extensions"]["properties"].pop("description")
            # Check if conditions, properties and values are the same
            if (
                r["extensions"]["folder"] == rule["folder"]
                and r["extensions"]["conditions"] == rule["conditions"]
                and r["extensions"]["properties"] == rule["properties"]
                and eval(r["extensions"]["value_raw"]) == eval(rule["value_raw"])
            ):
                # If they are the same, return the ID
                return r["id"]
    return None
```

geof77:
@lgetwan , @robin-tribe29 please review as the existing function breaks things and I can't get integration tests to succeed for PR #231

lgetwan:
@geoff: Please see my comment in #186: https://github.com/tribe29/ansible-collection-tribe29.checkmk/issues/186#issuecomment-1396692373

geof77:
@lgetwan : The PR introduces an error in the examples. "folder" and "location" are mutually exclusive. "folder" is deprecated and the value is picked up from location.

lgetwan:
@geof77: True! But gladly, your PR #223 fixed this. Thanks. :-)

PR #230, opened 2023-01-17
Better rule arguments parsing and default values for conditions

geof77:
Currently, if the rule "conditions" is not specified at all, empty defaults are applied. However, if some conditions are specified, default values are not applied to the other arguments. This has an impact on checking if a rule already exists (as the check is currently done), because the API returns empty arguments that are not present in the request. The user must not forget to manually specify these empty arguments. 

This PR fixes this issue, with the use of ansible argument specifications, instead of if statements.

This also implies a big update of the documentation section.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

User has to specify empty arguments for conditions.
The module uses if statements instead of built-in ansible argument checking. In fact it uses a mix.

## What is the new behavior?

The user only provides the needed arguments and does not have to care about empty ones. 
The user has better documentation and better argument checking.
The code uses ansible standard feature for argument parsing.


geof77:
Wrong branch

robin-tribe29:
@geof77 FYI: The target branch can be changed. If you cannot do that yourself, you can let me know and I will do it for you. :)

PR #233, opened 2023-01-19
Fix rules comparison for creation and deletion

geof77:
Fix problematic comparisons between the rules described in tasks and the existing rules.

The rule described is created before comparing it, so the module can use the API representation of the rule.

This temporary rule is deleted if an identical rule already exists when creating a rule, and after fetching the ID of the existing rule when deleting a rule.

Notable changes:

The comment and description parameters are omitted from the comparison, because they are not significant. Creating two identical rules with a different description or comment is not possible, and they are also ignored when deleting rules.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Correct rule comparison is nearly impossible due to differences between the API representation of the rule and the module's representation, especially for the value_raw.

Issue Number: #186

## What is the new behavior?
Rule comparison uses the API representation on both sides, so it 'always' succeeds.

### DIFF
--- a/plugins/modules/rule.py
+++ b/plugins/modules/rule.py
@@ -320,7 +320,7 @@ def create_rule(module, base_url, headers, ruleset, rule):
 
     # if existing rule, delete new rule and return existing id
     if e:
-        delete_rule(module, base_url, headers, r["id"])
+        delete_rule_by_id(module, base_url, headers, r["id"])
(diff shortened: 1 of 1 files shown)


lgetwan (approved):
I had a look at the code and tested it. Great work, @geof77!
Thanks a lot for your effort!

PR #234, opened 2023-01-20
Changelog and documentation for the improved rule module.

lgetwan:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Changelog for the new features of the rule module was missing.

## What is the new behavior?

- Added a changelog
- Added @geoff77 as an author
- Described the new behavior regarding idempotency in the module docs' notes.


changelogs/fragments/bugfix_rule_module.yml
```diff
@@ -1,6 +1,6 @@
 # https://docs.ansible.com/ansible/latest/community/development_process.html#changelogs-how-to
 bugfixes:
-    - Rule module - Now take the folder into account when checking for existing rules.
+    - Rule module - Now properly comparing the specified rule with the existing ones to achieve idempotency.
```
robin-tribe29, line 4: Maybe we want to use the `known_issues` keyword to hint at the "limitations" of the module in the current version?

plugins/modules/rule.py
```diff
@@ -91,6 +89,17 @@
...
+notes:
+    - To achieve idempotency, this module is comparing the specified rule with the already existing
+      rules based on conditions, folder, value_raw and enabled/disabled.
+    - To be able to compare the value_raw, which is internally stored in python format, the module
+      has to do a workaround: it is creating the specified rule, and then compares this rule with 
+      all existing rules.
+    - Then, in case of I(state) == absent, it will delete both rules: the specified one and the duplicate
+      found.
+    - Or, in case of I(state) == present, it will delete the new rule, if a duplicate is already there,
+      and keep it if not.
```
robin-tribe29, line 102: I like the explanation, I think it reads rather well.
We might want to add, that as a side effect, changes are created in Checkmk.

changelogs/fragments/bugfix_rule_module.yml
lgetwan: Done.

plugins/modules/rule.py
lgetwan: Also done.

PR #237, opened 2023-01-23
Fix password for activation

tonk:
In the activation handler for the agent the password variable `checkmk_agent_pass` is used.
In the `vars/main.yml` the credentials are assembled to `checkmk_agent_auth` and this the one that is used in the rest of
the code.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The activation of the host breaks the playbook in the handler part, as the variable `checkmk_agent_pass` is undefined.

## What is the new behavior?

The host is activated in the CheckMK server


robin-tribe29 (approved):
Good catch, thanks!

Will merge this later.

Issue #153, opened 2022-09-15
[BUG] Rule Module not correct?


elwood218:
Hi, we have tried the rule module created by @diademiemi . First of all thanks for that work! But we have tested and debugged it and we aren't sure if that is working correctly. Maybe we are missing a point.

If you get for example `r["extensions"]["conditions"]` it only got the keys but not the values for comparing.
Also if the `value_raw` is sorted it not really sorts the keys, instead it sorts every character. Maybe the last is working anyway kinda but the other are not working.

We have debugged it with adjusting the following function (reason for debug was to find out if the API output was changed):

```
def get_existing_rule(module, base_url, headers, ruleset, rule):
    # Get rules in ruleset
    rules = get_rules_in_ruleset(module, base_url, headers, ruleset)
    a_list = []
    b_list = []
    if rules is not None:
        # Loop through all rules
        for r in rules.get("value"):
    #        rule_exist = sorted(r)
    #        rule_defined = sorted(rule["properties"])
    #        exit_failed(module, "Rule_exist: %s, Rule_defined: %s , " % (rule_exist, rule_defined),)
            foo = sorted(r["extensions"]["conditions"])
            bar = sorted(rule["conditions"])
            a_list.append(foo)
            b_list.append(bar)
            # Check if conditions, properties and values are the same
            if (
                sorted(r["extensions"]["conditions"]) == sorted(rule["conditions"])
    #            and sorted(r["extensions"]["properties"]) == sorted(rule["properties"])
                and sorted(r["extensions"]["value_raw"]) == sorted(rule["value_raw"])
            ):
                # If they are the same, return the ID
                return r["id"]
        exit_failed(module, "Rule_exist: %s, Rule_defined: %s , " % (a_list, b_list),)
    return None
```

PS: I also wonder why the json output was choosed with `()`.. This is not really the standard way like other tools how jq could read it by default.

diademiemi:
What version of the Checkmk server are you using?

elwood218:
2.1.0p11

diademiemi:
The API output has changed in 2.1.0p10. What exactly does the rule module not do? Is the idem potency not correct?

elwood218:
As I already wrote that only the keys are compared and not the values.

diademiemi:
I'm busy for a few days but I'll have a look at this in the weekend and resolve it

elwood218:
[...]

PR #241, opened 2023-01-24
Use safe_eval instead of temporary rules for comparison

geof77:
In PR #233 we create temporary rules as a means to get the exact API representation for the value_raw.

Ansible's module_utils library provides a "safe_eval()" implementation of ast.literal_eval(). As it was suggested by a developer in #153 we could use this technique to evaluate the values as Python objects before the comparison, as other modules and checkmk api does.

This removes the need to create a temporary rule.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Create temp rule
Fill the logs
Slow

Issue Number: #153 #186

## What is the new behavior?
Use safe_eval from ansible module_utils
Does not fill the logs
Faster

### DIFF
--- a/plugins/modules/rule.py
+++ b/plugins/modules/rule.py
@@ -224,8 +224,8 @@ id:
 import json
 
 from ansible.module_utils.basic import AnsibleModule
-from ansible.module_utils.urls import fetch_url
 from ansible.module_utils.common.validation import safe_eval
+from ansible.module_utils.urls import fetch_url
 
 try:
     from urllib import urlencode
--- a/plugins/modules/rule.py
+++ b/plugins/modules/rule.py
@@ -458,7 +458,6 @@ def run_module():
 
     module = AnsibleModule(argument_spec=module_args, supports_check_mode=False)
 
-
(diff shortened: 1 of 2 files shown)


lgetwan:
This looks promising!

plugins/modules/rule.py
```diff
@@ -74,7 +74,7 @@
                 description: Properties of the rule.
                 type: dict
             value_raw:
-                description: Rule values as exported from the UI.
+                description: Rule values as exported from the GUI.
```
robin-tribe29, line 77: The term "GUI" is generally deprecated or discouraged. We should stick with "UI", or "web interface".

geof77: Looks like I had made a fortunate typo :)

robin-tribe29: Yeah, it really is a nitpick, but I probably would have changed it back some time in the future. :sweat_smile:

lgetwan (approved):
Successfully tested

PR #242, opened 2023-01-25
Added configuration setting for force_foreign_changes

tonk:
For the CheckMK agent we need the option `force_foreign_changes` to be `true`. In the code it is set to `false` and isn't configurable. This PR make this setting a variable, so you can choose what you like.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Currently the `force_foreign_changes` cannot be configured.

## What is the new behavior?

In this version it can be set to `true` or `false`


robin-tribe29:
I struggled with this one for some time and I came to the conclusion, that it is safer to not force foreign changes.
The rationale her is, that adding a host (or several) should not activate other potentially unfinished changes.
I do realize, that making it configurable seems beneficial, but I am still uncertain. Let me discuss this with @lgetwan and @NicolasVongeheur. Maybe other reading along care to share their opinion as well. :v:

lgetwan:
I would go for the "configurable" option. There might be customers (e.g. my former company) who tend to do things completely automated, and they would need a "force_foreign_changes" for that.
If there's a big exclamation mark in the docs, we're on the safe side, I'd say.

NicolasVongeheur (approved):
In my opinion, it is up to the users if he want the foreign changes activated or not. Giving them the option and a big warning actually looks pretty good to me

Issue #243, opened 2023-01-25
[BUG] downtime does not accept vars


AngelNev:
Verify first that your issue is not already reported [here](https://github.com/tribe29/ansible-collection-tribe29.checkmk/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc).
Where possible also test if the latest release and main branch are affected too.
Complete all sections as described!

**Describe the bug**
Using a variable from Jenkins and Ansible gives a error, saying "unsupported type for timedelta minutes component: str"
Using a fix int as in the example works. I think, the actual Pyhton script does not accept strings or should transform/assure it is an integer.

**Component Name**
`downtime`

**Ansible Version**
```console
ansible 2.9.27
```
**Checkmk Version**
```console
2.1.0p16
```
**Collection Version**
```console
0.16.1
```
**Environment**
RHEL8

**To Reproduce**
```console
- set_fact:
     START_AFTER_MINUTES: '{{ START_AFTER_MINUTES | int }}'

- debug:
    msg: "The 'START_AFTER_MINUTES' variable is NEW a {{ START_AFTER_MINUTES | type_debug }}"

- name: "Schedule host downtime."
      downtime:
        server_url: "{{ server_url }}"
        site: "{{ site }}"
        automation_user: "{{ automation_user }}"
        automation_secret: "{{ automation_secret }}"
        comment: "{{ COMMENT }}"
        host_name: "{{ SERVERNAME }}"
        start_after:
          minutes: "{{ START_AFTER_MINUTES }}"
        end_after:
          hours: 2
        state: "present"
        validate_certs: false
      delegate_to: localhost
```
**Expected behavior**
The host goes into maintenance.

**Actual behavior**
unsupported type for timedelta minutes component: str

robin-tribe29:
Hi @AngelNev and thanks for your report.
Can you provide the exact error message, that you receive? Thanks!

AngelNev:
Hi Robin, sure and sorry, yesterday I was in a hurry:

TASK [Schedule host downtime.] *************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: unsupported type for timedelta minutes component: str
[...]

PR #244, opened 2023-01-27
module downtimes: better parameter handling

lgetwan:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

#243 
The parameters start_after and end_after do not accept string values, which might sometimes be necessary, when variables are used.

## What is the new behavior?

- Strings are automatically being converted to integers.

### DIFF
--- a/changelogs/fragments/downtime.yml
+++ b/changelogs/fragments/downtime.yml
@@ -1,0 +1,4 @@
+# https://docs.ansible.com/ansible/latest/community/development_process.html#changelogs-how-to
+
+bugfixes:
+  - Downtime module - Fix handling of parameters start_after and end_after. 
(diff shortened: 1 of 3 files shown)


robin-tribe29:
@ogaida for your information, if you want to take a look. :)

I will take care of the sanity checks. :muscle:

ogaida:
why do you ask? python2 is not supported at all.

robin-tribe29 (approved):
Looks fine by me.

We will take a brief break from merges, but will be back within two weeks.

robin-tribe29:
> why do you ask? python2 is not supported at all.

After your input on changing some else's code without a word, I felt like it was appropriate to give you a chance to comment. If this change is good for you, that's great. :)

Regarding Python: Ansible 2.11 and 2.12 still support Python 2.6, hence we want to stay compatible. Of course the language version itself is end of life.

PR #246, opened 2023-01-29
[host module] also manage cluster hosts

elwood218:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

So far there only "normal" hosts can be added.

Issue Number: `e.g. #1`

## What is the new behavior?

- Create cluster hosts
- Update nodes of cluster host

## Other information

I have commented out a part in the code of changing a host to a cluster and vice versa. Maybe that function shall be used too? But for now it is a deleting the host an re-create. Can that be done in another way?

### DIFF
--- a/plugins/modules/host.py
+++ b/plugins/modules/host.py
@@ -4,6 +4,9 @@
 # Copyright: (c) 2022, Robin Gierse <robin.gierse@tribe29.com>
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 from __future__ import absolute_import, division, print_function
+from ansible.module_utils.urls import fetch_url
+from ansible.module_utils.basic import AnsibleModule
+import json
 
 __metaclass__ = type
 
@@ -155,11 +158,6 @@ message:
     sample: 'Host created.'
 """
 
-import json
-
-from ansible.module_utils.basic import AnsibleModule
-from ansible.module_utils.urls import fetch_url
-
 
 def exit_failed(module, msg):
     result = {"msg": msg, "changed": False, "failed": True}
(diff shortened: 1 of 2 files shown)


robin-tribe29:
Hi @elwood218 and thank you for your contribution, it is very much appreciated! :pray: 
Can we get the tests nice and green? Then we will review your changes and prepare for merging, if there are no problems.
We are in a brief freeze of merging, but will continue within a week.

plugins/modules/host.py
```diff
@@ -150,7 +213,14 @@ def get_current_host_state(module, base_url, headers):
...
 
-    return current_state, current_explicit_attributes, current_folder, etag
+    return (
+        current_state,
+        current_explicit_attributes,
+        current_folder,
+        current_cluster,
+        current_cluster_nodes,
+        etag,
+    )
```
lgetwan, line 223: Wouldn't it be better to return "cluster" as current_state instead of returning two separate variables?

lgetwan (changes requested):
Good idea to add that functionality!
I already successfully tested your modifications.
There's only one comment and one small change request below.

plugins/modules/host.py
elwood218: Don't know exactly what you mean. Do you mean "cluster.nodes" and "cluster.is_cluster" ?

lgetwan: No, I meant: In case of a cluster, you could return "cluster" in the variable "current_state" instead of introducing a new variable "current_cluster".

lgetwan: Hi @elwood218, 
as soon as you have removed the (IMHO) unneeded variable "current_cluster" from line 220, we can merge the PR into the devel branch.
Can you please check?

meni2029:
Hello @elwood218, it seems it was very close, why did you give up on this good contribution ?

PR #247, opened 2023-01-29
add module for adding passwords

elwood218:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Passwords cannot be added.

## What is the new behavior?

- Passwords can be added or deleted to/out of the password store.
- If passwords exists nothing will be done.

## Other information

If I would knew how to read passwords if I am logged in on the Checkmk server then it would be possible to compare passwords and change them. For now it only would be possible to create a password and could think about to force password set to overwrite it nevertheless if it differs or not.
The API lets only set "editable by" to a contact group the editor is part of. So as user "automation" I can only set "admin" if automation would be Administrator.

### DIFF
--- a/plugins/modules/password.py
+++ b/plugins/modules/password.py
@@ -3,6 +3,7 @@
 # Copyright: (c) 2018, Mathias Buresch <mathias.buresch@de.clara.net>
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 from __future__ import (absolute_import, division, print_function)
+
 __metaclass__ = type
 
 DOCUMENTATION = r'''
@@ -85,8 +86,8 @@ message:
 import json
 import re
 
-from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils.urls import fetch_url
+from ansible.module_utils.basic import AnsibleModule
 
 
 def exit_failed(module, msg):
@@ -112,7 +113,8 @@ def get_password(module, base_url, headers, norm_id):
     api_endpoint = "/objects/password/" + module.params.get("id", norm_id)
     url = base_url + api_endpoint
 
-    response, info = fetch_url(module, url, data=None, headers=headers, method="GET")
+    response, info = fetch_url(
+        module, url, data=None, headers=headers, method="GET")
 
     if info["status"] == 200:
         body = json.loads(response.read())
@@ -181,7 +183,8 @@ def delete_password(module, base_url, headers, norm_id):
     api_endpoint = "/objects/password/" + module.params.get("id", norm_id)
     url = base_url + api_endpoint
 
-    response, info = fetch_url(module, url, data=None, headers=headers, method="DELETE")
+    response, info = fetch_url(
+        module, url, data=None, headers=headers, method="DELETE")
 
     if info["status"] != 204:
         exit_failed(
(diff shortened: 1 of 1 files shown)


lgetwan (changes requested):
I successfully tested your new module. Thanks for contributing!

I like Your idea to provide a "force new password". What do you think about doing it like this:

- present: creates a pw, if absent and doesn't do anything if the pw is already present.
- absent: deletes a pw if existing
- reset: resets the "password" value if the pw exists. If the pw doesn't exist, it exits with "failed"

elwood218:
@lgetwan Hi. What do you mean by "reset"? For me reset more sounds like setting a changed password back to the previous. 
Btw is there a way to read password on cli?

lgetwan:
@elwood218: There might be better words than "reset", true.
I guess, there _is_ a way to decrypt the PW store on the CLI. I haven't tried, yet. Maybe, having a close look at ~/lib/check_mk/utils/password_store.py and encryption.py helps.Also checkout https://checkmk.com/werk/13633

elwood218:
@lgetwan Yes I already had a look into both you mentioned. Thought or hoped there is a CLI command for that. But then first let us implement like how it is now and later think of the other way.

plugins/modules/password.py
```diff
@@ -0,0 +1,282 @@
...
+description: Manage passwords in the Checkmk password store.
+
+options:
+    name:
+        description:
+            - Name/title of the password.
+        required: true
+        type: str
+        aliases: [title]
+    ident:
```
robin-tribe29, line 28: In the `tag_group` module we use "id". I think we should do that here too.
I am aware, that the API calls it "ident", but I think "id" is better understood.

plugins/modules/password.py
```diff
@@ -0,0 +1,282 @@
...
+# Delete a password.
+- name: "Delete a password."
+  tribe29.checkmk.host:
+    server_url: "http://localhost/"
+    site: "my_site"
+    automation_user: "automation"
+    automation_secret: "$SECRET"
+    name: "my-password"
+    state: "absent"
+"""
```
robin-tribe29, line 78: I think in both examples the "password" key is missing, right?

robin-tribe29 (commented):
Is there a reason, that you do not deal with the "shared" key from the API?

lgetwan:
Ok, my wording "reset password" was probably a bad choice. What I meant:

1. state = "absent": Delete password entry if existing.
2. state = "present": Create password entry if not existing. If existing, do NOT touch the actual password, even if the parameter is set.
3. state = "reset": Create password entry if not existing. If existing, set the actual password to the value of the password parameter.

What wording would you suggest instead of "reset"? "overwrite" or "force_new_password" might be alternatives, but I think, I'm not on the peak of my creativity, today. :-p

elwood218:
@lgetwan I would say "force" or "overwrite" because "force" for example is also used in other similar Ansible modules if I remember correct.

robin-tribe29:
Is there a perceivable use case, where you want to use "present" to ensure a password is there without changing anything? I feel like the answer here is "no". So I would expect, that "present" ensures the password I want to set is actually set, without caring about a potentially existing password.

I feel like "reset" is still more confusing, although I get your idea @lgetwan. But I also feel like "force" should not be necessary. If there is a use case for my former example, then I think we should go with "present", "absent" and "force".

P.S.: @elwood218 you marked my comments as resolved, but I see no new commits, nor a comment from you. Do you just need to push the changes?

PR #248, opened 2023-01-29
Fix become for service discovery in agent role

thorian93:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently, the task `"Discover services and labels on host."` fails after a long time, when `become: true` is set on a high level. This is due to a sudo timeout.

## What is the new behavior?

The aforementioned task now explicitly sets `become: false`, thus avoiding the timeout error.


robin-tribe29 (approved):
This is good for merge.

Not sure why the `automation_secret` is changed, but hat should sort itself out later.

PR #249, opened 2023-02-02
change 'is false' to '== false'

darrindg91:
Was getting an error: 
FAILED! => {"msg": "The conditional check 'checkmk_server_allow_downgrades | bool is false\nand item.stdout is version(item.item.version, '>')\n' failed. The error was: template error while templating string: no test named 'false'.  so replaced 'is false' to '== false'

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When trying to upgrade server versions, I see an error:

FAILED! => {"msg": "The conditional check 'checkmk_server_allow_downgrades | bool is false\nand item.stdout is version(item.item.version, '>')\n' failed. The error was: template error while templating string: no test named 'false'. 

## What is the new behavior?

The upgrade now works:

Pausing for 5 seconds
(ctrl+C then 'C' = continue early, ctrl+C then 'A' = abort)
[tribe29.checkmk.server : Short pause for patch update.]
Proceeding with updating site test from version 2.1.0p17.cre to version 2.1.0p20.cre.
This is a minor patch update. A backup will be created in /tmp. This can take a while! The site will be down during the update

-
-
-

### DIFF
--- a/roles/server/tasks/update-site.yml
+++ b/roles/server/tasks/update-site.yml
@@ -19,7 +19,7 @@
     msg: |
       Downgrading is not allowed since checkmk_server_allow_downgrades is set to false.
   when:
-    checkmk_server_allow_downgrades is defined 
+    checkmk_server_allow_downgrades is defined
     and not checkmk_server_allow_downgrades | bool
     and item.stdout is version(item.item.version, '>')
 
--- a/roles/server/tasks/update-site.yml
+++ b/roles/server/tasks/update-site.yml
@@ -19,7 +19,7 @@
     msg: |
       Downgrading is not allowed since checkmk_server_allow_downgrades is set to false.
   when:
-    checkmk_server_allow_downgrades | bool is false
+    checkmk_server_allow_downgrades | bool == false
     and item.stdout is version(item.item.version, '>')
 
 - name: "Short pause for patch update."
--- a/roles/server/tasks/update-site.yml
+++ b/roles/server/tasks/update-site.yml
@@ -19,7 +19,8 @@
     msg: |
       Downgrading is not allowed since checkmk_server_allow_downgrades is set to false.
   when:
-    checkmk_server_allow_downgrades | bool == false
+    checkmk_server_allow_downgrades is defined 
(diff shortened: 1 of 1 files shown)


roles/server/tasks/update-site.yml
```diff
@@ -19,7 +19,7 @@
     msg: |
       Downgrading is not allowed since checkmk_server_allow_downgrades is set to false.
   when:
-    checkmk_server_allow_downgrades | bool is false
```
msekania, line 22: Actually 
`checkmk_server_allow_downgrades | bool`
is sufficient

msekania: please, do not use `== false`

darrindg91: Wouldn't that be equivalent to 'is true'? Perhaps:
`not checkmk_server_allow_downgrades`
or something similar?

robin-tribe29: Jup, I think that is the best syntax. But do not forget `| bool` at the end.

msekania: should read

`not checkmk_server_allow_downgrades | bool`

darrindg91: @msekania 

`checkmk_server_allow_downgrades` is defined as false in the defaults of the role, so as long as this is the case this should not be an issue.

robin-tribe29: Yes, correct, thanks for catching that @msekania!
@darrindg91 can you update that line to `checkmk_server_allow_downgrades is defined and not checkmk_server_allow_downgrades | bool`? Then we are good to go.

darrindg91: @robin-tribe29 

I updated for the not defined case and the corrected 'is false' syntax.

robin-tribe29: Great, the logic looks good to me!
If you could fix the linting finding from the tests, we should be good to merge.

robin-tribe29 (approved):

Issue #181, opened 2022-10-25
[FEED] Check mode for rules


adamsilkcm:
**Provide your feedback**
It would be useful to have this show planned changes in ansible's `--check` mode.

**Component Name**

rule module

**Ansible Version**
independent

**Collection Version**

```console
$ ansible-galaxy collection list
Collection          Version
------------------- -------
tribe29.checkmk     0.11.0

```

robin-tribe29:
Hi Adam and thanks for the suggestion! This would for sure be a very useful feature.
We are very limited on resources however, so this is gonna end up at the end of a pile of work.
I labeled the issue as "help wanted", so maybe someone else can implement this earlier, than we can.

If there is any news from our side, we will update here.

geof77:
I have succesfully tested a basic implementation of test mode. However, it depends on PR #241 because we currently create a temporary rule to do the comparison. I think this is incompatible with test mode where we should not do any modification to the system.

robin-tribe29:
#241 was merged and will be released probably next week. Looking forward to your check mode implementation @geof77! :hugs:

robin-tribe29:
Check mode is implemented in #258.

PR #258, opened 2023-02-03
Feature rule check mode

geof77:
Check mode support for rules module.

In check mode, the module will only return what modifications it would do, without actually making any modificiation.

Suggestions welcome.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

No check mode

Issue Number: #181

## What is the new behavior?

Allow ansible's --check parameter.

### DIFF
--- a/plugins/modules/rule.py
+++ b/plugins/modules/rule.py
@@ -299,6 +299,9 @@ def create_rule(module, base_url, headers, ruleset, rule):
     if e:
         return (e["id"], not changed)
 
+    if module.check_mode:
+        return (None, changed)
+
     params = {
         "ruleset": ruleset,
         "folder": rule["folder"],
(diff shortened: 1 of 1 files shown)


plugins/modules/rule.py
```diff
@@ -388,6 +392,9 @@ def move_rule(module, base_url, headers, rule_id, location):
 
     url = base_url + api_endpoint
 
+    if module.check_mode:
```
lgetwan, line 395: When doing a "create rule on top of ruleset", the move_rule() gets a "None" for the rule_id and fails when setting the api_endpoint.
Also, the "get_rule_etag()" fails in that case, because of the invalid rule_id.

lgetwan (changes requested):
Great idea to enable the check mode! Thanks!
I just found one case (when move_rule() is involved) where it fails in check mode.

plugins/modules/rule.py
geof77: I reverted my last commit, that should fix these.

geof77:
Just added a message that we are in check mode, like this:
```
TASK [cmkserver : Test rule from collection] ***************************
changed: [ilx00073] => {"changed": true, "id": null, "msg": "Rule created (check_mode: no changes made)"}
```

lgetwan (approved):

PR #259, opened 2023-02-06
Fix galaxy import errors

lgetwan:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

The galaxy import test automation was complaining about variable naming and usage.

## What is the new behavior?

- Fixed the errors in the modules rule, host_group, contact_group, service_group, downtime and discovery
- The module tag_group still contains errors

### DIFF
--- a/plugins/modules/downtime.py
+++ b/plugins/modules/downtime.py
@@ -442,13 +442,6 @@ def run_module():
         module.params.get("site"),
     )
 
-    # # Decide if host- or service-downtime depending on elements of service_descriptions
-    # service_descriptions = module.params.get("service_descriptions")
-    # if service_descriptions == []:
-    #     hostdowntime = True
-    # else:
-    #     hostdowntime = False
-
     # here, we need only the state param
     state = module.params.get("state", "present")
 
--- a/plugins/modules/downtime.py
+++ b/plugins/modules/downtime.py
@@ -442,12 +442,12 @@ def run_module():
         module.params.get("site"),
     )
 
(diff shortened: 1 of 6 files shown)


plugins/modules/downtime.py
```diff
@@ -442,12 +442,12 @@ def run_module():
...
-    if service_descriptions == []:
-        hostdowntime = True
-    else:
-        hostdowntime = False
+    # # Decide if host- or service-downtime depending on elements of service_descriptions
+    # service_descriptions = module.params.get("service_descriptions")
+    # if service_descriptions == []:
+    #     hostdowntime = True
+    # else:
+    #     hostdowntime = False
```
robin-tribe29, line 450: Why do you comment this out instead of removing it altogether?
Is here a ToDo left? Then we should note that.

plugins/modules/rule.py
```diff
@@ -399,7 +398,7 @@ def move_rule(module, base_url, headers, rule_id, location):
             % (info["status"], info["body"]),
         )
 
-    r = json.loads(response.read().decode("utf-8"))
+    # r = json.loads(response.read().decode("utf-8"))
```
robin-tribe29, line 401: Same as above: Can we remove it completely, or is there something left to do?

lgetwan: Done!

plugins/modules/downtime.py
lgetwan: Done!

robin-tribe29 (approved):

PR #264, opened 2023-02-08
Allow firewall zone change

### LINKED ISSUE #263: [FEED] allow firewall zone change on EL 7/8
**Provide your feedback**

RHEL firewall configuration should allow to specify a concrete zone - defaulting to 'public'.

**Component Name**

role `agent` os specific part.

**Environment**

os type: EL 7/8

flybyray:
fixes #263

## Pull request type

- [X] Feature

On EL the zone is not set and defaults to 'public'.

Issue Number: #263 

## What is the new behavior?

An additional variable allows to override zone.

### DIFF
--- a/roles/agent/defaults/main.yml
+++ b/roles/agent/defaults/main.yml
@@ -20,6 +20,7 @@ checkmk_agent_discover: 'false'
 checkmk_agent_update: 'false'
 checkmk_agent_tls: 'false'
 checkmk_agent_configure_firewall: 'true'
+# checkmk_agent_configure_firewall_zone: 'public'
 checkmk_agent_force_install: 'false'
 checkmk_agent_prep_legacy: 'false'
 checkmk_agent_delegate_api_calls: localhost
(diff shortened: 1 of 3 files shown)


roles/agent/defaults/main.yml
```diff
@@ -20,6 +20,7 @@ checkmk_agent_discover: 'false'
 checkmk_agent_update: 'false'
 checkmk_agent_tls: 'false'
 checkmk_agent_configure_firewall: 'true'
+# checkmk_agent_configure_firewall_zone: 'public'
```
robin-tribe29, line 23: When we put the default here, I think we can ditch the default in `RedHat.yml`.
I try to avoid commented options in the defaults, as it feel not that elegant.

robin-tribe29 (changes requested):
Can you also please elaborate a little here in the PR description, what the firewall zones mean and what impact this option can have? It might make sense to put a few words in the README, so someone uncertain about the implications of changing this, can get an idea.

roles/agent/defaults/main.yml
robin-tribe29: Fixed it myself.

robin-tribe29 (approved):

PR #277, opened 2023-02-23
New Module "user"

lgetwan:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

There was no module for user management

## What is the new behavior?

- 90% of the features of the user related endpoints of the REST API are implemented
- Missing: the parameter "interface_options" is not yet usable 
- The module is idempotent in general, but:
- The password of a user is only set when the user was previously absent or if the state "reset_password" is used.


lgetwan:
To do: 

- Test the module more excessively
- Write an integration test
- Write the changelog

lgetwan:
The new module is now ready to be reviewed, @robin-tribe29

plugins/modules/user.py
```diff
@@ -0,0 +1,514 @@
...
+    module_args = dict(
+        server_url=dict(type="str", required=True),
+        site=dict(type="str", required=True),
+        validate_certs=dict(type="bool", required=False, default=True),
+        automation_user=dict(type="str", required=True),
+        automation_secret=dict(type="str", required=True, no_log=True),
+        name=dict(required=True, type="str"),
+        fullname=dict(type="str"),
+        password=dict(type="str", no_log=True),
+        enforce_password_change=dict(type="bool"),
```
robin-tribe29: Ansible shows a warning `[WARNING]: Module did not set no_log for enforce_password_change` because we do not set the `no_log` parameter here. However, this is not really an issue, because Ansible just guesses, that this field might contain sensitive information, which it does not.
According to [this](https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#argument-spec), we have to set `no_log=False` explicitly to get rid of the warning.
A quick fix and test on my end verified this.

robin-tribe29 (approved):

plugins/modules/user.py
robin-tribe29: I fixed it myself. If there is still a problem we can patch it later.

Issue #176, opened 2022-10-12
[FEED] Close similarity between modules


robin-checkmk:
> Hi @robin-tribe29,
> 
> somehow this close similarity of contact and host groups does not give me a rest.
> I have also checked and service group rest API interface is also quite similar.
> I do not know, whether this similarity will also stay in the future, but in case, I have a group module written that combines all this three group management in one (including test suite).
> 
> https://github.com/msekania/ansible-collection-tribe29.checkmk/tree/feature-groups
> 
> In current version, there is a group_type field which takes one of ["contact", "host", "service"] values, and the rest is provided either by group_name (instead of host_group_name,  contact_group_name) and title or list of dict-s named 'groups' (instead of host_group,  contact_group) with name and title entries.
> I can also add a backward compatibility features.
> 
> Alternatively, I can rewrite the module so that group_type field is removed and choice is made by providing:
> - host_group_name or host_groups;
> - contact_group_name or contact_groups;
> - service_group_name or service_groups;
> 
> Should I make an another pull request?
> We can decide then how to proceed.
> 
> Best,
> Michael

_Originally posted by @msekania in https://github.com/tribe29/ansible-collection-tribe29.checkmk/issues/168#issuecomment-1274386559_

# Porting Checklist
- [ ] activation
- [ ] contact_group
- [ ] discovery
- [ ] downtime
- [ ] folder
- [ ] host_group
- [ ] host
- [ ] rule
- [ ] service_group
- [ ] tag_group
- [x] user

lgetwan:
I don't think, that a similarity regarding the code justifies merging those three things together, because the actual items that are created, are of different kinds.
And it wouldn't simplify the usage of the Ansible module  from my point of view.
What we could do (and should do in the future) is creating a file utils.py that contains all the code that is duplicate across the modules. That would make it easier to add new modules.

msekania:
Hi Lars,

> And it wouldn't simplify the usage of the Ansible module from my point of view.
I agree that it does not simplify Ansible part.
It might also get problematic if REST API gets more complex and different for each group. With separate modules one can develop each part then separately. 

> What we could do (and should do in the future) is creating a file utils.py that contains all the code that is duplicate across the modules. That would make it easier to add new modules.

It's definitely a good idea, because there are several functions that repeat cross the modules not only along the groups.

What one can do for the moment, one can also add service groups module together with contact_groups or as a separate pull request.
[...]

PR #291, opened 2023-03-26
Introduce a centralized library to call the Checkmk API

godspeed-you:
The code to contact the Checkmk API is redundantly placed all over the modules. This projects aims to centralize this code into a single library to make our lifes more easy. Later changes may be done in one single place instead of changing several dozens of code lines.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe):

- Currently, if the API headers or other static information is changed, this needs to be adapted in an unclear amount of files and code lines. 
- Everytime a new API call to Checkmk is added, bugs or slightly different behaviour may be introduced.

## What is the new behavior?

- The centralized file may be called by every module to have a defined set of functionality returned information.
- This set may be used to derive a module specific class to address specific requirements.
- The result of this API call is stored in a defined and clearly structured data set. This allows to automatically provide a clear and reliable set of information after the API call.

## Other information

This PR contains a migrated module as an example on how to use this new library. The changes have been reduced to an absolute minimum to minimize potential bugs. So potential further refactorings may be skipped to have this change less complex.

### DIFF
--- a/plugins/modules/activation.py
+++ b/plugins/modules/activation.py
@@ -15,7 +15,7 @@ short_description: Activate changes in Checkmk.
 
 # If this is part of a collection, you need to use semantic versioning,
 # i.e. the version is of the form "2.5.0" and not "2.4".
-version_added: "0.0.2"
+version_added: "0.0.1"
 
 description:
 - Activate changes within Checkmk.
@@ -83,7 +83,7 @@ import time
 
 from ansible.module_utils.basic import AnsibleModule
 
-from ..module_utils.api import CheckmkAPI
+from ..module_utils.api import CheckmkAPI, result_dict
 
 HTTP_CODES = {
     # http_code: (changed, failed, "Message")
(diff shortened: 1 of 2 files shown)


lgetwan (commented):
I really like the way you unified and simplified the API calls!
I have a question regarding the definition of the inheriting class, though. In some modules, there are multiple, different POST or GET calls. How would you define the related 
`def post(self)`
calls for those?

msekania:
related #176

msekania:
@godspeed-you,

First of all thanks,

Concerning module, in essence it should at least cover all already implemented usage in different modules, otherwise it will be yet another modules which one has to adjust manually per module and not the central one.

lgetwan:
I agree to update the modules one by one. And I agree, it's probably easier to have separate feature branches for each module change.
Thanks for updating the user module, as well!

msekania:
@godspeed-you,

when structure is fixed and accepted, I will also help with group modules.

plugins/modules/activation.py
```diff
@@ -15,7 +15,7 @@
 
 # If this is part of a collection, you need to use semantic versioning,
 # i.e. the version is of the form "2.5.0" and not "2.4".
-version_added: "0.0.1"
+version_added: "0.0.2"
```
robin-tribe29, line 18: Why did you change this? This reflects the version of the collection the module was introduced to. There is no independent versioning of modules.

robin-tribe29 (changes requested):

plugins/modules/activation.py
godspeed-you: My bad. I overlooked the name of the variable and only read the comment. This is resolved by the last commit.

lgetwan (approved):
Looks fine!

robin-tribe29 (approved):

Issue #290, opened 2023-03-24
[BUG] module "rule" issues with rules that does not return parameter "properties.disabled"


msekania:
**Describe the bug**

Some of the rules do not return the "properties.disabled" parameter, but the module checks this and results in an error.

e.g.:
```json
"extensions": {
    "conditions": {
        "host_labels": [],
        "host_tags": [],
        "service_description": {
            "match_on": [
                "Check_MK HW/SW Inventory$"
            ],
            "operator": "one_of"
        },
        "service_labels": []
    },
    "folder": "/",
    "folder_index": 0,
    "properties": {
        "description": "Restrict HW/SW-Inventory to once a day"
    },
    "ruleset": "extra_service_conf:check_interval",
    "value_raw": "1440.0"
},
"extensions": {
    "conditions": {
        "host_labels": [
            {
                "key": "cmk/docker_object",
                "operator": "is",
                "value": "container"
            }
        ],
        "host_tags": [],
        "service_labels": []
    },
    "folder": "/",
    "folder_index": 0,
    "properties": {
        "description": "Make all docker container host states base on the \"Docker container status\" service"
    },
    "ruleset": "host_check_commands",
    "value_raw": "('service', 'Docker container status')"
},
"extensions": {
    "conditions": {
        "host_labels": [],
        "host_tags": [
            {
                "key": "criticality",
                "operator": "is_not",
                "value": "offline"
            }
        ],
        "service_labels": []
    },
    "folder": "/",
    "folder_index": 0,
    "properties": {
        "description": "Do not monitor hosts with the tag \"offline\""
    },
    "ruleset": "only_hosts",
    "value_raw": "True"
},
"extensions": {
    "conditions": {
        "host_labels": [],
        "host_tags": [],
        "service_labels": []
    },
    "folder": "/",
    "folder_index": 0,
    "properties": {
        "description": "Perform every two hours a service discovery"
    },
    "ruleset": "periodic_discovery",
    "value_raw": "{'check_interval': 120.0, 'severity_unmonitored': 1, 'severity_vanished': 0, 'severity_new_host_label': 1}"
},
"extensions": {
    "conditions": {
        "host_labels": [],
        "host_tags": [
            {
                "key": "snmp",
                "operator": "is",
                "value": "snmp"
            },
            {
                "key": "snmp_ds",
                "operator": "is_not",
                "value": "snmp-v1"
            }
        ],
        "service_labels": []
    },
    "folder": "/",
[...]

PR #296, opened 2023-03-31
Fix rule module for #290

geof77:
Fix for #290
## Pull request type
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Crash when the API does not return a value for the 'disabled' property

Issue Number: #290

## What is the new behavior?
Set a default value for the disabled property when comparing


lgetwan (approved):
Looks fine to me!

PR #297, opened 2023-03-31
Update CI Scenarios.

robin-tribe29:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

Our CI still checks the collection against Ansible 2.11 and Python 2.7. Both of which are EOL.

## What is the new behavior?

Support for Ansible 2.11 and Python 2.7 is removed, while support for Ansible 2.14 is added.
This does not mean, that the collection does not work anymore with older versions. We simply do not test against them anymore.


godspeed-you:
There are two changes mixed together: Version changes and the renaming of `name` into `"`host_name` (which is deprecated, right?). Is this intended?

robin-tribe29:
The renaming is a fix of an issue introduced by bulk renaming. In these cases (`discovery` and `downtime` module) we actually need `host_name`.

godspeed-you (approved):
Looks fine for me. As far as I can follow. Beside the fact, that a fix and the version declaration is mixed.

PR #298, opened 2023-03-31
introduce utils and types in module_utils

godspeed-you:
Next step to centralize and standardize commonly used functionality

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

There are several commonly used functionality in this code base, that is everytime slightly different handled. This leads to unpredictable behaviour between the modules and more space for bugs.  

## What is the new behavior?

This step tries to centralize specific types (e.g. the result of an API call) and commonly used functionality (e.g. a standardized way of returning the result in Ansible or a generic set of http results). 
Especially for newer Python versions, this enables a better way to define types and reduce eventually that way bugs on code level. As typing requires Python 3.5 or newer, this is more a preparation for future improvements. 

## Other information

A migration of the user module is already using this change.

### DIFF
--- a/plugins/module_utils/api.py
+++ b/plugins/module_utils/api.py
@@ -11,11 +11,10 @@ from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
 from ansible.module_utils.urls import fetch_url
-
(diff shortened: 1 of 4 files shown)


lgetwan (approved):
Tested successfully!

PR #301, opened 2023-04-01
Fix #175

### LINKED ISSUE #175: [BUG] Unexpected host-update behavior
Unfortunately I was negatively surprised today. I wanted to set a new host attribute and unfortunately lost all the remaining attributes that were already set.

If I see this correctly, in the case of an update I currently have to load all existing attributes first, then enrich them with the new attributes and then process them with tribe29.checkmk.host.

I find this procedure to be extremely cumbersome and extremely inflates the ansible playbooks. I would expect the module to recognize and preserve the attributes already set unless explicitly overridden.

What did i run?

i wanted to update an existing host, with:

```
ansible localhost -m tribe29.checkmk.host -a 'server_url="http://localhost/" site="cmk" automation_user="automation" automation_secret="12345" host_name="myhost" state="present" attributes="{{attr}}"' -e '{"attr": {"alias": "myalias"}}'
```

expexted:

- only alias attribute would change

what happend:

- ipaddress attribute and folder was reseted to defaults

ogaida:
# Hallo

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

- host attributes will be complete  overwritten on update. 
- folder must given

Issue Number: `#175`

## What is the new behavior?

added:
  - update_attributes handling like described in api
  - remove_attributes handling like described in api 
changed:
  - folder has no default any more, except on create it is "/"
 

## Other information

Ich habe die Args update_attributes und remove_attributes ohne Angabe der Folder getestet. Auch create ohne Folder geht. Habe einfach nach eurer API Definition hier entsprechend ergänzt. Wenn ihr Fragen habt, würde ich mich freuen, wenn wir einfach telefonieren oder per Teams oder was auch immer quatschen. Schöne Grüße, Olli


plugins/modules/host.py
```diff
@@ -308,13 +354,32 @@ def run_module():
...
         if attributes != {} and current_explicit_attributes != attributes:
-            set_host_attributes(module, attributes, base_url, headers)
-            msg_tokens.append("Host attributes changed.")
+            set_host_attributes(module, attributes, base_url, headers, "attributes")
+            msg_tokens.append("Host attributes replaced.")
+
+        if (
+            update_attributes != {}
+            and current_explicit_attributes
+            != current_explicit_attributes.update(update_attributes)
```
godspeed-you, line 368: Are you sure, that this works? Example:
```
>>> this = {'foo': {'bar': 1}}
>>> that = {'foo': {'bar': 1}}
>>> print("similar") if this == that else "not similar"
similar
>>> print("similar") if this == this.update(that) else "not similar"
'not similar'
```

plugins/modules/host.py
```diff
@@ -261,8 +299,10 @@ def run_module():
                 }
             ],
         ),
-        attributes=dict(type="raw", default=[]),
-        folder=dict(type="str", default="/"),
+        attributes=dict(type="raw", default={}),
```
godspeed-you, line 302: Am I correct, that this implies to change a task several times? Example:

1. update tasks by adding update/remove attributes entries
2. run on all hosts
3. update task to by removing update/remove attributes entries
4. don't forget to add the changes to the regular attributes (as they now are only used during creation)

Also, I guess, this adds more complexity and potential bugs (see next comment)

godspeed-you (commented):
I guess, the tests will fail, if the test order would be:

1. Create hosts
2. Create hosts again
3. Activate
4. Change hosts
5. Activate
6. Change hosts again 
7. Activate
8. Delete hosts
9. Activate 

Currently, there is no test for changing a host and no test for idempotence.

ogaida:
I found one first mistake: the samples for using the new args `update_attributes` and `remove_attributes` have the wrong name. It is an update and not a create task. Please change it, when you review the code. thanks!

plugins/modules/host.py
ogaida: So, it does not work, sorry for that. Do you know a solution? I am not a good dev.

plugins/modules/host.py
ogaida: i think it is not such a big problem, because you can use the attributes arg like before. I only extend the possibilties for updates from the api to the module. If i only want to change one attribute of a host, then it is easy to use with the update_attributes arg.

plugins/modules/host.py
godspeed-you: Me neither. ;) 
I'm working on a very similar problem. And I guess, if the problem is solved, we do not need the `update_attributes`/`delete_attributes` anymore. Will let you know if I found a solution.

ogaida: I found the pydantic python package. It does what we need:

```python
>>> from pydantic.utils import deep_update
>>> this = {'foo': {'bar': 1, 'bim': 2}}
>>> that = {'foo': {'bar': 1}}
>>> this == deep_update(this, that)
True
>>> that = {'foo': {'bar': 3}}
>>> this == deep_update(this, that)
False
>>> deep_update(this, that)
{'foo': {'bar': 3, 'bim': 2}}
>>>
```

ogaida:
I have have included the pydantic python module. Now the integration tests fail because of missing this module.

plugins/modules/host.py
ogaida: I switched to pydantic deep_update function. Now we need to extend the integration test environment with the pydantic python module.

robin-tribe29:
Thanks for your contribution @ogaida!

> I have included the pydantic python module. Now the integration tests fail because of missing this module.

The Ansible docs state, that dependencies should be kept at an absolute minimum. I am afraid introducing this Python dependency is uncalled-for. We should probably find a solution with a more basic approach. Maybe @godspeed-you finds a more minimalistic approach as he mentioned, he is facing a similar issue.

ogaida:
@robin-tribe29

Your answer is unacceptable to me.

if there is no other function which will update a python dictionary of dictionaries, you can write your own function. But how to use your own function? This will introduce a new python dependency either.

robin-tribe29:
I am no developer myself - far from it. So I really cannot say, if this is the optimal solution, then we need to do it.
I can merely add my experience, that installing Python dependencies is always added trouble for users and I just want to reduce friction as much as possible.
That being said, let's see, if someone can chime in with a constructive idea on how to tackle this.

plugins/modules/host.py
msekania: @godspeed-you, @ogaida 

> print("similar") if this == this.update(that) else "not similar"

Sorry, but what is the return value of `this.update(that)`?
It doesn't return any value (returns None).

So, first create temporary, then update, then compare

msekania:
> I can merely add my experience, that installing Python dependencies is always added trouble for users and I just want to reduce friction as much as possible.

I absolutely agree!

moreover, `this.update(that)` is a method that does not return anything, in the examples above, however, it is expected that it returns updated dictionary.
It just updates dictionary does not return anything.
So, one should use it appropriately.

plugins/modules/host.py
msekania: @ogaida explanation was about the first example in the stream.

What is the maxim level of the nesting? Do we know that?

As an alternative one can also use JSON serialization (JSON is already included) of the sorted dictionaries, but if some other solution does not exists.

And I will also avoid importing extra libraries if one can find a compact solution with already imported ones.
I'll also try to find out how one can do this.

robin-tribe29: Thanks gentlemen, for the constructive discussion!

lgetwan: What about doing it like this:
```
>>> this = {'foo': {'bar': 1, 'bim': 2}}
>>> that = {'foo': {'bar': 1}}
>>> temp = this.copy()
>>> temp.update(that)
>>> this
{'foo': {'bar': 1, 'bim': 2}}
>>> temp
{'foo': {'bar': 1}}
>>> this == temp
False
```

Is it not "deep" enough?

ogaida:
@robin-tribe29 3 questions:

1. How long do you want to wait for an alternative and constructive idea? 
2. Are you aware of, if someone has a constructive idea which we want to use in more than one module. How can we take care of a DRY implemantion without making a python module out of it? So we end up in the same dilemma.
3. Who makes the decisions here?

plugins/modules/host.py
ogaida: @Igetwan

as you have shown, the value of foo has been replaced totally by `{'bar': 1}`, so the bim value is lost. Compared to pedantic deep_update bim survived. the dict.update function replaces all values on the first level.

ogaida:
today I added error handling for missing modules. I followed the instructions of the ansible developer guide for importing shared code, see https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_best_practices.html#importing-and-using-shared-code .

This guide is also telling us btw:

- “Don’t repeat yourself” is generally a good philosophy
- Use shared code whenever possible - don’t reinvent the wheel.

 see https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_best_practices.html#functions-and-methods .

@all : I would be happy if you take a look at the guide.

✌️

msekania:
@ogaida 

https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/common/dict_transformations.py

will

`dict_merge`
`recursive_diff`

do the job?

ogaida:
@msekania 

Did you read the ansible guide?

It says "don’t reinvent the wheel". pydantic has what we need, it is ready, it is tested and it is used by many other frameworks.

The ansible guide says "use shared code". This code is already used for ansible documentation. 

If everyone develops everything from scratch, even though the modules are already finished, where are we going? The source code would be unnecessarily large, confusing, error-prone and ugly. Do you really want that?

msekania:
@ogaida 

have you read it completely or only that part? ;-)

And have you read my proposal?

https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/common/dict_transformations.py

It is exactly from the ansible core utils, sorry but I am definitely not reinventing wheel here!

ogaida:
@msekania

ok, sorry, I have a look at it. Thank you for the hint!

msekania:
@ogaida,

I have nothing against `pydantic`.

IMHO,  with `pydentic` one creates extra dependency and if one can solve the same problem with other libraries which are already supplied then I do not see the reason for including `pydentic`.
I also do decide nor have an influence on what tribe29 people will decide.

One can make proposals and hope that they will be accepted or fork the repository and do better.

ogaida:
@robin-tribe29 

The tests are looking good so far, is there a scenario you are missing?

we could add tests for update_attribute and remove_attribute and some more for idempotency of these ones. Should I make another pr for that?

robin-tribe29:
> we could add tests for update_attribute and remove_attribute and some more for idempotency of these ones. Should I make another pr for that?

You can add those to this PR, I think that makes the most sense. Thanks!

lgetwan (approved):
Looks good from my point of view.
I like the fact that it's not a breaking change, because the "attributes" parameter still works as before.
Thanks for your work, Oliver! :-)

ogaida:
@lgetwan  @robin-tribe29 

i have added tests in `tests/integration/targets/host/tasks/test.yml` and I made a minor change for update_attributes. Is this to late, or could we reopen the pr?

robin-tribe29:
We are still in the go. The update to the integration tests look fine to me. We will review all integration tests in the future anyway, but for now it looks good.

Once we get the last findings from the tests sorted out, I will merge this. :v:

robin-tribe29:
I already fixed the merge conflict, but something seems to be broken in the module itself.
@ogaida can you have another look?

PR #318, opened 2023-04-19
module:folder, Add update_attributes, remove_attributes and ansible check_mode; Fix #289

### LINKED ISSUE #289: [BUG] module "folder" not idempotent when setting attributes
I'd like to create a folder with some attributes set:
```
    - name: create single folder
      folder:
        site: "{{ site }}"
        path: /compute/testme
        name: testme
        attributes:
          contactgroups:
            groups:
              - "testme-admin"
            recurse_perms: true
        state: present
      delegate_to: localhost
```
On first use, this task succeeds and does exactly what I want. Repeating this task leads to an error
```
fatal: [testps -> localhost]: FAILED! => {"changed": false, "msg": "Error calling API. HTTP code 400. Details: b'{\"title\": \"Bad Request\", \"status\": 400, \"detail\": \"These fields have problems: name\", \"fields\": {\"name\": [\"Unknown field.\"]}}', "}
```
Without the "attributes" part, the task is idempotent and works as expected.

msekania:
Adds support of `update_attributes` and `remove_attributes`.

Adds ansible check_mode functionality

## Pull request type

- [X] Bugfix
- [X] Feature
- [X] Code style update (formatting, renaming)
- [ ] Build related changes

- Do not support `update_attributes` and `remove_attributes` ;
- No ansible check_mode.
- Idempotency #289 

## What is the new behavior?

- Adds `update_attributes` and `remove_attributes` features;
- Adds ansible check_mode.
- Fixes #289 (not finished yet!)

### DIFF
--- a/plugins/modules/folder.py
+++ b/plugins/modules/folder.py
@@ -140,7 +140,6 @@ import json
 # https://docs.ansible.com/ansible/latest/dev_guide/testing/sanity/import.html
 from ansible.module_utils.basic import AnsibleModule, missing_required_lib
 from ansible.module_utils.common.dict_transformations import dict_merge
-from ansible.module_utils.common.dict_transformations import recursive_diff
 from ansible.module_utils.urls import fetch_url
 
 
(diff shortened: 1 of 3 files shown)


msekania:
@robin-checkmk, @lgetwan 

"DONE"!

Currently there is only one solution for the `attributes` idempotency, namely using `update_attributes`.

REST API returns for attributes also default values that users do not set.
From returned values one also cannot identify which values correspond to defaults and which were set actively.

Creating and deleting intermediate dummy folder is not a solution for this issues to solve this issue, IMHO.

Exactly similar issues are also in `module:host`!

plugins/modules/folder.py
```diff
@@ -300,14 +391,20 @@ def run_module():
             )
 
     elif state == "present" and current_state == "absent":
-        create_folder(module, attributes, base_url, headers)
+        if update_attributes != {} and attributes == {}:
+            attributes = update_attributes
+        if not module.params["name"]:
```
msekania: @robin-checkmk 

This was an issue!
My fault!

lgetwan (approved):
I made some tests with different scenarios, and the module behaves as expected. :-)
Doing this, I realized that using "attributes" makes no longer sense if you want to create an idempotent playbook. You can use "update_attributes" when initially creating a folder.
Thanks for solving this idempotency issue!

Issue #314, opened 2023-04-17
[BUG] Autodiscovery of services no longer works from version 2.1.0p21 through to 2.2.0b3


ashleykleynhans:
**Describe the bug**

The `tribe29.checkmk.agent` role, has an option `checkmk_agent_discover` which last worked in version 2.1.0p20, but has not worked in any version since.

The host gets added to the CheckMk server, but the services are not automatically discovered and have to be discovered manually.

I tried this as a work-around but it only works sometimes, and not always, and also its annoying that when it does actually work, CheckMk first alerts that there are undiscovered services on the host before they are actually discovered by the tasks to discover the services and activate the changes.

```yml
  tasks:
    - name: Install CheckMK Agent
      ansible.builtin.include_role:
        name: tribe29.checkmk.agent

    - name: Discover Services
      tribe29.checkmk.discovery:
        server_url: "http://{{ checkmk_agent_server }}/"
        site: "{{ checkmk_agent_site }}"
        automation_user: "{{ checkmk_agent_user }}"
        automation_secret: "{{ checkmk_agent_pass }}"
        host_name: "{{ checkmk_agent_host_name }}"
        state: "fix_all"

    - name: Activate Changes
      tribe29.checkmk.activation:
        server_url: "http://{{ checkmk_agent_server }}/"
        site: "{{ checkmk_agent_site }}"
        automation_user: "{{ checkmk_agent_user }}"
        automation_secret: "{{ checkmk_agent_pass }}"
        force_foreign_changes: 'true'
      run_once: 'true'
```

**Component Name**

tribe29.checkmk.agent role/service discovery

**Ansible Version**

```console
$ ansible --version

ansible [core 2.14.4]
  config file = None
  configured module search path = ['/home/ubuntu/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/ubuntu/.cache/venv/ansible-3.10.5/lib/python3.10/site-packages/ansible
  ansible collection location = /home/ubuntu/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/ubuntu/.cache/venv/ansible-3.10.5/bin/ansible
  python version = 3.10.6 (main, Aug  2 2022, 15:11:03) [GCC 7.5.0] (/home/ubuntu/.cache/venv/ansible-3.10.5/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```

**Checkmk Version**

```console
2.2.0b3
```

[...]

PR #319, opened 2023-04-19
Fix service discovery behavior in the agent role

robin-checkmk:
## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

#314 

## What is the new behavior?

Before this change, it could occur, that no services were added during discovery, because no information was yet available. With the updated task, we ensure, that current data is there. Also, we only add new services, as a removal of services would indicate a subsequent run of the role and the removal would be undesired there.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -3,7 +3,7 @@ name: Bug report
 about: Create a report to help us improve
 title: "[BUG]"
 labels: bug
-assignees: robin-tribe29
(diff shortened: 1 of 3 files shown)


robin-checkmk:
I accidentally put some unrelated changes into this PR. However, they are minor and would be pushed to `devel` anyway.

robin-checkmk:
I found some curiosities, which need to be sorted out first. So for the time being this PR is on hold.

robin-checkmk:
Fix for the molecule failures is in #363 so we will ignore the errors here.

Issue #325, opened 2023-04-25
[BUG] Activate changes handler fails with self signed certificate


NicolasVongeheur:
**Describe the bug**
When using a self signed cert, the agents handler "Activate Changes" fails with
`FAILED! => {"changed": false, "msg": "-1 - Error calling API"}`

**Component Name**
`agent`

**Collection Version**
v0.20.0

`validate_certs: "{{ checkmk_agent_server_validate_certs }}"` should be added to that task

PR #326, opened 2023-04-25
[BUG] Activate changes handler fails with self signed certificate

NicolasVongeheur:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #325

## What is the new behavior?
Activate changes works even with self signed certs

### DIFF
--- a/roles/agent/handlers/main.yml
+++ b/roles/agent/handlers/main.yml
@@ -9,5 +9,5 @@
     force_foreign_changes: "{{ checkmk_agent_force_foreign_changes }}"
     validate_certs: "{{ checkmk_agent_server_validate_certs }}"
   delegate_to: "{{ checkmk_agent_delegate_api_calls }}"
-  run_once: true # noqa run-once[task]
+  run_once: true  # noqa run-once[task]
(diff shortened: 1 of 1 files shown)


roles/agent/handlers/main.yml
```diff
@@ -7,6 +7,7 @@
     automation_user: "{{ checkmk_agent_user }}"
     automation_secret: "{{ checkmk_agent_auth }}"
     force_foreign_changes: "{{ checkmk_agent_force_foreign_changes }}"
+    validate_certs: "{{ checkmk_agent_server_validate_certs }}"
   delegate_to: "{{ checkmk_agent_delegate_api_calls }}"
-  run_once: true  # noqa run-once[task]
+  run_once: true # noqa run-once[task]
```
robin-checkmk, line 12: The two blanks are on purpose here. Please fix that.

robin-checkmk (changes requested):
Good catch, thanks!

Just fix the nitpick, and we are good to merge.

robin-checkmk (approved):

PR #344, opened 2023-05-24
Migrated discovery and bakery module and added Details of HTTP Response to msg

Max-checkmk:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Bakery and Discovery are working but don't rely on new API.
With the new API there is not detailed output of the HTTP Response anymore.

## What is the new behavior?
- Bakery and Discovery Module are now migrated and are working with the new CheckMKAPI
- If there is (as an example) a HTTP 400 Response the body of the Response will be added to the msg.
This is the same behavior as with classic plugins.

## Other information
Output of the HTTP Response can be improved. Looks like before at the moment.

### DIFF
--- a/plugins/modules/bakery.py
+++ b/plugins/modules/bakery.py
@@ -87,8 +87,61 @@ message:
     sample: 'Done.'
 """
 
+import time
+
 from ansible.module_utils.basic import AnsibleModule
-from ansible.module_utils.urls import fetch_url
+from ansible_collections.tribe29.checkmk.plugins.module_utils.api import CheckmkAPI
+from ansible_collections.tribe29.checkmk.plugins.module_utils.utils import (
+    result_as_dict,
+)
+
+HTTP_CODES = {
+    # http_code: (changed, failed, "Message")
+    200: (True, False, "The operation was done successfully."),
+    204: (
+        True,
+        False,
+        "No Content: Operation done successfully. No further output.",
+    ),
+    400: (False, True, "Bad Request: Parameter or validation failure."),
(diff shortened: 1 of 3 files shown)


lgetwan:
Hi Max,
is this PR still under construction? If yes, then please mark it as draft.
Once, it's finished, I will review it.
Thanks! :-)

Max-checkmk:
>

robin-checkmk:
@lgetwan pending your code review I am happy with this.

plugins/modules/bakery.py
```diff
@@ -87,8 +87,61 @@
...
+            }
+        else:
+            data = ""
+
+        if self.params.get("state", "") == "baked":
+            action = "bake"
+        if self.params.get("state", "") == "signed":
+            action = "sign"
+        if self.params.get("state", "") == "baked_signed":
+            action = "bake_and_sign"
```
lgetwan, line 137: Either you believe that `state` can never be empty or you have to define an `else` case. ;-)

lgetwan (commented):
I really like the way the CheckmkAPI class simpifies the modules' code. :-D
I only have a small comment on your code.

plugins/modules/bakery.py
Max-checkmk: You're right. Would it be also possible to set `state` as required? After rethinking it, this should be anyway the case...

lgetwan: I agree!

lgetwan (approved):
Approved.

PR #347, opened 2023-06-01
Fixed integration tests and added 2.2 compatibility

Max-checkmk:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

- Integration-Tests just seem to run against 2.1 and 2.2, but all 3 sites are created as 2.0
- contact_group, host_group and service_group modules don't work with 2.2
- Integration test for the downtimes-module fails against 2.1 and 2.2 as services (cpu load, memory, ...) aren't found

## What is the new behavior?

- The prep.yml is looped now completely so every site gets created with the installed version from the task before.
- All _group modules have been fixed to work with 2.2
- Changed service description in the downtimes test to existing ones.

## Other information

Fix for _group-modules is not very pretty, but modules should be migrated to new API with version-check anyway.


tests/integration/targets/rule/tasks/prep.yml
```diff
@@ -22,11 +22,4 @@
...
   register: site_status
   changed_when: site_status.rc == "0"
-  loop: "{{ checkmk_versions }}"
-
-- name: "Gather Date and Time Facts on localhost."
-  ansible.builtin.setup:
-    gather_subset:
-      - date_time
-  delegate_to: localhost
-  run_once: true  # noqa run-once[task]
```
robin-checkmk, line 32: This cannot be removed. It is necessary for the test to work properly.

robin-checkmk (changes requested):

plugins/modules/contact_group.py
```diff
@@ -184,13 +184,24 @@ def get_current_contact_groups(module, base_url, headers):
...
+                    for el in tmp
+                ]
+            else:  # 2.0.0 and 2.1.0
+                current_groups = [
+                    {
+                        "name": el.get("href").rsplit("/", 1)[-1],
+                        "title": el.get("title", el.get("name")),
+                    }
+                    for el in tmp
+                ]
```
lgetwan, line 204: It would be nicer if you could use the getversion() method that you have introduced in #344

lgetwan (commented):
Thanks for fixing this!
I'm not sure if you can solve the comment I added in the code without depending on #344, too much.

plugins/modules/contact_group.py
Max-checkmk: Yes, but i would prefer to do this when migrating the whole plugin to the new API.

lgetwan: Sounds reasonable.

lgetwan (approved):
Approved.

Issue #367, opened 2023-06-15
[BUG] Discovery times out after upgrading from 2.2.0p1 to 2.2.0p3


ashleykleynhans:
**Describe the bug**

I upgraded from 2.1.0.p1 to 2.1.0p3 and am now getting time out errors with Discovery.

```
fatal: [172.31.138.62 -> localhost]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {
            "automation_secret": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
            "automation_user": "automation",
            "host_name": "eventsdashboard-172-31-138-62",
            "server_url": "http://cmk-prod.snappic.xyz:80/",
            "site": "prod",
            "state": "fix_all",
            "validate_certs": true
        }
    },
    "msg": "-1 - Error calling API Details: Connection failure: timed out"
}
```

I also tried to upgrade tribe29.checkmk to checkmk.general 2.1.0 but still having the same issue.

**Component Name**

checkmk.general.discovery

**Ansible Version**

```console
$ ansible --version
ansible [core 2.14.4]
  config file = /home/ubuntu/ansible/ansible.cfg
  configured module search path = ['/home/ubuntu/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/ubuntu/.cache/venv/ansible-3.10.5/lib/python3.10/site-packages/ansible
  ansible collection location = /home/ubuntu/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/ubuntu/.cache/venv/ansible-3.10.5/bin/ansible
  python version = 3.10.6 (main, Aug  2 2022, 15:11:03) [GCC 7.5.0] (/home/ubuntu/.cache/venv/ansible-3.10.5/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```

**Checkmk Version and Edition**

```console
e.g. 2.1.0p3 (CRE)
```

**Collection Version**

```console
$ ansible-galaxy collection list
Collection        Version
----------------- -------
ansible.posix     1.4.0
checkmk.general   2.1.0
community.general 7.0.1
```

**To Reproduce**

1. Set appropriate variables for the playbook in the example below.
2. Run the Ansible playbook below.

**Expected behavior**

Services should be discovered and no API timeouts.

**Actual behavior**

I get an API timeout error:
```
fatal: [172.31.138.62 -> localhost]: FAILED! => {
[...]

PR #369, opened 2023-06-15
Feature bulk_discovery module

msekania:
Introduces the **bulk_discovery** module

## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Introduces the **bulk_discovery** module that allows to (re)discover services on the given list of hosts (`host_names`) in parallel manner.

## What is the new behavior?

- bulk discovery on given hosts
- waits until the procedure is finished successfully or failed 

## Other information

```yaml
- name: "Add newly discovered services, update labels and remove vanished services on host; 3 at once"
  checkmk.general.bulk_discovery:
    server_url: "http://localhost/"
    site: "my_site"
    automation_user: "automation"
    automation_secret: "$SECRET"
    host_names: "[my_host_0, my_host_1, my_host_2, my_host_3, my_host_4, my_host_5]"
    state: "fix_all"
    bulk_size: 3
```
```

### CODE REVIEW

**Comment by robin-checkmk** [comment]
Hi @msekania and thanks for this contribution!
We briefly discussed internally, and we think we should implement bulk discovery into the existing `discovery` module, rather than building an additional module. This would be analogous to e.g., the `host_group` module. Do you have any objection to that?

We can also support you, if you want, but we do not want to overstep.

P.S.: While you are at it, your findings in #366 are accurate, and you could fix them yourself, when you are touching the `discovery` module anyway.

File: plugins/modules/bulk_discovery.py [hunk 1]
```diff
@@ -0,0 +1,219 @@
...
+#             endpoint=(
+#                 "/objects/host/"
+#                 + self.params.get("host_name")
+#                 + "/actions/discover_services/invoke"
+#             ),
+#             data=data,
+#             method="POST",
+#         )
+
+
```
lgetwan, line 152: You can delete this, because for the bulk_discovery, we don't have an old/deprecated endpoint, as we have for the normal discovery.

plugins/modules/bulk_discovery.py
```diff
@@ -0,0 +1,219 @@
...
+        do_full_scan=dict(type="bool", default=True),
+        bulk_size=dict(type="int", default=1),
+        ignore_errors=dict(type="bool", default=True),
+    )
+    module = AnsibleModule(argument_spec=module_args, supports_check_mode=False)
+
+    bulk_discovery = BulkDiscoveryAPI(module)
+    checkmkversion = bulk_discovery.getversion()
+    # if checkmkversion[0] == "2" and checkmkversion[1] == "0":
+    #     bulk_discovery = oldDiscoveryAPI(module)
```
lgetwan, line 193: You should use the getversion() call and exit with something like "unsupported in CMK 2.0.0", if you find a 2.0.0 version. The "bulk_discovery" endpoint doesn't exist there, yet.

tests/integration/targets/bulk_discovery/tasks/test.yml
```diff
@@ -0,0 +1,125 @@
...
+    folder: "{{ item.folder }}"
+    attributes:
+      site: "{{ outer_item.site }}"
+      ipaddress: 127.0.0.1
+    state: "present"
+  delegate_to: localhost
+  run_once: true  # noqa run-once[task]
+  loop: "{{ checkmk_hosts }}"
+
+- name: "{{ outer_item.version }} - Update host labels."
```
lgetwan, line 18: All bulk_discovery tests will fail when tested against CMK 2.0.0. You should add this as "it's ok, we expect that" into the tests.

lgetwan (commented):
Thanks for adding this module!
I added some comments regarding the handling of CMK 2.0.0, where the bulk_discovery endpoint didn't exist.

lgetwan:
Looks like multiple things were happening in parallel, this morning. ;-)
To be honest, I also like the idea of integrating the "bulk" variant into the existing discovery module, with something like:

- having a parameter "bulk [yes|no]"
- using a single host or a list, based on that parameter.

Max-checkmk:
> Looks like multiple things were happening in parallel, this morning. ;-) To be honest, I also like the idea of integrating the "bulk" variant into the existing discovery module, with something like:
> 
>     * having a parameter "bulk [yes|no]"
> 
>     * using a single host or a list, based on that parameter.

I've discussed this already with Robin and we think that it would be enough to have 2 options like host_name and host_names which are marked as mutually_exclusive. Host_group module is a good example for this.
But maybe a switch for bulk yes/no is easier to understand...

I wanted to start with this today, but @msekania was faster. ;)

plugins/modules/bulk_discovery.py
msekania: fine, I'll do that

msekania:
> I've discussed this already with Robin and we think that it would be enough to have 2 options like host_name and host_names which are marked as mutually_exclusive. Host_group module is a good example for this. But maybe a switch for bulk yes/no is easier to understand...
> 
> I wanted to start with this today, but @msekania was faster. ;)

@Max-checkmk @robin-checkmk 

Sorry, I was waiting for this feature for quite a while now ;-).

Of course one can combine discovery and bulk_discovery. I also rewrote it in a way to closely match `bulk_discovery`.
I will carry out the corresponding changes.

There are couple of issues though:
1. names `host_name` and `host_names` just single 's' decides whether **str** or **list of str** should be a parameter and **str** is also a sequence, hence can be interpreted as list and consequently errors running the module.
2.  property `state` in case of `bulk_discovery` to my knowledge has less choices, only 5 as compared to `discovery` with 6

msekania:
@robin-checkmk 

- what's about `hosts` instead of `host_names`, not the best alternative but IMHO still better than `host_names`,

- and `mode:` with either `single` or `bulk` value

robin-checkmk:
I like both approaches. :+1: 
The first resonates with the `host_group` module and the second one seems intuitive enough.

Max-checkmk:
I also like the idea of "hosts" and "host_name". With this we could drop the extra mode-option.
Reason for the missing choice in state is, that you're looking at 2.0 and 2.1.
"tabula_rasa" is new in 2.2.
A little confusing, but... refresh in 2.0/2.1 is tabula_rasa and refresh in 2.2 is now rescan.

I also thought of another check that if users try to use tabula_rasa against 2.0/2.1 we create a custom error.

msekania:
> I also thought of another check that if users try to use tabula_rasa against 2.0/2.1 we create a custom error.

I have included it

plugins/modules/discovery.py
```diff
@@ -178,24 +273,92 @@ def run_module():
...
+                time.sleep(3)
+    elif len(module.params.get("groups", [])) > 0:
+        if module.params.get("state") == "tabula_rasa":
+            result = RESULT(
+                # http_code=http_code,
+                msg="State 'tabula_rasa' is not supported by Bulk Discovery mode",
+                failed=True,
+                changed=False,
+            )
+            module.fail_json(**result_as_dict(result))
```
Max-checkmk, line 333: Sorry for the confusion. But "tabula_rasa" is also supported for the bulk discovery endpoint. But like with the single service discovery just with 2.2 and upwards.

So if the version is 2.0 or 2.1, doesn't matter if bulk or single the warning should be displayed.

plugins/modules/discovery.py
```diff
@@ -178,24 +273,92 @@ def run_module():
...
+
+        result = discovery.post()
+
+        # If the API returns 302, check the service completion endpoint
+        # until the discovery has completed successfully.
+        if result.http_code == "302":
+            servicecompletion = ServiceCompletionAPI(module)
+            while True:
+                result = servicecompletion.get()
+                if result.http_code == "302":
```
Max-checkmk, line 321: I'm not sure about this.

If we get the redirect 302 from the service discovery endpoint we should ask the Service Completion Endpoint until it's response is 204.
If the response form SC is 302 it says "service discovery is still running".

So i would suggest to revert this to `while result.http_code == "302"`

Max-checkmk (changes requested):

plugins/modules/discovery.py
msekania: > If the response form SC is 302 it says "service discovery is still running".
> 
> So i would suggest to revert this to `while result.http_code == "302"`

It should state `if result.http_code != "302"`

there are two different objects that return this code, that's why I prefer only to create the  `ServiceCompletionAPI` if `discovery` returned `302` and not always

plugins/modules/discovery.py
msekania: to my knowledge `tabula_rasa` is there for single discovery for 2.1 also

plugins/modules/discovery.py
Max-checkmk: Yeah. That's ok for me. Thought of the `while result.http_code == "302"` instead of `while True`.
But i don't know what's the best from a developer sight.

msekania: > Yeah. That's ok for me. Thought of the `while result.http_code == "302"` instead of `while True`

how do you want to check then without `result = servicecompletion.get()` which is in the body of while loop,

one requires do -- until loop which python does not have and alternative is while True: with break condition

Max-checkmk: Thought of having `result = servicecompletion.get()` still in the loop. Then the loop should end if the result is something different then 302. Right?

But you're new solution with single and bulk in one statement is great. Let's keep it that like this anyway!

msekania: > Then the loop should end if the result is something different then 302. Right?

Yes, correct!
I have made changes!

plugins/modules/discovery.py
Max-checkmk: At least not in my running 2.1.0p28 ;)

msekania:
@lgetwan @Max-checkmk 

in test file tests/integration/targets/discovery/tasks/test.yml, lines 30-53, the second task is already contained in the first one. Could it be that the original intent was different?

plugins/modules/discovery.py
msekania: 302 not "302"

msekania:
Done!

BUG that treated http_code as a str # ate a lots of time. https://github.com/Checkmk/ansible-collection-checkmk.general/issues/371

unfortunately, I have also blindly trusted this part of the accepted code.

one needs a bit better test, because existing one was also passing by returning '302' in Ansible output

IMHO:

    module_utils needs a better documentation;
    and it should get also more flexible in order to cover the rest of the REST API command, otherwise it's not a help.

Max-checkmk:
> Done!
> 
> BUG that treated http_code as a str # ate a lots of time. #371
> 
> unfortunately, I have also blindly trusted this part of the accepted code.
> 
> one needs a bit better test, because existing one was also passing by returning '302' in Ansible output
> 
> IMHO:
> 
>     * module_utils needs a better documentation;
> 
>     * and it should get also more flexible in order to cover the rest of the REST API command, otherwise it's not a help.

Thanks again Michael for discovering this!

module_utils has just started, so sure there is a lot room for improvement.
What REST API commands do you miss the most at the moment?

msekania:
@Max-checkmk,

I do not like that all fields of RESULT are mandatory.
May be one needs some other modification of it, which is also suited for other, not REST API, return messages.

Another point would be that Ansible messages will include more information that comes from REST API, at least one has to have this option.
I understand that none of the plugings do that now, but on the other hand `module_utils` does not give this possibility either.
One mainly tries to cover what is already there in the modules instead of making decisions how the standardized return messages should or might look like.

From my point of view, I would have rather written the module(s) which covers REST API in a way how it is also documented in the REST API (mainly keeping grouping).
One might also build this stuff on the current version.

.github/workflows/ans-int-test-bulk_discovery.yaml
Max-checkmk, line 1: As the bulk_discovery is now integrated into the discovery module this file can be removed

plugins/modules/discovery.py
```diff
@@ -24,17 +24,36 @@
...
+        description: The option whether to ignore errors in single check plugins. (Bulk mode only).
+        type: bool
+        default: True
 
 author:
     - Robin Gierse (@robin-checkmk)
+    - Michael Sekania (@msekania)
 """
 
 EXAMPLES = r"""
```
Max-checkmk, line 59: We should add an example for the bulk_discovery here.

tests/integration/targets/discovery/tasks/test.yml
```diff
@@ -123,3 +123,187 @@
...
+        server_url: "{{ server_url }}"
+        site: "{{ outer_item.site }}"
+        automation_user: "{{ automation_user }}"
+        automation_secret: "{{ automation_secret }}"
+        sites:
+          - "{{ outer_item.site }}"
+      delegate_to: localhost
+      run_once: true  # noqa run-once[task]
+
+- name: Chech bulk discovery mode for Check MK > 2.0
```
Max-checkmk, line 255: Chech -> Check

tests/integration/targets/discovery/tasks/test.yml
```diff
@@ -123,3 +123,187 @@
...
+  when: "'2.0.0' in outer_item.version"
+  block:
```
Max-checkmk, line 257: Why not doing the bulk tests right after the single ones without the separate versions block and using "when" like before if something just works for a specific version?

tests/integration/targets/discovery/tasks/test.yml
```diff
@@ -123,3 +123,187 @@
       - "{{ outer_item.site }}"
   delegate_to: localhost
   run_once: true  # noqa run-once[task]
+
+- name: Chech bulk discovery mode for Check MK > 2.0
+  when: "not '2.0.0' in outer_item.version"
+  block:
```
Max-checkmk, line 129: Why not doing the bulk tests right after the single ones without the separate version blocks and using "when" like before if something just works for a specific version?

tests/integration/targets/discovery/tasks/test.yml
```diff
@@ -123,3 +123,187 @@
...
+        site: "{{ outer_item.site }}"
+        automation_user: "{{ automation_user }}"
+        automation_secret: "{{ automation_secret }}"
+        hosts: "{{ checkmk_host_names }}"
+        state: "refresh"
+        bulk_size: 5
+      delegate_to: localhost
+      run_once: true  # noqa run-once[task]
+
+    - name: "{{ outer_item.version }} - Tabula Rasa. (Only 2.2). (Bulk)."
```
Max-checkmk, line 170: Name is "Tabula Rasa" but in the task it uses "refresh",

tests/integration/targets/discovery/tasks/test.yml
```diff
@@ -123,3 +123,187 @@
       - "{{ outer_item.site }}"
   delegate_to: localhost
   run_once: true  # noqa run-once[task]
+
+- name: Chech bulk discovery mode for Check MK > 2.0
```
Max-checkmk, line 127: Chech -> Check
Check MK -> Checkmk

tests/integration/targets/discovery/tasks/test.yml
```diff
@@ -123,3 +123,187 @@
...
+        folder: "{{ item.folder }}"
+        attributes:
+          site: "{{ outer_item.site }}"
+          ipaddress: 127.0.0.1
+        state: "present"
+      delegate_to: localhost
+      run_once: true  # noqa run-once[task]
+      loop: "{{ checkmk_hosts }}"
+
+    - name: "{{ outer_item.version }} - Update host labels. (Bulk)."
```
Max-checkmk, line 274: name says "Update host labels", but in the task there is tabula_rasa

plugins/modules/discovery.py
msekania: There are already there. 3rd and 4th

I have added (Bulk) now in the task name

tests/integration/targets/discovery/tasks/test.yml
msekania: fixed

tests/integration/targets/discovery/tasks/test.yml
msekania: One test what is in task and not in plain text in Name.
So in this case this test is already done with different task name though. Functionality is tested one does not need it twice IMHO.
Not my decision.

I'll leave it as it is.

.github/workflows/ans-int-test-bulk_discovery.yaml
msekania: done

robin-checkmk:
Does anyone have an idea, why some tasks in the integration tests fail with a timeout?
Might be related to #367.

msekania:
@robin-checkmk

and it's mostly random which combination of checkmk and python.

I was thinking of increasing timeout from 10 to 15 in `module_utils/api.py` for `_fetch`, just to test.

I'll do it right now

msekania:
@robin-checkmk ,

15 sec timeout if `_fetch' routine solves the problem which I think is partly the problem of test environment!

msekania:
@robin-checkmk,

by the way, two bots **Run QA** and **Sanity (Ⓐstable-2.14) ** are finished, but here it shows that they are still running.

robin-checkmk:
@msekania 

Adding that timeout does not feel great, but I am open to the workaround as long as it is good enough.

The actions sorted themselves out, apparently. Sometimes they just get stuck.

I took the liberty to change the error message for state `tabula_rasa` with Checkmk 2.0.0 and 2.1.0 slight. Also, I reworked the integration tests, to be a little more organized and maybe even efficient. Please do take a look and let me know what you think. I hope both changes are alright with you.

msekania:
@robin-checkmk 

I indeed like that use `block` sections!

What I do not like, however, are the task names, I would rather use the same name in tasks as in `state` and in brackets what that means.

for example
```yaml
    - name: "{{ outer_item.version }} - Rescan (Tabula Rasa in 2.1 and before) services. (Bulk)."
      discovery:
        site: "{{ outer_item.site }}"
        hosts: "{{ checkmk_host_names }}"
        state: "refresh"
        bulk_size: 5
      delegate_to: localhost
```

`state: "refresh"`

as a task name then,
```yaml
    - name: "{{ outer_item.version }} - Refresh (Rescan, Tabula Rasa in 2.1 and before) services. (Bulk)."
```

It's just a suggestion,  i do not have a strong opinion about it.

msekania:
Hi @robin-checkmk 

> Adding that timeout does not feel great, but I am open to the workaround as long as it is good enough.
> 
> The actions sorted themselves out, apparently. Sometimes they just get stuck.

I have a better solution, I'll commit it here and will not make a separate pull request

```python
        # retry if timed out
        num_of_retries = 2
        for i in range(num_of_retries):
            response, info = fetch_url(
                module=self.module,
                url="%s/%s" % (self.url, endpoint),
                data=self.module.jsonify(data),
                headers=self.headers,
                method=method,
                use_proxy=None,
                timeout=10,
            )

            http_code = info["status"]

            if http_code != -1:
```

In essence retry (currently I set it to twice, one can also increase number) if it's timed out.

I reduces `timeout` to previous 10, one can make it also smaller.

Max-checkmk (approved):
Looks good!

tests/integration/targets/discovery/tasks/test.yml
```diff
@@ -15,184 +15,229 @@
...
+        server_url: "{{ server_url }}"
+        site: "{{ outer_item.site }}"
+        automation_user: "{{ automation_user }}"
+        automation_secret: "{{ automation_secret }}"
+        hosts: "{{ checkmk_host_names }}"
+        state: "tabula_rasa"
+        bulk_size: 5
+      delegate_to: localhost
+      ignore_errors: true  # noqa ignore-errors
+      when: "'2.1.0' in outer_item.version"
```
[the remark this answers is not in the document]
robin-checkmk: If I am not mistaken, this is the difference between the Checkmk versions, where in 2.1.0 "tabula_rasa" is actually "refresh" and in 2.2.0 it is "tabula_rasa". I will double-check and update this.

robin-checkmk:
> What I do not like, however, are the task names, I would rather use the same name in tasks as in `state` and in brackets what that means.

I - hopefully - fixed that to your liking.
I also reorganized and cleaned up some more - in my perception duplicate - tasks.
It should be alright now. I would merge this PR then, if you still see problems in the tests, we can address them afterward.

PR #384, opened 2023-07-11
Solve timeout problems in discovery module

msekania:
Might solve timeout problems in `discovery` module.
 
In `discovery` module Check if discovery process is already running and wait until its finished before starting the new one.

In `module_utils/api.py`, retry `fetch_url` if time outs for 3 times, each time doubling the time out time.

## Pull request type

- [X] Bugfix
- [X] Feature
- [ ] Build related changes

Issue Number: `e.g. #1` 

## What is the new behavior?

- In `discovery` module Check if discovery process is already running and wait until its finished before starting the new one.
- In `module_utils/api.py`, retry `fetch_url` if time outs for 3 times, each time doubling the time out time.


plugins/modules/discovery.py
```diff
@@ -328,12 +328,28 @@ def run_module():
...
+
+        if single_mode:
+            if result.http_code != 302:
+                break
+        else:
+            if not (json.loads(result.content).get("extensions").get("active")):
+                break
+
+            time.sleep(3)
+
```
Max-checkmk, line 346: Is this block here just by accident? Exists already beginning with line 333 (347).

Max-checkmk (changes requested):

plugins/modules/discovery.py
msekania: No, it explicitly checks if discovery started by some other host is already running.

For the case when task runs on several host . I just got an idea from the report that setting `throttle: 1` helps to avoid timeout issues.

Max-checkmk: Ah, now i understood. Good idea, yes!
What do you think of packing this block into a def function as it is totally identical?

What was you idea about throttling? As a hint for users in their playbooks?

msekania: In comment
https://github.com/Checkmk/ansible-collection-checkmk.general/issues/367#issuecomment-1628532206
it was reported that from one of the users setting `throttling: 1` eliminated the timeout issues,
therefore I thought that it could be a race condition problem with already started discovery processes. I'm not 100% sure though.

msekania: > What do you think of packing this block into a def function as it is totally identical?

Yes, and I have taken care of it.

Max-checkmk (approved):
Looks good from my side! ;)

PR #403, opened 2023-08-07
Allow discovery throttling

guenhter:
When a lot of  agents are connected and the ansible agent role runs, then the two discovery tasks are overwhelming the check mk server (it it doesn't have the luxury to have 16 cores and 32 GB RAM :) ). This PR makes it possible to optionally define a throttling for the discovery tasks. By default it is not throttled at all.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes


guenhter:

roles/agent/README.md
```diff
@@ -91,6 +91,11 @@ The folder into which the automatically created host will be places.
 
 Automatically discover services on the host where the agent was installed.
 
+    checkmk_agent_discover_throttle_value: 0
+
+Limits the parallelism of the discovery tasks so that the server doesn't get overwhelmed by 
+many parallel requests. If the value is '0', throttling is deactivated.
```
robin-checkmk, line 97: This might need a little more explanation for the average user.
IIRC using `throttle`means that only `X` parallel jobs run, right?
And that would be exactly what I am missing here.

playbooks/usecases/remote-registration.yml
```diff
@@ -36,6 +36,7 @@
     checkmk_agent_add_host: 'true'
     # These are some generic agent options you might want to configure.
     checkmk_agent_discover: 'true'
+    checkmk_agent_discover_throttle_value: 0
```
robin-checkmk, line 39: In conjunction with my remark in the README, maybe you want to go for a different variable name?
Something like `checkmk_agent_discover_parallel`? I am not entirely happy with that either, but maybe we can figure something out together.

robin-checkmk (changes requested):
Thanks for your contribution!
For your awareness, there is a major rework of the agent role in #156 which we will release first with the next release. But I am positive, that we can then rebase your PR on that and release if one release later.

roles/agent/README.md
guenhter: Good point. I've changed the description. Maybe you can have another look. I'm happy to adapt it again.

playbooks/usecases/remote-registration.yml
guenhter: Done. I named it `checkmk_agent_discover_max_parallell_tasks`. Do you think this is understandable?

robin-checkmk: I think `parallel` has only one `l` at the end, but apart from that it feels well understandable. The variable name gets quite long this way, but I'd rather have long variable names, than obscure ones. Thanks!

roles/agent/README.md
robin-checkmk: Looks good to me, thanks!

playbooks/usecases/remote-registration.yml
guenhter: :) I've changed that typo in the code already.

robin-checkmk: Just saw, that the typo is only in this comment, but the code looks good. :+1:

guenhter:
I'm not certain if my changes made the tests failing...

robin-checkmk:
> I'm not certain if my changes made the tests failing...

They do not. The tests are currently flaky. You could actually try to enable throttling here, to see if that helps at least some of them. But I will take of it eventually, so for the time being, your changes are fine.

guenhter:
I've set the thorttling to 2 for the mulecule tests. Let's see...

robin-checkmk:
Okay, the one left failure is arbitrary, so as I already said, I would consider your contribution validated. :tada: 

We will most likely merge this for the next release, which would be `2.6.0`. Stay tuned and thanks again!

Issue #389, opened 2023-07-28
[BUG] folder module: update endpoint now requires which update option you want


lgetwan:
**Describe the bug**

After https://checkmk.com/werk/15888, it's no longer possible to provide more than one of the parameters _attributes_, _remove_attributes_ and _update_attributes_
The ansible module still alows this, though.

**Component Name**
folder

**Ansible Version**
any

**Checkmk Version and Edition**
any

**Collection Version**
2.3.0

**Current behavior**
The folder module allows the user tries to provide more than one of these parameters and then runs into an ugly error message

**Expected behavior**
The folder module should write a friendly error message if the user tries to provide more than one of these parameters.
The documentation should also tell to use one of them, only.

lgetwan:
@msekania 
That sounds great! There's no refactoring for the folder module in our pipeline, you can go for it.
Our developers are aware that the host_config endpoint has a similar issue, but they don't have it in their backlog, yet.

PR #404, opened 2023-08-07
Fix module folder for future checkmk releases

msekania:
Fixes issue #389 

## Pull request type

- [X] Bugfix
- [X] Feature
- [ ] Build related changes

Currently all three options *attributes*, *remove_attributes*, and *update_attributes* can be used simultaneously, which is no longer supported by API for Check MK >= v2.2.0p7

Issue Number: #389 

## What is the new behavior?

- issue a warning message if more then one option from *attributes*, *remove_attributes*, and *update_attributes* is used simultaneously for Check MK  < v2.2.0p7 
- fail with error message if more then one option from *attributes*, *remove_attributes*, and *update_attributes* is used simultaneously for Check MK  >= v2.2.0p7


robin-checkmk:
@msekania thanks for tackling this!
Is your branch up-to-date with our `devel`? Looks like there are some unrelated changes in here.

msekania:
> Is your branch up-to-date with our `devel`? Looks like there are some unrelated changes in here.

apparently not, changes should be only in folder.py

robin-checkmk:
I'll take a look at the failing tests and the weirdness with the additional changes.

msekania:
@robin-checkmk 

thanks!

robin-checkmk:
I understand the issue with the integration tests now but the solution is not trivial.
I will update this comment, once I have news.

robin-checkmk:
@msekania I implemented a hotfix for the integration tests, but the actual fix is already in `devel`.
So the current test findings should be valid, and if you pull the `devel` branch into yours, the integration tests should be fixed properly.

plugins/modules/folder.py
```diff
@@ -333,6 +361,62 @@ def run_module():
...
+            )
+            or (
+                int(checkmkversion[0]) == 2
+                and int(checkmkversion[1]) == 2
+                and int(patchlevel[0]) == 0
+                and patchtype == "p"
+                and int(patchlevel[1]) >= 7
+            )
+        ):
+            version_ge_220p7 = True
```
lgetwan, line 408: Can you put this into a separate function to keep the run_module() function a bit shorter?
BTW: This algorithm will not work for daily build versions, but I won't consider this a show stopper. ;-)

lgetwan (changes requested):
Looks fine! Thanks for your contribution.
I just added one comment that you should solve before we merge this.

msekania:
@robin-checkmk,

I pulled devel branch, and merged it in the current, which required rebase.

robin-checkmk:
@msekania I am uncertain what exactly happened, but I think your idea of creating a new and clean PR ist the best approach. Go ahead and create it, I will merge tomorrow then and release.
Thanks!

PR #407, opened 2023-08-07
Fix integration tests for enterprise editions

thorian93:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [x] Build related changes

When users outside the Checkmk organization create a pull request, the integrations against enterprise editions fail.
The reason is missing access to a GitHub secret, that contains download credentials for the enterprise edition.

## What is the new behavior?

When the download credentials are not available, the integration tests will skip the relevant tests.

## Other information

An alternative approach might be to use CCE instead of CEE, as that edition needs no credentials to download. But it would not be the cleanest approach, as there are technical differences.

### DIFF
--- a/plugins/modules/activation.py
+++ b/plugins/modules/activation.py
@@ -1,6 +1,8 @@
 #!/usr/bin/python
 # -*- encoding: utf-8; py-indent-offset: 4 -*-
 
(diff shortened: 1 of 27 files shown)


thorian93:

robin-checkmk (approved):

PR #415, opened 2023-08-17
Add a Lookup Plugin to get the Checkmk Version of a Server

lgetwan:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

There are no lookup plugins at all

## What is the new behavior?

- With the lookup plugin, the user can query the server's version.

## Other information

- The full demo playbook now also uses this lookup

### DIFF
--- a/playbooks/demo/lookup.yml
+++ b/playbooks/demo/lookup.yml
@@ -1,0 +1,14 @@
+---
+- name: "Showcase Lookup Plugins."
+  hosts: test
+  strategy: linear
+  gather_facts: false
+  vars_files:
+    - ../vars/auth.yml      # This vars file provides details about your site
+  tasks:
+
+    - name: "Get CMK version."
+      debug:
+        msg: "Version is {{ lookup('checkmk.general.version', server_url+'/'+site, validate_certs=False, automation_user=automation_user, automation_secret=automation_secret)}}" 
+      delegate_to: localhost
+      run_once: 'true'
--- a/playbooks/demo/lookup.yml
+++ b/playbooks/demo/lookup.yml
@@ -7,8 +7,16 @@
     - ../vars/auth.yml      # This vars file provides details about your site
   tasks:
 
-    - name: "Get CMK version."
+    - name: "Get Checkmk version."
       debug:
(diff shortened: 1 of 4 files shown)


playbooks/demo/lookup.yml
```diff
@@ -0,0 +1,14 @@
...
+  hosts: test
+  strategy: linear
+  gather_facts: false
+  vars_files:
+    - ../vars/auth.yml      # This vars file provides details about your site
+  tasks:
+
+    - name: "Get CMK version."
+      debug:
+        msg: "Version is {{ lookup('checkmk.general.version', server_url+'/'+site, validate_certs=False, automation_user=automation_user, automation_secret=automation_secret)}}"
```
robin-checkmk, line 12: Jinja2 spacing should be improved, but then we hit the line length. I can take care of it later, have to look up the syntax again.

robin-checkmk (commented):
I cannot get the demo playbook to work, Ansible complains about not finding the lookup module. But that might be a local issue on my machine.

plugins/lookup/version.py
```diff
@@ -0,0 +1,108 @@
...
+EXAMPLES = """
+- name: We could read the file directly, but this shows output from command
+  ansible.builtin.debug:
+      msg: "CMK version: {{ lookup('checkmk.general.version', 'https://myserver/mysite', automation_user='automation', automation_secret='$SECRET'}}."
+"""
+
+RETURN = """
+  _list:
+    description:
+      - server CMK version
```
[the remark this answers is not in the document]
lgetwan: Corrected.

playbooks/demo/lookup.yml
lgetwan: I've put the lookup reply into a variable and used it in the msg.
That way, I was able to add line breaks to make the lookup call look nicer.

robin-checkmk: Love it!

robin-checkmk (approved):

Max-checkmk (approved):
Cool stuff! Collection is growing! ;)

PR #417, opened 2023-08-18
Add Lookup Plugins for hosts and folders

lgetwan:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

It's hard to query information about existing folders

## What is the new behavior?

- There's a lookup plugin that exposes the REST API endpoint /objects/folder_config/<folder>
- There's a lookup plugin that exposes the REST API endpoint /domain-types/folder_config/collections/all
- There's a lookup plugin that exposes the REST API endpoint /objects/host_config/<folder>
- There's a lookup plugin that exposes the REST API endpoint /domain-types/host_config/collections/all


robin-checkmk:
Does it make sense, to name `all_folders.py` just `folders.py`? Feels a little more efficient.

lgetwan:
> Does it make sense, to name `all_folders.py` just `folders.py`? Feels a little more efficient.

Sounds reasonable.
I just took that from the redoc, but with Ansible, it's legitimate to simplify things.

lgetwan:
OMG, finally...

lgetwan:
Depending on the release that we put this PR in, we need to change the "version_added: " value in the code.

lgetwan:
Shall we introduce the integration tests for thees plugins in this PR, directly?

msekania:
@lgetwan, @robin-checkmk 

thanks first of all for lookup plugins and "templates" for the future ones.

I am a bit concerned that the two plugins `folder` and `folders` differ only by one last character `s`. However, I do not have a better suggestion.
  
I see a lot of new stuff coming in this week. I'll be glad to test these new things when I'm back from vacation.

robin-checkmk:
> Shall we introduce the integration tests for thees plugins in this PR, directly?

Yes please. I will add the plugins to the auto labeller.

robin-checkmk:
The shenanigans with the labels should be gone after merge. I updated the matching already.

Is this PR ready @lgetwan or is it intentionally still in draft state?

lgetwan:
Yesterday, I spent almost the whole day getting the integration tests to work. I removed a bug in the folder module (not lookup!). Identifying that one was tricky.
Before Feierabend, I did a last git push. Looks like that was the first successful one, so I will remove the draft status now. :-)

plugins/modules/folder.py
```diff
@@ -237,8 +237,8 @@ def get_current_folder_state(module, base_url, headers):
     else:
         exit_failed(
             module,
-            "Error calling API. HTTP code %d. Details: %s."
```
robin-checkmk, line 240: There is no changelog mentioning the bug in the folder module.
We should proper document this though, right?

tests/integration/targets/lookup_folders/tasks/test.yml
```diff
@@ -0,0 +1,61 @@
...
+             automation_secret=automation_secret)
+         }}"
+  delegate_to: localhost
+  run_once: true  # noqa run-once[task]
+  loop_control:
+      label: "{{ item.id }}"
+
+- name: "Verify number of folders."
+  ansible.builtin.assert:
+    that: "( 1 + checkmk_folders|length ) == folders|length"
```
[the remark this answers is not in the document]
lgetwan: Done.
Reason: The list returned by the lookup plugin also contains the main folder, which is not in the variable that is used to initially create the folders.

plugins/modules/folder.py
lgetwan: You're right!

msekania:
@lgetwan, @robin-checkmk 

Just a design question/proposal.
It might be better to take `site` as a separate parameter (required) like in modules.
Anyway, there are extra required parameters for authentication, so why not to take `site` also.
IMHO it will be better than current string addition.

msekania:
@lgetwan, @robin-checkmk,

I have the following extension for all lookup plugins:
(I'll demonstrate it only for file `plugins/lookup/folder.py` lines `19-32`)
```yaml
      site_url:
        description: site url
        required: True
      automation_user:
        description: automation user for the REST API access
        required: True
      automation_secret:
        description: automation secret for the REST API access
        required: True
      validate_certs:
        description: Wether or not to validate TLS cerificates
        type: boolean
        required: False
        default: True
```
**could be changes as**

```yaml
      site_url:
        description: site url
        required: True
      vars:
        - name: ansible_lookup_checkmk_site_url
      env:
        - name: ANSIBLE_LOOKUP_CHECKMK_SITE_URL
      ini:
        - section: checkmk_lookup
          key: site_url
      automation_user:
        description: automation user for the REST API access
        required: True
      vars:
        - name: ansible_lookup_checkmk_automation_user
      env:
        - name: ANSIBLE_LOOKUP_CHECKMK_AUTOMATION_USER
      ini:
        - section: checkmk_lookup
          key: automation_user
      automation_secret:
        description: automation secret for the REST API access
        required: True
      vars:
        - name: ansible_lookup_checkmk_automation_secret
      env:
        - name: ANSIBLE_LOOKUP_CHECKMK_AUTOMATION_SECRET
      ini:
        - section: checkmk_lookup
          key: automation_secret
      validate_certs:
        description: Weather or not to validate TLS certificates
        type: boolean
        required: False
        default: True
      vars:
        - name: ansible_lookup_checkmk_validate_certs
      env:
        - name: ANSIBLE_LOOKUP_CHECKMK_VALIDATE_CERTS
      ini:
        - section: checkmk_lookup
          key: validate_certs
```
The advantage of the proposed version is that you can define these variables once e.g. as playbook variables (or in environment or ...) and then simply use the

```yaml
- name: As an example set variables here
  ansible.builtin.set_fact:
     ansible_lookup_site_url="{{ server_url + '/' + site }}"
     ansible_lookup_checkmk_automation_user="{{ automation_user }}"
     ansible_lookup_checkmk_automation_secret="{{ automation_secret }}"
     validate_certs="{{ False }}"

- name: Get the attributes of folder /tests
  ansible.builtin.debug:
    msg: "Attributes of folder /network: {{ lookup('checkmk.general.folder',  '~tests')  }}"
```
**instaed of**

```yaml
- name: Get the attributes of folder /tests
  ansible.builtin.debug:
    msg: "Attributes of folder /network: {{ attributes }}"
  vars:
    attributes: "{{
                    lookup('checkmk.general.folder',
                        '~tests',
                        site_url=server_url + '/' + site,
                        automation_user=automation_user,
                        automation_secret=automation_secret,
                        validate_certs=False
                        )
                 }}"
```

p.s. I will indeed appreciate if `site` will become a similar parameter and not a part of a string concatenation in jinja2 template. (`url` and `site` instead of `site_url`)

robin-checkmk:
@msekania while I can see your points, this is more like a general concept question and not specific to the lookup plugins, right? Because if that is the case, I would ask you to create a dedicated issue, so we can look at it holistically and not only for a specific use case.

msekania:
@robin-checkmk actually they are specific to lookup plugins, e.g., see code/doc-s of of existing ansible lookup plugins, and also examples on the ansible documentation page.

msekania:
@robin-checkmk, 
concerning the second point,
everything is already in the comment/report, with the usage example at the end. I can also wait when everything is merged to the main branch and make a separate PR for all 4 lookup plugin modules

msekania:
@robin-checkmk  Ah, sorry, I thought you meant the second comment.

anyway, in other modules (at least plugin/modules) we have `server_url` and `site` and not `site_url`, or do I misseing something here.
For examle, we have:
```yaml
- name: "Cerate some rule"
  checkmk.general.rule:
    server_url: "{{ test_setver_url }}"
    site: "{{ site }}"
    ruleset: "discovery_systemd_units_services_rules"
...
```

and not 

```yaml
- name: "Cerate some rule"
  checkmk.general.rule:
    server_url: "{{ test_setver_url }}/{{ site }}"
    ruleset: "discovery_systemd_units_services_rules"
...
```

robin-checkmk:
> anyway, in other modules (at least plugin/modules) we have `server_url` and `site` and not `site_url`, or do I missing something here?

My bad, you are right. I did not look it up, my mind tricked me.
Let's wait for @lgetwan to see, what he thinks about the proposed changes.

robin-checkmk (approved):

lgetwan:
@msekania:
Thanks a lot for your input! 
Initially, I wanted to keep the number of variables needed to call a lookup plugin as low as possible. But using "site" and "server_url" makes sense.
I also tried to test the vars/env/ini feature, but it wasn't as simple as I thought. So I suggest to keep that for later. Maybe, you can take over that task, once this PR is in the main branch.

@robin-checkmk:
Regarding the failing sanity test:
This is related to the "folder" module, not the "folder lookup" module, So I will not fix this here.

robin-checkmk:
@lgetwan makes sense. So from your end, we are good to go?

lgetwan:
I'd say so. :+1:

Max-checkmk (approved):

PR #418, opened 2023-08-21
Handle 409 response

Max-checkmk:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

In case the API returns
    - 409 "Conflict: A discovery background job is already running" for single mode
    - 409 "Conflict: A bulk discovery job is already active" in bulk mode
the job is failed and ansible stops.

## What is the new behavior?

- We check for the 409 response and try again every 0.5 seconds until running the discovery is possible again.

## Other information

This PR is mainly for a bug where the discovery endpoint doesn't respond with a redirect and the completion endpoint answers with 404 "no running discovery found"

In case of the bulk discovery it would be worth a discussion if waiting as long as it needs is the right option.
Alternatives would be
    - 409 results in a failed task and stops the playbook (current behavior).
    - 409 results not in a failed task, but also there are no changes made.

### DIFF
--- a/plugins/modules/discovery.py
+++ b/plugins/modules/discovery.py
@@ -129,6 +129,7 @@ HTTP_CODES = {
     403: (False, True, "Forbidden: Configuration via WATO is disabled."),
     404: (False, True, "Not Found: Host could not be found."),
     406: (False, True, "Not Acceptable."),
+    409: (False, False, "Conflict: A discovery background job is already running"),
     415: (False, True, "Unsupported Media Type."),
     500: (False, True, "General Server Error."),
 }
@@ -153,7 +154,7 @@ HTTP_CODES_BULK = {
     400: (False, True, "Bad Request."),
     403: (False, True, "Forbidden: Configuration via WATO is disabled."),
     406: (False, True, "Not Acceptable."),
-    409: (False, True, "Conflict: A bulk discovery job is already active"),
+    409: (False, False, "Conflict: A bulk discovery job is already active"),
     415: (False, True, "Unsupported Media Type."),
     500: (False, True, "General Server Error."),
 }
@@ -351,6 +352,14 @@ def run_module():
     result = wait_for_completion(single_mode, servicecompletion)
 
     result = discovery.post()
+
+    # In any case the API returns 409 (discovery running) we wait for half a second and try again.
+    # This can happen as example in versions where the endpoint doesn't respond with the correct redirect.
+    while (single_mode and result.http_code == 409) or (
+        len(module.params.get("hosts", [])) > 0 and result.http_code == 409
+    ):
+        time.sleep(0.5)
+        result = discovery.post()
 
     # If single_mode and the API returns 302, check the service completion endpoint
     # If not single_mode and the API returns 200, check the service completion endpoint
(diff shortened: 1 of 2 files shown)


plugins/modules/discovery.py
```diff
@@ -352,6 +353,14 @@ def run_module():
 
     result = discovery.post()
 
+    # In any case the API returns 409 (discovery running) we wait for half a second and try again.
+    # This can happen as example in versions where the endpoint doesn't respond with the correct redirect.
+    while (single_mode and result.http_code == 409) or (
+        len(module.params.get("hosts", [])) > 0 and result.http_code == 409
+    ):
+        time.sleep(0.5)
```
robin-checkmk, line 361: Considering, that this means bombarding the Checkmk API with a lot of request, we should probably dial it down to 5 to 10 seconds, maybe even more.

tests/integration/targets/discovery/tasks/test.yml
```diff
@@ -17,7 +17,7 @@
 
 - name: "Run Single Discoveries."
   block:
-    - name: "{{ outer_item.version }} - Refresh (Tabula Rasa in 2.1 and before, Rescan in 2.2 and newer)."
+    - name: "{{ outer_item.version }} - Rescan services."
```
robin-checkmk, line 20: Why do you change this?

tests/integration/targets/discovery/tasks/test.yml
```diff
@@ -145,7 +145,7 @@
       run_once: true  # noqa run-once[task]
       loop: "{{ checkmk_hosts }}"
 
-    - name: "{{ outer_item.version }} - Bulk: Refresh (Tabula Rasa in 2.1 and before, Rescan in 2.2 and newer)."
+    - name: "{{ outer_item.version }} - Bulk: Rescan services."
```
robin-checkmk, line 148: Why do you change this?

robin-checkmk (changes requested):

tests/integration/targets/discovery/tasks/test.yml
Max-checkmk: Like in 2.2 this is also just a rescan in 2.1. Just the description in the API Docs is wrong in 2.1.

tests/integration/targets/discovery/tasks/test.yml
Max-checkmk: Like in 2.2 this is also just a rescan in 2.1. Just the description in the API Docs is wrong in 2.1.

plugins/modules/discovery.py
Max-checkmk: I wouldn't got too high. At least with a single host service discovery this shouldn't take too long.
One further option would be to use different wait times for single and bulk. What do you think?

tests/integration/targets/discovery/tasks/test.yml
robin-checkmk: I remember. Agreed.

plugins/modules/discovery.py
robin-checkmk: As far as I can see, you only handle single mode currently, right?

Well, I still think 5 seconds is quite a low interval. For bulk discoveries I would go up to 10 to 30 seconds.

Max-checkmk: So with the last commit it's 1 second for single and 10 for bulk.

robin-checkmk (approved):

PR #426, opened 2023-08-25
New time period module

Max-checkmk:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

There is no module for time periods

## What is the new behavior?

- Added a module for creating, updating and deleting time periods

### DIFF
--- a/plugins/modules/timeperiod.py
+++ b/plugins/modules/timeperiod.py
@@ -60,22 +60,52 @@ author:
 
 EXAMPLES = r"""
 # Creating and Updating is the same.
-- name: "Create a new time period."
+- name: "Create a new time period. (Attributes in one line)"
   checkmk.general.timeperiod:
     server_url: "http://localhost/"
     site: "my_site"
     automation_user: "automation"
     automation_secret: "$SECRET"
     name: "worktime"
-    title: "Work time"
-    active_time_ranges: [
-    exceptions: [
+    title: "Worktime"
+    active_time_ranges: '[{"day": "all", "time_ranges": [{"start": "09:00:00", "end": "17:00:00"}]}]'
+    exceptions: '[{"date": "2023-12-24", "time_ranges": [{"start": "10:00:00", "end": "12:00:00"}]}]'
+    exclude: '[ "Lunchtime" ]'
+    state: "present"
 
-    ]
-    ]
+- name: "Create a new time period. (Attributes in multiple lines)"
+  checkmk.general.timeperiod:
+    server_url: "http://localhost/"
+    site: "my_site"
+    automation_user: "automation"
+    automation_secret: "$SECRET"
+    name: "worktime"
+    title: "Worktime"
+    active_time_ranges: [
+              {
+                  "day": "all",
+                  "time_ranges": [
+                      {
+                          "start": "8:00",
+                          "end": "17:00"
+                      }
+                  ]
+              },
+          ]
+    exceptions: [
+              {
+                  "date": "2023-12-24",
+                  "time_ranges": [
+                      {
+                          "start": "8:00",
+                          "end": "12:00"
+                      }
+                  ]
+              },
+          ]
     exclude: [
-        "lunch"
-    ]
+         "Lunchtime"
+          ]
     state: "present"
 
 - name: "Delete a time period."
@@ -297,13 +327,13 @@ def patched_version(checkmkversion):
     if (
         checkmkversion[0] == "2"
         and checkmkversion[1] == "2"
-        and int(checkmkversion[2]) >= 9
+        and int(checkmkversion[2].replace("0p", "", 1)) >= 9
     ):
         return True
     if (
         checkmkversion[0] == "2"
         and checkmkversion[1] == "1"
-        and int(checkmkversion[2]) >= 33
+        and int(checkmkversion[2].replace("0p", "", 1)) >= 33
     ):
         return True
     return False
@@ -347,6 +377,7 @@ def run_module():
         if result.http_code == 200:
             checkmkversion = timeperiodget.getversion()
 
+            # Check for value "exclude" as this wasn't possible to update before Werk #16052.
             if module.params.get("exclude") and not patched_version(checkmkversion):
                 result = RESULT(
                     http_code=0,
@@ -360,6 +391,7 @@ def run_module():
 
             timeperiodupdate = TimeperiodUpdateAPI(module)
             timeperiodupdate.headers["If-Match"] = result.etag
+
             # Different output of "Show a time period" in Version 2.0
             if checkmkversion[0] == "2" and checkmkversion[1] == "0":
                 existingalias = json.loads(result.content).get("alias")
--- a/plugins/modules/timeperiod.py
+++ b/plugins/modules/timeperiod.py
@@ -250,6 +250,8 @@ HTTP_CODES_UPDATE = {
     500: (False, True, "General Server Error."),
 }
 
+updatevalues = ("alias", "active_time_ranges", "exceptions", "exclude")
+
 
 class TimeperiodCreateAPI(CheckmkAPI):
     def post(self):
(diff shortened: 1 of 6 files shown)


lgetwan (changes requested):
The documentation could be a bit more detailed. Probably, you could use the tasks from your integration test.

I tested the module with some use cases:
- When creating a timeperiod that already exists, a failure is shown (should be "OK").
- When updating a timeperiod without any actual change, the result is "changed" (should be "OK").
- When adding an exclude time range, this isn't visible in the UI.
- When deleting a timeperiod that doesn't exist, a failure is shown (should be "OK").

Here's the playbook i used:
```---

# Make sure to keep the tests in sync!
# Rather use the tests than this playbook directly.

- name: "Test timeperiods"
  hosts: all
  gather_facts: 'no'
  vars_files:
    - ./vars/config.yml

  tasks:
  - name: create timeperiod
    checkmk.general.timeperiod:
      site: "{{ site }}"
      name: "lunch"
      alias: "Mahlzeit"
      active_time_ranges: [
              { 
                  "day": "all",
                  "time_ranges": [
                      {
                          "start": "12:00",
                          "end": "13:00"
                      }
                  ]
              }
          ]
      state: present
    delegate_to: localhost
    run_once: 'yes'

  # fails. should be "OK"
  - name: create timeperiod again
    checkmk.general.timeperiod:
      site: "{{ site }}"
      name: "lunch"
      alias: "Mahlzeit"
      active_time_ranges: [
              { 
                  "day": "all",
                  "time_ranges": [
                      {
                          "start": "12:00",
                          "end": "13:00"
                      }
                  ]
              }
          ]
      state: present
    delegate_to: localhost
    run_once: 'yes'

  - name: create timeperiod that uses the first one
    checkmk.general.timeperiod:
      site: "{{ site }}"
      name: "working_hours"
      alias: "Ballern"
      active_time_ranges: [
              { 
                  "day": "all",
                  "time_ranges": [
                      {
                          "start": "8:00",
                          "end": "17:00"
                      }
                  ]
              },
          ]
      exclude: ["Mahlzeit"]
      state: present
    delegate_to: localhost
    run_once: 'yes'

  - name: update timeperiod
    checkmk.general.timeperiod:
      site: "{{ site }}"
      name: "working_hours"
      active_time_ranges: [
              { 
                  "day": "all",
                  "time_ranges": [
                      {
                          "start": "9:00",
                          "end": "18:00"
                      }
                  ]
              },
          ]
      state: present
    delegate_to: localhost
    run_once: 'yes'

  # fails. should be "OK"
  - name: update timeperiod. again.
    checkmk.general.timeperiod:
      site: "{{ site }}"
      name: "working_hours"
      active_time_ranges: [
              { 
                  "day": "all",
                  "time_ranges": [
                      {
                          "start": "9:00",
                          "end": "18:00"
                      }
                  ]
              },
          ]
      state: present
    delegate_to: localhost
    run_once: 'yes'

  - name: remove used timeperiod
    checkmk.general.timeperiod:
      site: "{{ site }}"
      name: "lunch"
      state: absent
    delegate_to: localhost
    run_once: 'yes'
    ignore_errors: true

  - name: remove timeperiod
    checkmk.general.timeperiod:
      site: "{{ site }}"
      name: "working_hours"
      state: absent
    delegate_to: localhost
    run_once: 'yes'

  # fails. should be "OK"
  - name: remove timeperiod again
    checkmk.general.timeperiod:
      site: "{{ site }}"
      name: "working_hours"
      state: absent
    delegate_to: localhost
    run_once: 'yes'

  - name: remove used timeperiod
    checkmk.general.timeperiod:
      site: "{{ site }}"
      name: "lunch"
      state: absent
    delegate_to: localhost
    run_once: 'yes'
```

Max-checkmk:
> The documentation could be a bit more detailed. Probably, you could use the tasks from your integration test.
> 
> I tested the module with some use cases:
> 
>     * When creating a timeperiod that already exists, a failure is shown (should be "OK").
> 
>     * When updating a timeperiod without any actual change, the result is "changed" (should be "OK").
> 
>     * When adding an exclude time range, this isn't visible in the UI.
> 
>     * When deleting a timeperiod that doesn't exist, a failure is shown (should be "OK").
> 
> 
> ```

Hi Lars,

thanks for the feedback. I had a dumb mistake in it, which is now corrected.
And i forgot to finish the examples, which is also done now.

Please try it again.
What i didn't change is the error which is thrown, if someones tries to delete something what doesn't exist.
Not sure about this...

Nice weekend! ;)

msekania:
> What i didn't change is the error which is thrown, if someones tries to delete something what doesn't exist.
> Not sure about this...

@Max-checkmk,

`state: absent` means that it's not there, that means that someone does not try to delete something but rather to ensure that it's not there, hence no error.

plugins/modules/timeperiod.py
```diff
@@ -0,0 +1,444 @@
...
+                    msg="Can't update exclude value in Checkmk 2.2.0p8 or 2.1.0p32 and before. See Werk #16052",
+                    content="",
+                    etag="",
+                    failed=True,
+                    changed=False,
+                )
+                module.fail_json(**result_as_dict(result))
+
+            timeperiodupdate = TimeperiodUpdateAPI(module)
+            timeperiodupdate.headers["If-Match"] = result.etag
```
lgetwan, line 393: An Update should only be done in the case the mentioned parameters are different in the already existing Timeperiod. 
Otherwise running the same playbook again and again will always do a change in the environment. You have to avoid this.

lgetwan (changes requested):

lgetwan:
Nice weekent y'all!

Max-checkmk:
> `state: absent` means that it's not there, that means that someone does not try to delete something but rather to ensure that it's not there, hence no error, rather no change. And changed if it was there and deleted.

You're right. I just had a look at some other modules and as far i see at the moment we're not really consistent for all modules here. ;)
I will do the change for this module with the next commit!

plugins/modules/timeperiod.py
Max-checkmk: Nice idea, but this will need a lot of time to accomplish.

So as example we want to update a time period for all weekdays with these time ranges:
`active_time_ranges: '[{"day": "all", "time_ranges": [{"start": "08:00:00", "end": "17:00:00"}]}]'`

And we try to compare this with the existing time ranges of a time period, we get this:
`[{'day': 'monday', 'time_ranges': [{'start': '08:00', 'end': '17:00'}]}, {'day': 'tuesday', 'time_ranges': [{'start': '08:00', 'end': '17:00'}]}, {'day': 'wednesday', 'time_ranges': [{'start': '08:00', 'end': '17:00'}]}, {'day': 'thursday', 'time_ranges': [{'start': '08:00', 'end': '17:00'}]}, {'day': 'friday', 'time_ranges': [{'start': '08:00', 'end': '17:00'}]}, {'day': 'saturday', 'time_ranges': [{'start': '08:00', 'end': '17:00'}]}, {'day': 'sunday', 'time_ranges': [{'start': '08:00', 'end': '17:00'}]}]`

So we would need to adjust
- the given time ranges to every single day of the week
- remove the seconds from the input, as internally there are no seconds (in contrast to the API Docs example...)

We redirect the raw data to the API so at first you would have to convert the data to something you can work with in python.

lgetwan: I see, that's too complex.

But what do you think of this as an intermediate solution:
If the data provided by the task is completely equal with the current data from the API, we don't change anything. But if the task e.g. has seconds in the data, and the API has not, we do post a change to the API.

Many users are running their playbooks daily or hourly, and that would help them a lot, already.

Max-checkmk: Please have a look at my last commit. Now there is a rudimentary comparison between existing and new values.
Based on that we could add the other checks (full week and seconds) sometime.

lgetwan (approved):
Works as described. :-)

robin-checkmk:
@Max-checkmk please check the failing sanity check and fix it. I think we had this issue with the host or folder module in the past. Maybe @lgetwan can help here.

Max-checkmk:
> @Max-checkmk please check the failing sanity check and fix it. I think we had this issue with the host or folder module in the past. Maybe @lgetwan can help here.

Well, the problem seems to be inside the folder-module and not in this module. As far as i can read the logs from the test...

robin-checkmk:
True. Then we will ignore the failing test here and fix it in `devel` directly.

Issue #65, opened 2022-06-29
[FEED] Support CME


lgetwan:
In many API endpoints, the CME has "customer" as a mandatory field. We have to take care of this.

robin-tribe29:
Note: CME support can affect the agent role too (See #121 for reference). We should keep an eye on that.

edermi:
I assume CME refers to the CheckMK Enterprise edition? If yes, How is the agent deployment support going? I'd like to avoid having to install the agent on all my systems by hand.

robin-tribe29:
Hi @edermi, no CME is the Checkmk Managed Service Edition. This issue focuses on modules, as they need to be able to handle an additional attribute. Regarding your question about the Agent rollout: If you are using the CEE (Enterprise Edition) you are good to go. If you are actually using the CME, I would need to check a thing or two.

edermi:
Thanks, understood.
[The agent README](https://github.com/tribe29/ansible-collection-tribe29.checkmk/blob/main/roles/agent/README.md) states that CheckMK Enterprise Edition is not supported, hence I was reluctant to try it. I'd like to avoid this issue going off-topic, is there any further documentation? I'm especially interested in how this plays with the bakery and auto-updates.

robin-tribe29:
#165 adds CME support for the agent role.

robin-checkmk:
This is still relevant, hence I marked it to never become stale.

muehlings:
I had to patch two modules host_group.py and user.py yesterday for my transition from multiple scripts and manual doings to ansible. Is it correct that Checkmk expects the customer always as lowercase string ("provider" despite the fact it's written "Provider" in my backend)?

Here are my changes (just in case I do a silly update of my local collection):

host_group.py
```
def create_single_host_group(module, base_url, headers):
    name = module.params["name"]

    api_endpoint = "/domain-types/host_group_config/collections/all"
    params = {
        "name": name,
        "alias": module.params.get("title", name),
        "customer": module.params.get("customer", 'provider'),
    }
    url = base_url + api_endpoint

    response, info = fetch_url(
        module, url, module.jsonify(params), headers=headers, method="POST"
    )

    if info["status"] != 200:
        exit_failed(
            module,
            "Error calling API. HTTP code %d. Details: %s, "
            % (info["status"], info["body"]),
        )


def create_host_groups(module, base_url, groups, headers):
[...]

PR #427, opened 2023-08-25
Add "customer" parameter in module and request

muehlings:
Adds the neccessary parameter "customer" to the corresponding CME API endpoints in the ansible modules.

- user.py
- host_group.py
- service_group.py
- contact_group.py
- password.py

## Pull request type

- [x] Feature

Issue Number: `#65`

## What is the new behavior?

- Make ansible modules ready for Checkmk CME API endpoints.

TODO:

- Identify all modules who need the "customer" parameter.
- Modify the POST/PUT logic? I think the other Checkmk versions will fail at requests with additional parameters.
- ~~Problems with host.py in 2.2.0p8 CEE detected (API endpoint change which prevents ipaddress+alias+tags?)~~(was related to the discovery module, because of an exception the create/change was not applied)
- Problems with auth_type in user API endpoint? Wrong default when not set?


muehlings:
A default to `customer: "provider"` will not work, because the extra field "customer" is not supported in non-CME environments.
I tried to add the logic in the module: 
grafik

My playbook decides wether the parameter should be given to the module:
grafik

Then the only thing you need to pay attention to is the inventory. For non-CME site don't add the customer in the vars. I'm not using the bulk methods. I use dictionaries and then loop over the items.

A new approach could be the lookup module for version. Then we can look up it the parameter is neccessary or not.

robin-checkmk:
@muehlings you are correct and to me your code looks sensible. A review pending I am good with the solution. Once that is sorted out, I will take care of the roles and such, which need updates as well. Just like you pointed out, when it comes to the inventory.

plugins/modules/user.py
```diff
@@ -427,9 +437,11 @@ def run_module():
         automation_secret=dict(type="str", required=True, no_log=True),
         name=dict(required=True, type="str"),
         fullname=dict(type="str"),
+        customer=dict(type="str"),
         password=dict(type="str", no_log=True),
         enforce_password_change=dict(type="bool", no_log=False),
-        auth_type=dict(type="str", choices=["password", "automation"]),
+        # changed 'auth_type' to required=True, because endpoint won't create usable user without auth_type
+        auth_type=dict(type="str", choices=["password", "automation"], required=True),
```
robin-checkmk, line 444: Two questions: Did this ever work? And was there a (implicit) default value?
The reason for the questions: This could be a breaking change, if it worked before without specifying the `auth_type`.

muehlings: I did not explicitly supply the auth_type and the customer was created, because the API endpoint does not require the auth_type. The user gets created and the change-password-task tells me that the user was changed.
```
TASK [Create a user.] ***
ok: [SMUE] => (item=testuser)

TASK [Set user password.] ***
changed: [SMUE] => (item=testuser)
```
The user is created with the right authentication method but is not usable, because the password could not be set:
grafik
So there seems to be no implicit default.

See documentation (2.2.0p7 CME):
grafik
Required: username, fullname, customer

There a several options:
- Modify the API endpoint and make "auth_type" a required parameter (because the user is not fully created or usable, even if you supplied a password) [but then the module requires at least that Checkmk version]
- Workaround in the module as i did (make auth_type a required field even if the endpoint does not enforce it)
- Combination of both: Use the workaround now and wait for the changes in the endpoint. Then the workaround becomes obsolete in the future. Perhaps we should add the failsafe default 'password' in the module as this should be the normal behavior an enduser would expect. [this should be the preferred option]

robin-checkmk: I might be missing something, but we could also go for a default value of `password`, right? The bug only requires, that we provide an `auth_type`, right?

We should find a solution that works as stable as possible and potentially avoid current or future bugs here.

But I also think it can be sensible to make the `auth_type` a required parameter.

Maybe a brief voting by reacting to this comment (ping @lgetwan @Max-checkmk):
:tada: means: Default to `password` and do not make `auth_type` required.
:rocket:  means: Make `auth_type` required.

P.S.: @muehlings can you look at the integration tests for the user module in the meantime? I added the `auth_type` to some tests and now they fail hard. But I do not understand the failure.

muehlings: Just a guess:
User.from_module(module.params) wants to return params["state"] but params has no key "state" to return.
Should be line 487 in user.py:
`    desired_user = User.from_module(module.params)`
wants to return
`        return cls(params["name"], state=params["state"], attributes=attributes)`

I currently have no idea how to debug at my side, because I don't see this error.

muehlings: I think I left the auth_type as it is (for the tests). 
I'm successfully running the integration test for the user module atm:
grafik
I'm setting the customer variable when it is neccessary and omit it in case of CRE, CFE, CEE.

The YAML wasn't intuitive. I set
```
- name: "{{ outer_item.version }} - Unset customer when needed."
  set_fact:
    customer: null
  when: outer_item.edition == "cre"
```
and had to check for "None"
`customer: "{{ (customer != None ) | ternary(customer,omit) }}"`
and under the hood "null" was "" (empty string).

I tested all easier approaches like `customer: {{customer|default(omit)}}` but all these variants failed, because setting `customer: null` 
set 
`customer: ""` 
instead of unsetting the variable. 

I think I will need some more time to publish it.

tests/integration/targets/user/vars/main.yml
```diff
@@ -1,21 +1,24 @@
...
     site: "stable_ent"
+  - version: "2.2.0p12"
+    edition: "cre"
+    site: "stable_raw"
   - version: "2.1.0p35"
     edition: "cre"
     site: "old_raw"
 
-server_url: "http://127.0.0.1/"
+server_url: "https://maas.pdv-systeme.de/"
```
robin-checkmk, line 16: You probably do not want to set your internal domain as a default. :)

tests/integration/targets/user/vars/main.yml
```diff
@@ -1,21 +1,24 @@
...
 automation_user: "cmkadmin"
 automation_secret: "d7589df1"
 
 download_url: "https://download.checkmk.com/checkmk/{{ item.version }}/check-mk-{{ checkmk_server_edition_mapping[item.edition] }}-{{ item.version }}_0.{{ ansible_distribution_release }}_amd64.deb"  # noqa yaml[line-length]
-download_user: "d-gh-ansible-dl"
+download_user: "hotline@pdv-systeme.de"
```
robin-checkmk, line 21: This looks like a company-specific user.

plugins/modules/password.py
```diff
@@ -33,6 +33,11 @@
         required: false
         type: str
 
+    customer:
+        description: For the Checkmk Managed Edition, you need to specify which customer a password belongs to.
+        required: false
+        type: str
```
robin-checkmk, line 39: Is there a reason, that this description differs from the description for the other modules?
Maybe go with the following sentence for all modules?
`When using the Checkmk Managed Edition you need to specify which customer this object belongs to.`

robin-checkmk (changes requested):

plugins/modules/user.py
robin-checkmk: Did you also try something like this?
`{% if customer is defined and customer | length %}customer{% else %}""{% endif %}`?
Anyway, I am uncertain, if I really understand the problem.

Let me try to summarize:

- A user is not using the CME and omits the `customer` attribute: All good.
- A user is not using the CME but provides the `customer` attribute: Ugly crash in Checkmk. We should catch that in the module.
- A user is using the CME and omits the `customer` attribute: Expected error.
- A user is using the CME and provides the `customer` attribute: All good.

If those assumptions are correct, then I am uncertain if we even need the logic you implemented in the tests. We could simply use two blocks. One tests the CME with the `customer` attribute, one tests the others without the attribute. That way, we do not need to implement magic in the tests. Because the level of convenience you are aiming for here, would make sense in a role, that should be ready to use. But in the tests, we are in control of what happens.

TL;DR: I feel like we are good with the current behavior regarding the `customer` attribute. I can take care of the tests for you. And I would add some documentation, so the expected input is clear. Does that make sense to you @muehlings?

Also, we should catch the second scenario and print a nice message like "You tried to pass the `customer` attribute to a non-CME site. Please review the documentation." or so.

tests/integration/targets/user/vars/main.yml
robin-checkmk: I cleaned it up, as I needed the proper value for the tests to work on GitHub.

plugins/modules/user.py
muehlings: Hello Robin!
Answer to TL;DR: Everything ok from my side. Please fix the tests and documentation if neccessary.

robin-checkmk (approved):

robin-checkmk:
@lgetwan or @Max-checkmk can one of you take a look at the code and give a second opinion about this change please? Thanks!

Issue #437, opened 2023-09-12
[BUG] Prort check for agent port runs unconditionally, even if push agent is used


hb9hnt:
In version 3.2.0 a port check for TCP port 6556 is introduced in

https://github.com/Checkmk/ansible-collection-checkmk.general/blob/0925d6fd69e0d986a259e3b9d094019ce6d28148/roles/agent/tasks/Linux.yml#L106

This port check has no condition regarding whether the agent is set to push or pull mode, it runs in every case, even if the installed agent is set to push.

I'll omit the full data for my setup for now because I think it is not necessary because this bug is IMHO obvious. Please correct me if I'm wrong and I'll add all the details requested.

robin-checkmk:
Thank you for your feedback @hb9hnt! The feature well intended and necessary for Windows host, which seem to be dragging. I thought it to be a good general feature. But I did not take the agent in push mode into consideration. I also did not take custom ports into consideration.

Do you think it makes sense to make this feature enabled by default, but configurable? Because from the top of my mind I see no easy way to learn if the agent is in push or pull mode. Of course, it is possible, but that would involve querying the Checkmk server API, which in turn might be undesirable.

hb9hnt:
Hey @robin-checkmk - IMHO it would be ok to introduce a variable `checkmk_agent_push` or similar that can be set to true if you deploy a push agent. 

We discovered a second place where this variable can be put to use: In case of a push agent it makes sense to trigger a initial

```sh
cmk-agent-ctl push 
```

on the machine before running a discovery on Checkmk.

hb9hnt:
@robin-checkmk No worries, I already introduced said variable in my local copy and will continue to use this until your more general approach is ready.

robin-checkmk:
@hb9hnt if you already got something, maybe you fancy creating a PR and become a contributor? We can pan out the details in the PR then. :) No pressure, just a suggestion.

hb9hnt:
@robin-checkmk It really just a quick and dirty variable and two checks. I can submit this as PR but I'm sure your general approach is the way to go.

hb9hnt:
Ok, I created a PR with what I did. Feel free to turn it down and replace it with a better approach. ;)

https://github.com/Checkmk/ansible-collection-checkmk.general/pull/440

It's close to your draft. However I also threw in the initial push.

PR #440, opened 2023-09-18
Fix agent role for push agents

hb9hnt:
## Pull request type

- [x] Bugfix (kindof)
- [x] Feature
- [ ] Build related changes

There is an inbound port check to the host which runs no matter whether we use pull or push agents. Also, if you reploy a push agent, initial discovery will fail, since there is no data.

Issue Number: `#437` 

## What is the new behavior?

- There is a new variable `checkmk-agent-push' which defaults to false.
- If set to `true` the port scan is skipped.
- If set to `true` we trigger a initial push using `cmk-agent-ctl push` beforce running discovery.

## Other information
As said in the issue, that's just the approach I took. I'm not saying it is a clean or general approach. It's confusing because it introduces a variable called `checkmk-agent-push` which doesn't actually activate push agents. Therefore it duplicates information that already has to be present as setting on the target folder.

### DIFF
--- a/roles/agent/defaults/main.yml
+++ b/roles/agent/defaults/main.yml
@@ -31,6 +31,7 @@ checkmk_agent_folder: "{{ checkmk_folder_path | default('/') }}"
 checkmk_agent_force_foreign_changes: 'false'
 checkmk_agent_host_attributes:
   ipaddress: "{{ checkmk_agent_host_ip | default(omit) }}"
+checkmk_agent_push: false
 
 
 # If you trust your local hostnames, you could also use the following
--- a/roles/agent/defaults/main.yml
+++ b/roles/agent/defaults/main.yml
@@ -31,7 +31,7 @@ checkmk_agent_folder: "{{ checkmk_folder_path | default('/') }}"
 checkmk_agent_force_foreign_changes: 'false'
 checkmk_agent_host_attributes:
   ipaddress: "{{ checkmk_agent_host_ip | default(omit) }}"
-checkmk_agent_mode: pull 
+checkmk_agent_mode: pull
 
 
 # If you trust your local hostnames, you could also use the following
--- a/roles/agent/defaults/main.yml
+++ b/roles/agent/defaults/main.yml
@@ -31,7 +31,7 @@ checkmk_agent_folder: "{{ checkmk_folder_path | default('/') }}"
 checkmk_agent_force_foreign_changes: 'false'
 checkmk_agent_host_attributes:
   ipaddress: "{{ checkmk_agent_host_ip | default(omit) }}"
-checkmk_agent_push: false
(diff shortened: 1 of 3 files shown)


robin-checkmk:
recheck

roles/agent/defaults/main.yml
```diff
@@ -31,6 +31,7 @@ checkmk_agent_folder: "{{ checkmk_folder_path | default('/') }}"
 checkmk_agent_force_foreign_changes: 'false'
 checkmk_agent_host_attributes:
   ipaddress: "{{ checkmk_agent_host_ip | default(omit) }}"
+checkmk_agent_push: false
```
robin-checkmk, line 34: I feel like `checkmk_agent_mode: pull` would be more natural. What do you think?

robin-checkmk:
@hb9hnt would you please sign the CLA and take a look at the failed tests? I think both are minor things, but they prevent us from moving forward. Thanks!

hb9hnt:

hb9hnt:
@robin-checkmk Sorry, I completely forgot about this pull request. I fixed the linting issue and renamed the variable according your suggestion - which I completely agree with.

roles/agent/tasks/Linux.yml
```diff
@@ -107,3 +107,4 @@
   ansible.builtin.wait_for:
     port: 6556
     timeout: 60
+  when: checkmk_agent_mode != 'push'
```
robin-checkmk, line 110: Not sure, if the conditional should be noted positively ( == pull), but as there is no objective answer, I think it is okay.

robin-checkmk (approved):
I will take some moments next week to think about other places, where we would need to handle the push mode. But what we got so far is fine and ready to merge. So it will definitely make the next release.

roles/agent/tasks/Linux.yml
thorian93: Okay, looking at this now, I realize, that there is a good reason to make this `when: checkmk_agent_mode == "pull"`, as that also covers the use-case mentioned below.

hb9hnt:
@thorian93 / @robin-checkmk Since `checkmk_agent_mode: ssh` isn't used anywhere a hint in the readme should be enough. Maybe we could also add a comment to the `defaults/main` file.

Do you want me to add this to the PR?

robin-checkmk:
@hb9hnt I do not want to trouble you with documentation, so I can gladly do that part. If you just change the conditional from the comment above, then we are good.

roles/agent/tasks/Linux.yml
robin-checkmk: @hb9hnt this one. Keeping the mode "ssh" in mind, this conditional should read when: `checkmk_agent_mode == "pull"`.

PR #448, opened 2023-10-09
Fetch checkmk agent from delegated node

mueller-ma:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently delegation of agent download is broken if `delegate_to` is not the Ansible server.

## What is the new behavior?

I have the following setup:
- A: Checkmk server
- B: Host that should be configured
- C: Host running Ansible
- D: Host already configured with Ansible

It's possible to delegate the download of the agent to a different host (D).

The tasks `Download * Agent` download the agent from A to D. The next task `Transfer GENERIC or folder-specific * Agent` copies the agent from C to B, which fails if C and D are different hosts. Therefore I added a fetch task that copies the agent from D to C.

## Other information

In some scenarios it's not (easily) possible to delegate the agent download to the Ansible server.

### DIFF
--- a/roles/agent/tasks/Debian.yml
+++ b/roles/agent/tasks/Debian.yml
@@ -93,6 +93,20 @@
   tags:
     - download-package
 
+- name: "{{ ansible_os_family }} Derivatives: Fetch GENERIC or folder-specific {{ checkmk_agent_edition | upper }} Agent."
+  ansible.builtin.fetch:
+    src: "{{ checkmk_agent_agent.file.cee }}"
+    dest: "{{ checkmk_agent_agent.file.cee }}"
+    flat: true
+    mode: "0644"
+  when: |
+    checkmk_agent_edition | lower != "cre"
+    and not checkmk_agent_host_specific | bool
+    and checkmk_agent_delegate_download != inventory_hostname
+  delegate_to: "{{ checkmk_agent_delegate_download }}"
+  tags:
+    - download-package
+
 - name: "{{ ansible_os_family }} Derivatives: Transfer GENERIC or folder-specific {{ checkmk_agent_edition | upper }} Agent."
   ansible.builtin.copy:
     src: "{{ checkmk_agent_agent.file.cee }}"


robin-checkmk:
@mueller-ma two questions:
1. What version of the collection are you using?
2. Does this issue only exist for Debian derivates, or does it also affect the other OS families?

mueller-ma:
I'm using the latest version 3.2.0. I haven't tested it on other OS, but from looking at the code this might be an issue there as well.

robin-checkmk:
@mueller-ma I worked through your problem and proposed solution for some time and I have two outcomes: Either you are misunderstanding how the role works, or I do not understand your problem.

Let me elaborate:
The tasks in line 51 and 78 already handle the downloading, while the task in line 96 transfers the agent to the target host, if the `checkmk_agent_delegate_download` is not the target host. All lines refer to the current `devel` branch. So we opted for two approaches: Direct download to host B or download to host C and upload from there to host B.
Your suggested code change just adds our task from line 96 but in reversed direction, meaning that host B fetches it from host C.

What I still fail to grasp: Where is the difference between host C and D? Or better, **why** is there a difference between those two?

mueller-ma:
The current tasks do the following:
- Lines 51 and 78: Download the agent from the checkmk server to the server `checkmk_agent_delegate_download`
- Line 96: Transfer the agent from the Ansible server to `inventory_hostname`

As long as the Ansible server is the same as `checkmk_agent_delegate_download` the role works fine. However in my environment this isn't the case. I'm using Windows Subsystem for Linux which cannot be used as `delegate_to` target as there's no SSH server running in it.

So my commit closes the gap from `checkmk_agent_delegate_download` to the Ansible server.

robin-checkmk:
But why don't you download the agent from the Checkmk server to your Ansible host and push it to the target from there? I think that is the bit I am missing. Because if you cannot directly download the agent from the Checkmk server to the target system, then why not use your Ansible host as the proxy? Why do you involve another host here?

mueller-ma:
I'm using a Windows client with WSL to run Ansible. I'd need to set `checkmk_agent_delegate_download ` to my Windows client, However Ansible will fail then. `delegate_to: checkmk_agent_delegate_download` requires an open SSH port. Running an openssh server on WSL is difficult due to several reasons (I don't want to get too off-topic here).
Therefore it's impossible for me to delegate the download to the Ansible server.

robin-checkmk:
Ah, now I get the full picture I think.
Are you aware of the `ansible_connection: local` host variable? I think if you set that for your WSL localhost, it should work, as it does not need SSH.
https://docs.ansible.com/ansible/latest/inventory_guide/connection_details.html#running-against-localhost

mueller-ma:
Thanks for the `localhost` hint. With the following host variables I don't need my patch anymore:
````
checkmk_agent_delegate_download: localhost
checkmk_agent_delegate_api_calls: localhost
````

As far as I can see the delegation only works for `localhost` or a hostname that resolves to the same host than `localhost`. Maybe the variable should be changed to `checkmk_agent_delegate_to_localhost_download: true`?

robin-checkmk:
I am glad to learn, that you could solve the issue!

From my perspective, the current state is good, but I will keep this PR in mind and if the issue arises again, I will invest time to understand how we can improve the situation.

Thank you for your cooperation and time!

Issue #452, opened 2023-10-19
[FEAT] Configure no_log property for agent registration


S0obi:
**Description**

Agent registration, executed using shell commands, may leak the automation password in OS syslog. I would like to propose to give the ability to users to set `no_log: true` in order to reduce the risk of leaking the password in an uncontrolled way.

Even if this change may not be perfect, we think that it will still reduce the likelihood of unexpected password leakage.

In order to reduce the user ability to troubleshoot and keep having a good experience with the ansible collection, we propose to keep the current behavior, meaning keeping the logging. This behavior should be documented in the role documentation.

PR #456, opened 2023-10-24
feat: add no log capability for agent and server roles

S0obi:
## Pull request type
- [X] Feature

Resolves: #452

## What is the new behavior?

Introduce `checkmk_agent_no_log` variable to better protect the automation password to be logged in system logs.

### DIFF
--- a/changelogs/fragments/no_log.yml
+++ b/changelogs/fragments/no_log.yml
@@ -1,0 +1,3 @@
+minor_changes:
+  - Agent role - Avoid logging passwords by default for extra security
+  - Server role - Avoid logging passwords by default for extra security
--- a/roles/agent/README.md
+++ b/roles/agent/README.md
@@ -153,6 +153,9 @@ Define an IP address which will be added to the host in Checkmk. This is optiona
 
 Define attributes with which the host will be added to Checkmk.
 
+    checkmk_agent_no_log: 'true'
+
+Whether to log sensitive information like passwords, Ansible output will be censored for enhanced security by default. Set to `false` for easier troubleshooting. Be careful when changing this value in production, passwords may be leaked in operating system logs.
 
 ## Tags
 Tasks are tagged with the following tags:
--- a/roles/agent/defaults/main.yml
+++ b/roles/agent/defaults/main.yml
@@ -31,7 +31,7 @@ checkmk_agent_folder: "{{ checkmk_folder_path | default('/') }}"
 checkmk_agent_force_foreign_changes: 'false'
 checkmk_agent_host_attributes:
   ipaddress: "{{ checkmk_agent_host_ip | default(omit) }}"
-
+checkmk_agent_no_log: 'true'
 
 # If you trust your local hostnames, you could also use the following
 # to use the local hostname instead of the inventory hostname:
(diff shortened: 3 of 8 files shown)


S0obi:

S0obi:
recheck

roles/agent/defaults/main.yml
```diff
@@ -31,7 +31,7 @@ checkmk_agent_folder: "{{ checkmk_folder_path | default('/') }}"
 checkmk_agent_force_foreign_changes: 'false'
 checkmk_agent_host_attributes:
   ipaddress: "{{ checkmk_agent_host_ip | default(omit) }}"
-
+checkmk_agent_no_log: 'false'
```
robin-checkmk, line 34: I think we should actually go for a default of `true`, as I would want a "security by default" approach here.
Ideally one does not need the output of those commands most of the time.

robin-checkmk (changes requested):
What do you think about adding your logic to the server role as well?
I like the style, how you implemented things, so if you are up to it, go right ahead.

roles/agent/README.md
```diff
@@ -153,6 +153,9 @@ Define an IP address which will be added to the host in Checkmk. This is optiona
 
 Define attributes with which the host will be added to Checkmk.
 
+    checkmk_agent_no_log: false
+
+The value of the `no_log` attribute. Set to `true` for enhanced password security.
```
robin-checkmk, line 158: We should also maybe clarify a little more, what happens here. You and I get it, but for the more inexperienced users, I would like to put this in simple terms.

roles/agent/defaults/main.yml
S0obi: Ok, I am glad if we follow this approach, I will do my best to improve the documentation to help users especially regarding troubleshooting.

roles/agent/README.md
```diff
@@ -153,6 +153,9 @@ Define an IP address which will be added to the host in Checkmk. This is optiona
 
 Define attributes with which the host will be added to Checkmk.
 
+    checkmk_agent_no_log: 'true'
+
+Whether to log sensitive information like passwords, Ansible output will be censored for enhanced security by default. Set to `false` for easier troubleshooting. Be carreful when changing this value in production, passwords may be leaked in operating system logs.
```
robin-checkmk, line 158: Just a typo left: "carreful" should be "careful". :)

roles/agent/defaults/main.yml
robin-checkmk: I was just thinking to align the configurability. So both roles behave the same. If you do not want to implement it here, I will gladly do so myself at a later time. No pressure. :)

S0obi: I introduced a new `checkmk_server_no_log` variable to align the behavior between agent and server roles. I also took the opportunity to fix the server role README that was messed up, most probably due to a failed copy & paste 👀

changelogs/fragments/no_log.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - Avoid logging passwords by default for extra security
```
robin-checkmk, line 2: Last nitpick: Please write the changelog like this:
- Agent role - Avoid logging passwords by default for extra security
- Server role - Avoid logging passwords by default for extra security
While it seems redundant, after compilation the context is important.

roles/agent/defaults/main.yml
robin-checkmk: Looking good, thanks!

changelogs/fragments/no_log.yml
robin-checkmk: And while you are at it: Fix the missing newline the tests complain about and then we are ready for merge. :)

S0obi: Fixed 🎉

S0obi: Fixed another new line missing and that failed the pipeline

robin-checkmk (approved):
Looks good! Pending the tests, this PR will be merged for the next release.

PR #459, opened 2023-10-25
Add support for omd config

robin-checkmk:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

OMD configuration of sites cannot be managed.

## What is the new behavior?

- This PR introduces management ability for OMD configuration.
- As it is necessary for a site to be stopped before OMD configuration can happen, handling has been added accordingly.
- Also, the molecule tests were improved with a focus on sites.
- And two new states were introduced: `enabled` and `disabled`.


robin-checkmk:
ToDo: Figure out a way to start and stop sites for configuration without surprising the user.

roles/server/tasks/sites.yml
```diff
@@ -36,6 +36,20 @@
...
+- name: "Enable Sites."  # noqa no-changed-when
+  become: true
+  ansible.builtin.shell: |
+    set -o pipefail
+    omd config {{ item.name }} AUTOSTART on
+  args:
+    executable: /bin/bash
+  no_log: true
+  loop: "{{ checkmk_server_sites }}"
+  when: (item.state == "enabled") or (item.state == "started")
```
lgetwan, line 48: Hm, why does a "started" site need AUTOSTART enabled?
Maybe it's just a matter of personal preferences, but I'd expect the following from the states:

- `started`: site is enabled and started. AUTOSTART could be enabled in a separate task.
- `stopped`: site is enabled and stopped. AUTOSTART could be enabled in a separate task, but makes not much sense.
- `enabled`: site is enabled but not actively started or stopped. AUTOSTART could be enabled in a separate task.
- `disabled`: site is disabled (with `omd disable`)
- `present`: same as `enabled`
- `absent`: site is destroyed

lgetwan (commented):
I like this feature! 
Probably, the possible values of "state" should be documented better, as my intuition seems to be different.

roles/server/tasks/sites.yml
robin-checkmk: My idea is to keep this logic aligned with the Checkmk Appliance.
Of course, one might want to only stop a site temporarily. But you never know, when a playbook might break and in the worst case a site that should be down comes up.

More specifically:
`started` and `stopped` were explained above. TL;DR: En-/disabling the site does not cause harm, and you know in which state the site will be. Otherwise, a server might be rebooted, and a site stay stopped, or starts, even if the configuration says otherwise.
`enabled`: Your idea is not a real state. You would not know, in which state a site is then.
`disabled`: Although there is an `omd` command, I intended this to be the opposite of `enabled`.


Ultimately: This is how we handle it in the role. If a user needed some elaborate playbook with specific in-between states, they would not use this role. But this is only my take on the matter. Maybe someone else can chime in and provide an opinion.

lgetwan: Keeping the states aligned with the Appliance sounds reasonable!

robin-checkmk:
> I like this feature! Probably, the possible values of "state" should be documented better, as my intuition seems to be different.

I added some clarification in 388a970dbdffce34771df084d84713636f906e2d.
However there is still work needed, but that would lead to a breaking change, so it is tracked in #464.

PR #462, opened 2023-10-27
Release 3.4.0

robin-checkmk:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [x] Build related changes

## What is the new behavior?

- #440 
- #456 
- #459
- #465
- #427


robin-checkmk:
@msekania and @Max-checkmk: Can you please take a look at a3dbf379b1536ec9601b9f4c10ff218d473d0aba and let me know if the change seems sound to you? I wanted to improve the tests, but not make big deal out of it. So you just taking a peek if looks sound and giving me a thumbs up should be fine. Thanks! :pray:

msekania:
@robin-checkmk,

Yes, these checks are better.
However, here one takes granted that the error messages from the Ansible remain as they are now. 
Anyway, one only needs to remember that these can be a possible sources of errors and make appropriate adjustments when necessary.

robin-checkmk:
@msekania I do quite like the fact, that the tests will fail, if anything changes. Because then we can understand what exactly is going on and adapt if necessary. Otherwise, we might miss something.

Thanks for your feedback!

lgetwan (approved):
Looks good to me. Go for it!

PR #475, opened 2023-11-13
Added errorhandling to lookup API

Max-checkmk:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- If there's an error when calling the url, the error is passed uncontrolled to the user

## What is the new behavior?

- Errors when calling the endpoint will be catched in the lookup_api and returned to the lookup-modules for output

## Other information

I'm sure there's a nicer way to do this. So any improvements appreciated.
Main problem is that ansible doesn't allow using the ansible.errors library in the api folder, just inside the modules.


plugins/module_utils/lookup_api.py
```diff
@@ -35,9 +45,22 @@ def get(self, endpoint="", parameters=None):
...
-        raw_response = open_url(
-            url, headers=self.headers, validate_certs=self.validate_certs
-        )
-        response = to_text(raw_response.read())
+
+        try:
+            raw_response = open_url(
+                url, headers=self.headers, validate_certs=self.validate_certs
+            )
+            response = to_text(raw_response.read())
```
lgetwan, line 53: Actually, it would be nicer to directly do a return here:
`return to_text(raw_response.read())`

plugins/module_utils/lookup_api.py
```diff
@@ -35,9 +45,22 @@ def get(self, endpoint="", parameters=None):
...
+        except HTTPError as e:
+            if e.code in HTTP_ERROR_CODES:
+                response = json.dumps(
```
lgetwan, line 56: Do a return here, too. 
And in the following cases, as well.
Then, we get completely rid of the variable "response".

lgetwan (changes requested):
I appreciate a lot to have a logging mechanism! :-D
Only minor change requests below...

plugins/lookup/folder.py
```diff
@@ -92,6 +93,11 @@ def run(self, terms, variables, **kwargs):
             response = json.loads(
                 api.get("/objects/folder_config/" + term.replace("/", "~"))
             )
+            if "code" in response:
+                raise AnsibleError(
+                    "Received error for %s - %s: %s"
+                    % (response["url"], response["code"], response["msg"])
```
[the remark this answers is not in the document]
Max-checkmk: You're right, even when i always use all 3 values in the lookup_api.
Changed with last commit!

plugins/module_utils/lookup_api.py
Max-checkmk: Accepted and committed.

lgetwan (approved):
Looks perfect, now!

PR #478, opened 2023-11-22
Adding bakingstatus lookup

Max-checkmk:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

There is no lookup for the bakingstatus

## What is the new behavior?

- Added a lookup-module for the bakingstatus

### DIFF
--- a/.github/labels-issues.yml
+++ b/.github/labels-issues.yml
@@ -67,3 +67,6 @@ lookup:host:
 
 lookup:hosts:
   - 'Component Name: lookup_hosts'
+
+lookup:bakingstatus:
+  - 'Component Name: lookup_bakingstatus'
--- a/.github/labels-prs.yml
+++ b/.github/labels-prs.yml
@@ -64,3 +64,6 @@ lookup:host:
 
 lookup:hosts:
   - 'plugins/modules/lookup/hosts.py'
+
+lookup:bakingstatus:
+  - 'plugins/modules/lookup/bakingstatus.py'
--- a/.github/workflows/ans-int-test-lkp-bakingstatus.yaml
+++ b/.github/workflows/ans-int-test-lkp-bakingstatus.yaml
@@ -1,0 +1,70 @@
+# README:
+# - When changing the module name, it needs to be changed in 'env:MODULE_NAME' and in 'on:pull_requests:path'!
+#
+# Resources:
+# - Template for this file: https://github.com/ansible-collections/collection_template/blob/main/.github/workflows/ansible-test.yml
+# - About Ansible integration tests: https://docs.ansible.com/ansible/latest/dev_guide/testing_integration.html
+
+env:
+  NAMESPACE: checkmk
+  COLLECTION_NAME: general
+  MODULE_NAME: lookup_bakingstatus
+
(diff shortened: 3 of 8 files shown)


robin-checkmk:
Honestly, I do not like the name of the module. Would it be terrible to name it `lookup_bakery` to be aligned to the actual module?

Max-checkmk:
> Honestly, I do not like the name of the module. Would it be terrible to name it `lookup_bakery` to be aligned to the actual module?

I just took the name of the endpoint, so i don't have any preferences on that.
At least at the moment there is no other endpoint which has something to do with the bakery.

robin-checkmk:
Then I would ask you to rename it, if you see no good arguments against it.

plugins/lookup/bakery.py
```diff
@@ -0,0 +1,99 @@
...
+
+DOCUMENTATION = """
+    name: bakery
+    author: Max Sickora (@max-checkmk)
+    version_added: "3.5.0"
+    short_description: Get the bakery status of a Checkmk server
+    description:
+      - Returns the bakery status of a Checkmk server as a string, e.g. 'running'
+    options:
+      _terms:
```
lgetwan, line 16: I'm not sure if and how to solve that, but some users requested to have "server_url" and "site" separately.
But what to use the "terms" for, then? The module doesn't expect anything like an item or parameters (besides the connection parameters that are equal in all our lookup modules).
@robin-checkmk: Can we just ignore the _terms variable completely, then? Technically, yes, but it feels odd, though.
Same problem in the version looku module...

lgetwan (dismissed):
Looks fine and works well!
Please see my small comment on the "_terms"

plugins/lookup/bakery.py
robin-checkmk: It certainly makes sense, to have `server_url` and `site` configurable independently, e.g. to enable users to use their high level variables here. As `_term` can be different things altogether, there is no generic approach.
Maybe we can add `server_url` and `site` to the module(s) and compile `_term` internally? If that works technically, it should solve our issue.

tests/integration/targets/lookup_bakery/tasks/test.yml
```diff
@@ -0,0 +1,13 @@
...
+    msg: "Bakery status is {{ bakery }}"
+  vars:
+    bakery: "{{ lookup('checkmk.general.bakery',
+                    server_url + '/' + outer_item.site,
+                    validate_certs=False,
+                    automation_user=automation_user,
+                    automation_secret=automation_secret)
+              }}"
+  delegate_to: localhost
+  register: looked_up_bakery
```
robin-checkmk, line 13: Can we add some assertion logic here, that actually tests, if we get an expected response?
Currently, we would only see hard failures, but there might be situations, where just the output changes unexpectedly, without the module to fail.

For reference, you could look at `verify.yml` in the molecule tests of a role.

plugins/lookup/bakery.py
lgetwan: Ok, let's use "site" and "server_url". 
Then, we don't need the "_terms", at all and will just ignore it of the user provides it.
I will change the version lookup module to act similarly.

PR #479, opened 2023-11-22
Added redirect option for activation endpoint

Max-checkmk:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- Redirect option is always set to false in the activation endpoint.
- In the agent role we have a fixed wait time of 30sec after activating changes

## What is the new behavior?

- It's now possible to set redirect to true
- In the agent role the activation handler has redirect set to true, so there's no need to wait anymore

## Other information

Even with redirect set to true, there is no need to handle the redirect (like with service discovery endpoint).
This works without further changes.

### DIFF
--- a/plugins/modules/activation.py
+++ b/plugins/modules/activation.py
@@ -24,6 +24,10 @@ description:
 extends_documentation_fragment: [checkmk.general.common]
 
 options:
+    redirect:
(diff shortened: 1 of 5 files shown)


robin-checkmk:
As discussed outside of GitHub, we will clarify some things internally before moving this forward.
Technically it looks fine, we just want to check, that there are no side effects.
Once that has been cleared, we can merge this for release `4.0.0`.

Issue #500, opened 2023-12-15
[BUG] Password module complains about required arguments, which are not actually required


robin-checkmk:
**Describe the bug**

The password module behaves as if parameters are required, which are not required by the Checkmk REST API.

**Component Name**

Component Name: password

**Ansible Version**

```console
$ ansible --version
ansible [core 2.16.1]
```

**Checkmk Version and Edition**

```console
2.2.0p16 CCE
```

**Collection Version**

```console
$ ansible-galaxy collection list
Collection                    Version
----------------------------- -------
ansible.posix                 1.4.0  
checkmk.general               4.0.1  
community.general             5.5.0  
```

**To Reproduce**
1. Define a password, omitting e.g., the `documentation_url`.
2. Run the playbook
3. See the error under 'Actual behavior'.

**Expected behavior**

The password entry is created without the omitted parameters.

**Actual behavior**

```
failed: [host -> localhost] (item={'name': 'vsphere_ro', 'title': 'vSphere Read-Only User', 'password': 'topsecret1', 'comment': '2023-12-15T15:42:55Z Set by Ansible', 'owner': 'admin'}) => {"ansible_loop_var": "item", "changed": false, "item": {"comment": "2023-12-15T15:42:55Z Set by Ansible", "name": "vsphere_ro", "owner": "admin", "password": "topsecret1", "title": "vSphere Read-Only User"}, "msg": "400 - Bad Request: Parameter or validation failure. Details: b'{\"title\": \"Bad Request\", \"status\": 400, \"detail\": \"These fields have problems: documentation_url, shared\", \"fields\": {\"documentation_url\": [\"Field may not be null.\"], \"shared\": [\"Field may not be null.\"]}}'"}

```

**Minimum reproduction example**

```yaml
    - name: "Create Passwords."
      tags: [rules, config]
      checkmk.general.password:
        server_url: "{{ server_url }}"
        site: "{{ site }}"
        automation_user: "{{ automation_user }}"
        automation_secret: "{{ automation_secret }}"
      - name: "vsphere_ro"
        title: "vSphere Read-Only User"
        password: "topsecret1"
        comment: "{{ ansible_date_time.iso8601 }} Set by Ansible"
        owner: "admin"
        state: "present"
      delegate_to: localhost
      run_once: true  # noqa run-once[task]
      loop: "{{ checkmk_passwords }}"
```

**Additional context**

PR #506, opened 2023-12-18
Fix for #500

Max-checkmk:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Non given values will be None and transferred to the API.

Issue Number: #500 

## What is the new behavior?

- List comprehension is now added also to the CreateAPI

### DIFF
--- a/plugins/modules/password.py
+++ b/plugins/modules/password.py
@@ -218,27 +218,16 @@ HTTP_CODES_UPDATE = {
 
 class PasswordsCreateAPI(CheckmkAPI):
     def post(self):
-        if self.params.get("customer", None) is not None:
-            data = {
-                "ident": self.params.get("name", ""),
-                "title": self.params.get("title", ""),
-                "customer": self.params.get("customer", ""),
-                "comment": self.params.get("comment", ""),
-                "documentation_url": self.params.get("documentation_url", ""),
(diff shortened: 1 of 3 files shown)


plugins/modules/password.py
```diff
@@ -240,6 +240,9 @@ def post(self):
                 "shared": self.params.get("shared", ""),
             }
 
+        # Remove all keys without value, as otherwise they would be None.
+        data = {key: val for key, val in data.items() if val}
+
```
lgetwan, line 245: You could simplify the whole post() function a lot. I think, this should give you the same result:

```
    def post(self):

        keys = [
                "name",
                "title",
                "customer",
                "comment",
                "documentation_url",
                "password",
                "owner",
                "shared",
        ]

        def _map(key):
            if key == "name":
                return "ident"
            return key

        data = { _map(key): params[key] for key in keys if key in params}

        return self._fetch(
            code_mapping=HTTP_CODES_CREATE,
            endpoint="/domain-types/password/collections/all",
            data=data,
            method="POST",
    )
```

Max-checkmk: After a short test i choose the old variant just without the extra if-statement.

robin-checkmk (approved):

Issue #481, opened 2023-11-29
[BUG] tag_group module does not properly handle error


robin-checkmk:
**Describe the bug**

The tag group module says `"msg": "Tag group created."`, although there is clearly an error.  
Full error message:
```console
fatal: [my_host.local]: FAILED! => {"changed": true, "http_code": 400, "info": {"body": "{\"title\": \"Bad Request\", \"status\": 400, \"detail\": \"These fields have problems: tags\", \"fields\": {\"tags\": [\"Tags IDs must be unique. You've used the following at least twice: 'cloud_provider_azure'\"]}}", "connection": "close", "content-length": "200", "content-security-policy": "default-src 'self' 'unsafe-inline' 'unsafe-eval' ssh: rdp:; img-src 'self' data: https://*.tile.openstreetmap.org/ ; connect-src 'self' https://crash.checkmk.com/ https://license.checkmk.com/api/verify; frame-ancestors 'self' ; base-uri 'self'; form-action 'self' javascript: 'unsafe-inline'; object-src 'self'; worker-src 'self' blob:", "content-type": "application/problem+json", "date": "Wed, 29 Nov 2023 12:46:38 GMT", "msg": "HTTP Error 400: BAD REQUEST", "permissions-policy": "accelerometer=(), autoplay=(), camera=(), encrypted-media=(), fullscreen=(), geolocation=(), gyroscope=(), magnetometer=(), microphone=(), midi=(), payment=(), picture-in-picture=(), usb=()", "referrer-policy": "origin-when-cross-origin", "server": "Apache", "status": 400, "url": "http://192.168.56.112//my_host/check_mk/api/1.0/domain-types/host_tag_group/collections/all", "x-content-type-options": "nosniff", "x-frame-options": "sameorigin", "x-permitted-cross-domain-policies": "none", "x-xss-protection": "1; mode=block"}, "msg": "Tag group created."}
```

**Component Name**

Component Name: tag_group

**Ansible Version**

```console
$ ansible --version
ansible [core 2.15.6]
```

**Checkmk Version and Edition**

```console
2.2.0p14 (CCE)
```

**Collection Version**

```console
$ ansible-galaxy collection list
checkmk.general               4.0.0
```

**To Reproduce**
1. Create a playbook that creates tag groups.
2. Use the same `id` twice.
3. See the error.

**Expected behavior**

The error message gets printed properly.

**Actual behavior**

The message indicates the tag_group was created, although it was not.

[...]

PR #445, opened 2023-09-29
Feature/tag group add help

meni2029:
## Pull request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

Currently the "help" is not supported by the tag_group module. It is supported by the API.

## What is the new behavior?

- Adding the "help" attribute in the tag_group module.

## Other information

I just replicated the handling of "topic" in the code. I checked that "help" is supported by the API.

PR #512, opened 2023-12-22
Rewrite of Tag_group module

Max-checkmk:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- tag_group-module doesn't use CheckmkAPI
- help parameter can't be set
- repair flag can't be set
- Issue Number: #481 

## What is the new behavior?

- Module was rewritten with CheckmkAPI
- Parameter help can be used now
- It's now possible to define the flag "repair" when updating or deleting a host_tag
- Issue with the return msg when configuring duplicated tags is solved witht his PR.

## Other information

PR #445 wanted to introduce the help parameter, but with the existent code it was complicated to to accomplish it.


Max-checkmk:
Works already, but still planning to enhance the tests. So please wait with the merge! ;)

tests/integration/targets/tag_group/vars/main.yml
```diff
@@ -1,14 +1,8 @@
...
-    site: "stable_cme"
   - version: "2.2.0p17"
     edition: "cre"
     site: "stable_raw"
-  - version: "2.2.0p17"
-    edition: "cee"
-    site: "stable_ent"
   - version: "2.1.0p37"
     edition: "cre"
     site: "old_raw"
```
robin-checkmk, line 8: We might have discussed this, but why are you removing the test sites for commercial editions here?

robin-checkmk (commented):
I just looked at the code I can review and did not yet test anything.

plugins/modules/tag_group.py
```diff
@@ -60,20 +71,21 @@
...
-        title: ESXi
-      - id: vCenter
-        title: vCenter
-      - id: HyperV
-        title: HyperV
-      - id: KVM
-        title: KVM
+    name: Datacenter
+    title: Datacenter
+    topic: Custom_Tags
```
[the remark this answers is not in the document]
Max-checkmk: Changed with last commit

lgetwan (commented):
There is no check if the desired state is already present, so we get a "changed" after each run. The original module did check that and return an "ok", if nothing had to be changed.
I'll check if that can be rebuilt without too much effort.

lgetwan:
@Max-checkmk, I improved the idempotency. If you're happy with the result, we can merge, I'd say.

tests/integration/targets/tag_group/vars/main.yml
Max-checkmk: As we discussed i've added the cee again.

plugins/modules/tag_group.py
```diff
@@ -38,19 +40,28 @@
...
+        type: bool
     state:
         description: The desired state
         default: "present"
         choices: ["present", "absent"]
         type: str
 
 author:
-    - Stefan Mühling (@muehlings)
+    - Max Sickora (@Max-checkmk)
```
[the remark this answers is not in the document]
Max-checkmk: Although the code changed nearly completely, you're right and i've added him again.

Max-checkmk:
@muehlings Hi Stefan, as this plugin was based on your code, i hope it is ok for you, that we did a rewrite on this?
Maybe you want to have a look on this, if it still satisfies your demands. ;)

muehlings (approved):
Looks good.

muehlings (approved):

lgetwan (approved):
Looks perfect, now! :-)

plugins/modules/tag_group.py
```diff
@@ -38,43 +40,65 @@
...
-      - id: HyperV
-        title: HyperV
-      - id: KVM
-        title: KVM
+    name: Datacenter
+    title: Datacenter
+    topic: Tags
+    help: "something useful"
+    tags:
+      - ident: No_Datacenter
```
robin-checkmk, line 81: I just realized now, that you used `ident` here. While that is in line with the REST API wording, I'd rather use `id`. Objections?

lgetwan: In the create & update endpoints, the API uses "ident", but when using a show endpoint, "id" is used.
Thus, we could also use "id" here, as a conversion has to be done, anyway...

muehlings: I tried to use identical names to reduce confusion. Often the API returns other keys then it expects when writing to it. And the format of the object isn't identical either. I voted against id because of:
grafik

Then id_ would be the better name for that field. I tried to write some converters but it was impossible to guess the right names and object. I think I've seen "name" as replacement for "id" in version 2.2 of the API recently.

I'm currently parsing the swagger.yaml to fetch all domains, methods, objects and responses to automatically create modules from that definition (or at least understand the API better). Originally I tried to use the Ansible content-builder, but it took to long to understand which parts have to be provided from me to make it work.

robin-checkmk: While I see your reasoning @muehlings, I still think `id` is the better choice here. And my reason is simple: This is a user-facing name. The average Ansible user does not care about the shenanigans going on inside the REST API. In fact, they do not want to think about that at all, but they just want to use the module without too much research or reading documentation.
So even as a non-native speaker, what makes more sense to you (ignoring the background knowledge that you have): `id` or `ident`?
On top of intuitivity, what is the field in question really? I know we struggle with that in Checkmk in several places, but effectively it is an ID. Not a name, not someting else.

muehlings: You can choose whatever fits best.

My idea was: Choose/use the name the API returns to put it back into it without renaming confusion. But conversions have to happen anyway, because the returned object differs from the expected one when writing it. If there is an internal guideline (e.g. "Avoid 'id' as variable/name and use 'ID' [or 'name' or 'ident'] instead), you would be one who could know it. I can't streamline that as external participant.

GET ("items" should be described in #/components/schemas/HostTag1, I don't have that definition in my 2.2.0p17 openapi):
grafik

POST/PUT (expect 'ident'):
grafik

I think the internal way is "domaintype_id" now:
grafik

I left out " 'id " in grep, but there are still some "id" fields.

lgetwan:
I renamed the "ident" to "id".

Now, I ask myself why the integration test is not triggered, here.
@robin-checkmk, do you have an idea?

lgetwan:
Ok. Integration test is running again. It didn't run before due to a merge conflict in the test target.
From my point of view, the PR is read to be merged. Any vetos?

robin-checkmk (approved):
Everything looks fine by me.

PR #508, opened 2023-12-20
The rule module has now a parameter rule_id

lgetwan:
The rule module has now a parameter rule_id that can be used to identify an existing rule.

## Pull request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

- It was not possible to delete a rule with a known rule_id
- It was not possible to modify existing rules, at all

## What is the new behavior?

- When a rule_id is given, the user can delete or modify the rule

PR #517, opened 2023-12-27
Fix idempotency for relative rule location

meni2029:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Rule module: the location options _rule_id_ and _folder_ are mutually exclusive. The problem comes when creating a rule with a relative location to an existing rule_id (before/after) and an expected folder other than Main. The expected folder for the rule is not known by the module, therefore it is not capable to recognise an already existing rule. It always compares with rules in default folder Main. Idempotency is not achieved for rules outside of Main folder and duplicates are created at each Ansible run.

## What is the new behavior?

- new function get_rule_by_id
- modified function get_existing_rule : when location option rule_id is set, the expected folder is retrieved from the neighbour rule
- now the function get_existing_rule  is able to acurately check if the rule already exists in the expected folder


robin-checkmk:
Thank you for this contribution @meni2029! :pray: 
Our code-savvy people are on well-deserved holiday, but as soon as they are back, they will take a look at your changes.
Have you seen #508? :eyes: I am not sure, if these two PRs somehow correlate, but @lgetwan should be able to distinguish that, once he is back.

lgetwan:
Thanks for detecting this incorrect behavior, @meni2029!
If I understood your code correctly, it improves the situation already, but considering this case

- The desired rule is already in the correct folder but not located before/after its intended neighbor rule

the task will reply a "changed" instead of an "ok". Fixing this would bring us one step closer to idempotency (which is really challenging for this module... :exploding_head:).

lgetwan (approved):
You're right! The module is still not perfect, but your PR brings an improvement. Thanks for your contribution!

PR #525, opened 2024-01-16
Don't become in checkmk_agent_auto_activate

mueller-ma:
Root isn't required for this step.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently this handler fails on my machine as `localhost` doesn't allow password-less sudo.

## What is the new behavior?

Now sudo isn't used anymore which fixes this issue.


robin-checkmk:
On first sight this makes sense. I will queue it for the next release and double-check before, if I see any risk for regressions. Thanks for noticing!

robin-checkmk (approved):
Seems reasonable to me.

Issue #515, opened 2023-12-27
[BUG] Investigate behavior of the parent attribute when passed to the `host` or `folder` module


robin-checkmk:
**Describe the bug**

When passing the parent `parent` attribute to the `host` or `folder` module one has to use a very particular syntax:
```YAML
parents: ["my_parent_host"]
```
If we do not pass it as a list, but a string for example, the modules do odd things, like interpreting every character as a dedicated host.

**Component Name**

Component Name: host, folder

**Ansible Version**

```console
$ ansible --version

```

**Checkmk Version and Edition**

```console
2.2.0p17 (CRE)
```

**Collection Version**

```console
$ ansible-galaxy collection list
`4.1.0`
```

**To Reproduce**
Steps to reproduce the behavior: See above.

**Expected behavior**

The modules are more robust towards the data type.
Or they would refuse to use "false" values and print an error.

**Actual behavior**

See above.

**Minimum reproduction example**

```yaml

```

**Additional context**

msekania:
@robin-checkmk, 

This is quite normal behavior for other Ansible modules too, which wait for a list but receive a string.

If I am not mistaken, the package module handles both inputs adequately.
One can implement the similar semantic here too, but I'm not sure if it's worth it.

robin-checkmk:
Thanks for your input @msekania!

I do agree, that technically this is fine, but is quite inconvenient for a user.
If it is not too much effort, I think implementing this for convenience would be great.
But we do not want to invest too much time probably.

msekania:
Hi @robin-checkmk,

several points:
- why it's marked like bug, when it is not?
- It also states `parents`, plural, so one could guess that list is what is expected.
- I have an idea (have not tested though) how one can include the desired conversion in the corresponding modules when attribute contain `parents`, but what's about similar attributes then?

robin-checkmk:
Hi @msekania,

1. While I see your point, I think to a user it sure feels like a bug.
2. That comes down to the fact that the REST API uses the plural, although in many use cases, there is only a single parent. But that is not really the point of this issue.
3. I see your point and I do realize that it is just one among a lot of possible attributes. I need to look for another example, but I have none in mind right now. However, the parent attribute is validated in the UI and you can only add existing hosts as parents. This is currently broken for the REST API, but once that is fixed, we will get errors directly instead of having non-existing parents configured.

[...]

PR #539, opened 2024-01-27
Extend the functionality of the folder module

msekania:
Extends the functionality of the folder module.
See also #533 .

## Pull request type

Please check the type of change your PR introduces:

- [X] Bugfix
- [X] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [X] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

Issue Number: #533 

## What is the new behavior?

- One can supply a) a list o strings or b) dict as a romote_attributes parameter. 
- a) reproduces the functionality of the REST API, and was already implemented
- b) removes the specified attribute if the given folder had such an attribute.

PR #536, opened 2024-01-25
Rewrite folder module and migrate to new collection API.

lgetwan:
## Pull request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

The folder module was still a bit oldschool spaghetti code.

## What is the new behavior?

The folder module now uses the new API

PR #545, opened 2024-01-31
Feature parse parents attribute as list

msekania:
Improve Fix/Solve #515 

## Pull request type

if `parents` attribute is provided in `attributes` or `update_attributes` of host or folder module parses files as list

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

#515 
if one provides `parents: test` it would be parsed as ['t', 'e', 's', 't'] (because `parents` awaits list as a parameter)
now it will be parsed as ['test']

## What is the new behavior?

- if `parents` gets a string (or some other type) as a parameter ansible utils will take care that it will be interpreted as a list


msekania:
@lgetwan, @robin-checkmk ,

this PR can be merged right away, before #536

robin-checkmk:
Thanks for the fix @msekania! I will leave it up to @lgetwan to review the actual code, but I am very happy, then we got a handle on this!

lgetwan (commented):
This change also seems to be picked into your other PR #539. I reviewed it there, but it looks slightly different, there: You're using a list FOLDER_PARENTS_PARSE, there, which even looks better. 
I'm not sure how to solve this without messing with git.

msekania:
> This change also seems to be picked into your other PR #539. I reviewed it there, but it looks slightly different, there: You're using a list FOLDER_PARENTS_PARSE, there, which even looks better. I'm not sure how to solve this without messing with git.

@lgetwan, 

I have modified it accordingly

lgetwan:
This PR should be merged after #539. When doing this, we might get a minor merge conflict in the folder.py, but never mind.

lgetwan (approved):
Thanks for implementing this, Michael!

PR #562, opened 2024-02-17
Rewrite host module using new api, add checkmode and extend functionality

msekania:
Rewrite host module using new api
Add checkmode and add extend functionality similar to folder module

## Pull request type

- [ ] Bugfix
- [X] Feature
- [X] Code style update (formatting, renaming)
- [ ] Build related changes

Ansible checkmode is not supported

## What is the new behavior?

- Adds support of ansible check mode

- The host module now uses the new API

- In check mod, the returned change status is set to True if the changes would have taken place in normal mode.

- One can supply a) a list o strings or b) dict as a romote_attributes parameter.
- a) reproduces the functionality of the REST API, and was already implemented
- b) removes the specified attribute if the given folder had such an attribute.

### DIFF
--- a/plugins/modules/host.py
+++ b/plugins/modules/host.py
@@ -237,21 +237,16 @@ class HostAPI(CheckmkAPI):
         # Get the current host from the API and set some parameters
         self._get_current()
 
-        if self.current.get("folder"):
-            if self.params.get("folder"):
-                tmp_folder = self._normalize_folder(self.params.get("folder"))
-                if tmp_folder != self.current.get("folder"):
-                    self.desired["folder"] = tmp_folder
-            else:
-                if self.current.get("folder") != "/":
-                    self.desired["folder"] = "/"
+        if self.params.get("folder"):
+            tmp_folder = self._normalize_folder(self.params.get("folder"))
         else:
-            if self.params.get("folder"):
-                self.desired["folder"] = self._normalize_folder(
-                    self.params.get("folder")
-                )
-            else:
-                self.desired["folder"] = "/"
+            tmp_folder = self._normalize_folder("/")
+
+        if current_host.state == "present":
+            if tmp_folder != self.current.get("folder"):
+                self.desired["folder"] = tmp_folder
+        else:
+            self.desired["folder"] = tmp_folder
 
         self._changed_items = self._detect_changes()
 
--- a/plugins/modules/host.py
+++ b/plugins/modules/host.py
@@ -242,9 +242,7 @@ class HostAPI(CheckmkAPI):
         else:
             tmp_folder = "/"
 
-        if not self.current.get("folder"):
-            self.desired["folder"] = tmp_folder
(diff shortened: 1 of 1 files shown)


msekania:
@lgetwan @robin-checkmk,

I have also included version check for interface similar to folder module concerning deprecated simultaneous use of `attributes`, `update_attributes`, `remove_attributes` after certain version.
However, I am not sure whether host API (being similar to folder API) will also impose those restriction.

There is also a tricky part concerning the moving host to the different folder, namely, affective attributes might be different when host is moved to the different folder. This issues was also on the former versions of the host module. I have just realized it recently, chasing the silly bug in new implementation.

I have not a better proposal for the moment therefor the same logic is implemented in the PR. 

May be it will be better (less error prone) to strictly split move to folder and  any update (removal) of the attributes. What do you think?
In the similar way one might also include rename host feature from the REST API.

msekania:
@robin-checkmk, @lgetwan 

One more question concerning the check mode.

After rewriting using new API, check mode returns `changed: false`  with list of changes to be carried on not in check mode.
Following ansible behavior,, we should return `changed: true` if a normal run without check mode would carry out changes.
The check mode should reflect this explicitly.

So I would suggest we modify `_check_output` function as follows
```
    def _check_output(self, mode):
        return RESULT(
            http_code=0,
            msg="Running in check mode. Would have done an %s" % mode,
            content="",
            etag="",
            failed=False,
            changed=True,
        )
```

affects at least folder, host, and current proposal of rule modules

lgetwan:
> There is also a tricky part concerning the **moving host** to the different folder, namely, effective attributes (`?effective_attributes=true` feature in `_get_current()`) might be changed when host is moved to a different folder. This issues was also in the former versions of the host module. I have just notices it recently, when I was chasing the silly bug in the new implementation. I have not have a better proposal for the moment, therefor the same logic is implemented in the current PR.
> 
> However, it might be better (less error prone) to strictly split **move to folder** and any **update (removal) of the attributes**, like in the REST API. What do you think? In the similar way one might also include rename host feature from the REST API. Possibly with `state: move` option? Which takes care about **move to folder** and **rename host**.

I agree. Moving a host to another folder or renaming it should be separated from updating its attributes.
Users should be aware that moving/renaming a host usually changes its inherited attributes and/or matching rules.
Our module should not change this behavior, so maybe we should prevent the user from changing folder _and_ attributes at the same time.

msekania:
> Our module should not change this behavior, so maybe we should prevent the user from changing folder _and_ attributes at the same time.

Yes, good idea indeed, and will avoid unexpected behavior.
On the other hand might be a breaking change for those who are already using this combined behavior. One should communicate these change appropriately.

I can carry out the corresponding changes in code, if one agrees to go this path.

robin-checkmk:
As @lgetwan is already on board, I give my approval as well.

msekania:
DONE!

> As @lgetwan is already on board, I give my approval as well.

Issue #518, opened 2023-12-28
[FEED] Enable managing cluster objects with the `host` module


robin-checkmk:
**Provide your feedback**

It should be possible to configure cluster objects in Checkmk, either through the `host` module, or through a dedicated module.

**Component Name**

Component Name: host

**Ansible Version**

```console
$ ansible --version

```

**Collection Version**

```console
$ ansible-galaxy collection list

```
**Environment**

**Screenshots**

**Additional context**

There was a PR once, but the contributor did not follow through: #246

msekania:
@robin-checkmk,

in coming weeks I'll have time for it.
I will need this part of the functionality for myself in the future, so I can see if I can implement it right away.

robin-checkmk:
Implemented in #562.

msekania:
Hi @robin-checkmk,

Not yet, it will come with the next version not with the 4.4.0

> Implemented in #562.

robin-checkmk:
I was lazy and closed the issue "early" in a sense. I felt like that is okay, as I reported it. :grimacing: 
But the work is done, I just need to find time to finalize the release.

PR #563, opened 2024-02-18
New Cluster Module

msekania:
Create/Modify/Delete cluster host of given nodes.
Stand alone module to reduce complexity of the host module.

See #518 

## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Currently there is no any module that manages creation of the host as an aggregate cluster of the given list of nodes.
See #518 

## What is the new behavior?

- Creates cluster host of given nodes, with the given name and folder
- Modifies list of cluster nodes
- Manages attributes of the given cluster host
- Deletes specified cluster host

### DIFF
--- a/.github/labels-issues.yml
+++ b/.github/labels-issues.yml
@@ -17,6 +17,9 @@ module:activation:
 module:bakery:
   - 'Component Name: bakery'
 
+module:cluster:
+  - 'Component Name: cluster'
+
 module:contact_group:
   - 'Component Name: contact_group'
 
--- a/.github/labels-prs.yml
+++ b/.github/labels-prs.yml
@@ -24,6 +24,11 @@ module:bakery:
       - changed-files:
           - any-glob-to-any-file: 'plugins/modules/bakery.py'
 
+module:cluster:
+  - any:
+      - changed-files:
+          - any-glob-to-any-file: 'plugins/modules/cluster.py'
+
 module:contact_group:
   - any:
       - changed-files:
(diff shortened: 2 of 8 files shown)


msekania:
Can be also included in host module introducing the new state
`state: cluster`

Not quite sure whether increasing complexity of the host module is feasible.
On the other hand rest of the management is done by host module.

robin-checkmk:
My gut feeling says, I would rather see it in the host module, but I cannot judge the complexity code-wise. Also, I am not yet sure, if my gut feeling is a good indicator, as I prefer fact based decisions.

> On the other hand rest of the management is done by host module.

@msekania does this mean, that one needs both this new module and the `host` module to manage cluster hosts, or does it mean, that you duplicated the code?

msekania:
@robin-checkmk,


@robin-checkmk,

> @msekania does this mean, that one needs both this new module and the `host` module to manage cluster hosts, or does it mean, that you duplicated the code?

Yes, I have duplicated the code in the last two **working** iterations.

Concerning including it to host module, after the last test I got an idea how one can do it.
It will require careful testing though.
I'll try to prepare clean extension to host module and make PR against #562

robin-checkmk:
@msekania as #562 is progressing, where do we stand? Should we close this PR, or should we keep it open, in case incorporating the cluster management in the `host` module does not work?

msekania:
@robin-checkmk, I'll close this PR. I have another version of cluster module which will build on #562.
So as soon as it is accepted I can carry out few adjustments in the module which I have already prepared in the forked version and make a PR with cluster extensions for host module.

PR #572, opened 2024-02-22
Migrate Rule module to new API

lgetwan:
I migrated the rule module to the new API and made it a bit nicer.
Will update this description soon. Just wanted to save my work before the weekend. ;-)

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Rule module is a bit ugly and only ~23% idempotent.

## What is the new behavior?

Rule module
- is nicer,
- uses the new API,
- is ~87% idempotent.
- The rules lookup module accepts filtering for folders, now.

## Other information

100% idempotency will be reached, once the value_raw is handled differently by Checkmk and its REST API.

### DIFF
--- a/plugins/modules/rule.py
+++ b/plugins/modules/rule.py
@@ -232,9 +232,6 @@ EXAMPLES = r"""
             "value": "yes"
           }
         ],
-        "host_name": {},
-        "host_tags": [],
-        "service_labels": []
       }
       properties: {
         "comment": "Ansible managed",
@@ -258,19 +255,20 @@ EXAMPLES = r"""
     ruleset: "checkgroup_parameters:memory_percentage_used"
     rule:
       rule_id: "{{ item.id }}"
-    loop: "{{
-             lookup('checkmk.general.rules',
-               ruleset=outer_item.ruleset,
-               comment_regex='Ansible managed',
-               server_url=server_url,
-               site=site,
-               automation_user=automation_user,
-               automation_secret=automation_secret,
-               validate_certs=False
-               )
-           }}"
-    loop_control:
-      label: "{{ item.id }}"
+    state: "absent"
+  loop: "{{
+           lookup('checkmk.general.rules',
+             ruleset='checkgroup_parameters:memory_percentage_used',
+             comment_regex='Ansible managed',
+             server_url=server_url,
+             site=site,
+             automation_user=automation_user,
+             automation_secret=automation_secret,
+             validate_certs=False
+             )
+         }}"
+  loop_control:
+    label: "{{ item.id }}"
 """
 
 RETURN = r"""


lgetwan:
@msekania: You already spent a lot of time diving into the rule module and its flaws. Can you have a look at my completely rewritten rule module and test it, if you have some time?

msekania:
@lgetwan, yes I'll do 

> @msekania: You already spent a lot of time diving into the rule module and its flaws. Can you have a look at my completely rewritten rule module and test it, if you have some time?

tests/integration/targets/rule/tasks/test.yml
```diff
@@ -11,32 +11,54 @@
   delegate_to: localhost
   run_once: true  # noqa run-once[task]
   loop: "{{ checkmk_var_rules }}"
+  register: created_rules
 
-- name: "{{ outer_item.version }} - {{ outer_item.edition | upper }} - Activate."
-  activation:
+- name: "{{ outer_item.version }} - {{ outer_item.edition | upper }} - Debug registered variable."
+  debug:
+    var: created_rules
```
robin-checkmk, line 18: Was this only for testing, or should it stay to better understand subsequent issues?

tests/integration/targets/rule/vars/main.yml
```diff
@@ -116,35 +116,3 @@ checkmk_var_rules:
...
-            'group_time': 900,
-            'mode': 2,
-            'service_filters':(
-              'combined', {'service_whitelist': ['^E2E.*']}
-            )
-          },
-          'severity_new_host_label': 0,
-          'severity_unmonitored': 0,
-          'severity_vanished': 0
-        }"
```
robin-checkmk, line 150: Why do we remove this rule set? Because of the breaking change in Checkmk 2.3?
Should we replace it or migrate it?

msekania:
@lgetwan,

I've gone through the code and have more than a few suggestions for improvement and cleanup. I'll do the code review comments right after.

As for the rest, overwriting, it's good overall, but I don't like that the functionality has been reduced compared to the existing version. Namely, the check whether a rule already exists no longer takes place if rule_id is not specified. I understand that the existing version was not perfect in this regard, but it was still useful and all my use cases relied on this functionality. Currently there is also no way to compensate the "removed" functionality with the lookup rule plugin.

I am currently working to bring that functionality back, if you do not mind.

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
 
-    return None
+        # Set defaults
+        for what, defaults in DESIRED_DEFAULTS.items():
+            for key, default in defaults.items():
+                if not desired["rule"].get(what):
+                    desired["rule"][what] = {}
 
+                tmp_prop = desired["rule"].get(what, {})
+                if not desired["rule"].get(what, {}).get(key):
```
msekania, line 477: should read
```
                if not desired["rule"].get(what).get(key):
```
`get(what, {})`    `{}` is not required see lines 473-474

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
+                msg="ERROR: The %s value_raw has invalid format" % state
+            )
 
-    url = base_url + api_endpoint
+        return safe_value_raw
 
-    response, info = fetch_url(
-        module, url, module.jsonify(params), headers=headers, method="POST"
-    )
+    def _detect_changes(self):
```
msekania, line 499: remove extra clean line

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
-        return (None, changed)
+    def _build_default_endpoint(self):
+        return "%s/%s" % (
+            RuleEndpoints.default,
+            self.rule_id,
+        )
 
-    headers["If-Match"] = get_rule_etag(module, base_url, headers, rule_id)
+    def _get_current(self):
+        current = {}
```
msekania, line 534: add
```
        state = "absent"
```

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
-    }
+        result = self._fetch(
+            code_mapping=RuleHTTPCodes.get,
+            endpoint=self._build_default_endpoint(),
+            method="GET",
+        )
+
+        if result.http_code == 200:
+            current["rule"] = {}
+            self.state = "present"
```
msekania, line 543: change with
```
            state = "present"
```

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
-    url = base_url + api_endpoint
+            content = json.loads(result.content)
+            extensions = content["extensions"]
 
-    info = fetch_url(
-        module, url, module.jsonify(params), headers=headers, method="PUT"
-    )[1]
+            for key, value in extensions.items():
+                if key in CURRENT_RULE_KEYS:
+                    current["rule"][key] = value
```
msekania, line 551: more compact version
```
            current["rule"] = {key: value for key, value in extensions.items() if key in CURRENT_RULE_KEYS}
```

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
 
-    if info["status"] not in [200, 204]:
-        exit_failed(
-            module,
-            "Error calling API. HTTP code %d. Details: %s, "
-            % (info["status"], info["body"]),
+        else:
+            self.state = "absent"
+        return current
```
msekania, line 555: change this three lines with
```
        return (current, state)
```

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
+            return True
 
-def delete_rule(module, base_url, headers, ruleset, rule):
-    changed = True
-    rule_id = get_existing_rule(module, base_url, headers, ruleset, rule)
+        if self.is_new_rule:
+            location = self.desired.get("rule").get("location")
+            if location and not (
+                location.get("folder", "") == "/"
+                and location.get("position", "") == "bottom"
```
msekania, line 578: if default values then may be the right ones
```
                location.get("folder", "/") == "/"
                and location.get("position", "bottom") == "bottom"
```

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
+        if not self._moving_needed():
+            return
 
-def delete_rule_by_id(module, base_url, headers, rule_id):
-    api_endpoint = "/objects/rule/"
+        location = self.desired.get("rule", {}).get("location")
+        data = {"position": POSITION_MAPPING[location.get("position")]}
 
-    url = "%s%s%s" % (base_url, api_endpoint, rule_id)
+        pos = location.get("position", "")
```
msekania, line 591: If default values, may be the right ones?
```
        pos = location.get("position", "bottom")
```

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
+        if pos in ["top", "bottom"]:
+            data["folder"] = location.get("folder", "/")
+        elif pos in ["before", "after"]:
+            data["rule_id"] = location.get("neighbour")
```
msekania, line 596: just for the guide
```
        # else:
        #     # cannot happen
```

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
+            content=list(results.values())[-1].content,
+            etag=list(results.values())[-1].etag,
+            failed=any(r.failed for r in list(results.values())),
+            changed=any(r.changed for r in list(results.values())),
+        )
 
-def get_rule_etag(module, base_url, headers, rule_id):
-    api_endpoint = "/objects/rule/" + rule_id
+    def create(self):
+        data = self.desired.get("rule", {}).copy()
```
msekania, line 623: should read
```
        data = self.desired.get("rule").copy()
```
rule is required parameter

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
+        location = data.pop("location", {})
+        data["ruleset"] = self.desired.get("ruleset")
+        data["folder"] = location.get("folder", "/")
 
-    url = base_url + api_endpoint
+        if not data.get("value_raw"):
+            self.module.fail_json(msg="ERROR: The parameter value_raw is mandatory.")
```
msekania, line 629:
```
            self.module.fail_json(msg="ERROR: The parameter value_raw is mandatory when 'state is present'.")
```
IMHO, better message

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
 
-    api_keywords = {
-        "top": "top_of_folder",
-        "bottom": "bottom_of_folder",
-        "before": "before_specific_rule",
-        "after": "after_specific_rule",
-    }
+        move_result = self._move_if_needed()
+        if move_result:
+            m = self._merge_results({"created": create_result, "moved": move_result})
```
msekania, line 649: not used not needed

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
-        params["rule_id"] = location["rule_id"]
-    else:
-        params["folder"] = location["folder"]
+    def edit(self):
+        data = self.desired.get("rule", {}).copy()
+        data.pop("location")
+        self.headers["if-Match"] = self.etag
+
+        if not data.get("value_raw"):
+            self.module.fail_json(msg="ERROR: The parameter value_raw is mandatory.")
```
msekania, line 660:
```
            self.module.fail_json(msg="ERROR: The parameter value_raw is mandatory when 'state is present'.")
```
IMHO, better message

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
-    url = base_url + api_endpoint
+        if edit_result.failed:
+            return edit_result
 
-    info = fetch_url(
-        module, url, module.jsonify(params), headers=headers, method="POST"
-    )[1]
+        move_result = self._move_if_needed()
+        if move_result:
+            return self._merge_results({"editd": edit_result, "moved": move_result})
```
msekania, line 681: spelling
`edited` instead of `editd`

plugins/modules/rule.py
```diff
@@ -535,15 +724,15 @@ def run_module():
...
-                        ("position", "before", ("rule_id",)),
-                        ("position", "after", ("rule_id",)),
-                    ],
-                    mutually_exclusive=[("folder", "rule_id")],
+                    # required_if=[
+                    #    ("position", "top", ("folder",)),
+                    #    ("position", "bottom", ("folder",)),
+                    #    ("position", "before", ("neighbour",)),
+                    #    ("position", "after", ("neighbour",)),
+                    # ],
```
msekania, line 734: uncomment!

plugins/modules/rule.py
```diff
@@ -553,76 +742,51 @@ def run_module():
...
-    }
+    # Create an API object that contains the current and desired state
+    current_rule = RuleAPI(module)
 
-    base_url = "%s/%s/check_mk/api/1.0" % (
-        module.params.get("server_url", ""),
-        module.params.get("site", ""),
+    result = RESULT(
+        http_code=0,
+        msg="Invalid parameters provided.",
```
msekania, line 750:
```
        msg="",
```
fits better

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
-            "Error calling API. HTTP code %d. Details: %s, "
-            % (info["status"], info["body"]),
+        else:
+            self.state = "absent"
+        return current
+
+    def _check_output(self, mode):
+        return RESULT(
+            http_code=0,
+            msg="Running in check mode. Would have done an %s" % mode,
```
msekania, line 560:
```
            msg="Running in check mode. Would have %s" % mode,
```
IMHO, better message

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
+            content="",
+            etag="",
+            failed=False,
+            changed=False,
```
msekania, line 564:
```
            changed=True,
```
user should see in check-mode if changes would be carried out in the normal mode. follows Ansible beheviour

msekania (changes requested):
I cannot make any changes to this branch otherwise I could apply the changes directly or via merge request

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
+        self.params = self.module.params
+        self.rule_id = self.params.get("rule", {}).get("rule_id")
+        self.is_new_rule = self.rule_id is None
 
-    return json.loads(response.read().decode("utf-8"))
+        self.desired = self._clean_desired(self.params)
 
+        self._changed_items = []
+        self.current = None
+        self.etag = ""
```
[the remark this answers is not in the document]
msekania:
```
        # when neighbour is specified, verify that it exists otherwise give warning
        neighbour_id = self.params.get("rule").get("location").get("neighbour", "")
        if neighbour_id != "":
            (neighbour, state) = self._get_rule_by_id(neighbour_id)

            if state == "absent":
                self.module.warn("Specified neighbour: '%s' does not exist" % neighbour_id)
            else:
                self.desired["location"]["folder"] = neighbour.get("location").get("folder")
```

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
 
+        if self._raw_value_eval("current", current) != self._raw_value_eval(
+            "desired", desired
+        ):
+            changes.append("raw_value")
 
-def modify_rule(module, base_url, headers, ruleset, rule):
-    changed = True
-    rule_id = rule.get("rule_id")
+        desired_location = desired.get("rule", {}).get("location")
```
msekania, line 515: I think, this should read
```
        desired_location = desired.get("location")
```

lgetwan:
Hi @msekania 

Thanks a lot for your comprehensive review! 
I need some time to reply to the single findings though, as my calendar is currently packed with customer calls.

> As for the rest, overwriting, it's good overall, but I don't like that the functionality has been reduced compared to the existing version. Namely, the check whether a rule already exists no longer takes place if rule_id is not specified. I understand that the existing version was not perfect in this regard, but it was still useful and all my use cases relied on this functionality. Currently there is also no way to compensate the "removed" functionality with the lookup rule plugin.
> 
> I am currently working to bring that functionality back, if you do not mind.

I know that my changes are breaking existing stuff. I decided to do that, because the current module is kind of unreliable and unpredictable in its behavior, and so I wanted to move the module and its use cases to the direction of how the API is intended to be used. I will discuss with my colleagues if we might bring back the previous behavior at least as a side-feature.
Or probably (as you also wrote) it would help extending the rules lookup module, so it includes conditions and raw_value to search for. Would that be an option for you? If not, How should the rules lookup module behave, so you can keep your current use cases?

Best regards
Lars

msekania:
Hi @lgetwan,

General problem which I have with a new version (without modification) is that that one creates all the time the same rule, because there is no any mechanisms that allows to check and stop recreation of the same rule again and again.
Currently one requires to provide rule_id in order to compare, edit, or move the rule. This will stay as it is.

I would appreciate if one would have a mechanism to try to deduce the rule_id if user is not providing it from the rest of parameters. Positioning in folder issues one can solve by introducing extra position value "any" in addition to "top", "bottom", "before", and "after", which will ignore precise position check within folder in `is_equal` check,

I have finished implementation of this features, and it's more or less like what I would like to have. I also found some extra bugs in current version.

In which form I  should provide the changes?

Best regards,
Michael

robin-checkmk:
> In which form I should provide the changes?

@msekania may I suggest a branch based on this branch? That should allow you to open a PR onto this one.

tests/integration/targets/rule/tasks/test.yml
lgetwan: Yes, that was only for testing. Removed it.

plugins/modules/rule.py
```diff
@@ -155,15 +149,15 @@
...
         "description": "Allow even higher memory usage",
         "disabled": false,
-        "documentation_url": "https://github.com/Checkmk/ansible-collection-checkmk.general/blob/main/plugins/modules/rules.py"
+        "documentation_url": "https://github.com/Checkmk/ansible-collection-checkmk.general/blob/main/plugins/modules/rule.py"
       }
       value_raw: "{'levels': (85.0, 99.0)}"
       location:
         position: "after"
-        rule_id: "{{ response.id }}"
+        rule_id: "{{ response.content.id }}"
```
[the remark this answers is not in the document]
lgetwan: Thanks for fixing this in your PR, alredy.

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
 
-    params = {
-        "ruleset_name": ruleset,
-    }
+        self.ruleset = self.params.get("ruleset")
 
-    url = "%s%s?%s" % (base_url, api_endpoint, urlencode(params))
+        self.rule_list = self._get_ruleset(self.module.params.get("ruleset"))
+        self.folder_rule_list = [
+            k for k, v in self.rule_list.items() if v == self.folder
```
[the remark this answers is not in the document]
lgetwan: Thanks for fixing this in your PR, alredy.

plugins/modules/rule.py
```diff
@@ -235,276 +236,464 @@
...
 
-    return json.loads(response.read().decode("utf-8")).get("value")
+        if result.http_code == 200:
+            content = json.loads(result.content)
+            return {
+                r.get("id"): r.get("extensions", {}).get("folder")
+                for r in content.get("value")
+            }
 
+        return []
```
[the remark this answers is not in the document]
lgetwan: Thanks for fixing this in your PR, alredy.

plugins/modules/rule.py
lgetwan: Thanks for fixing this in your PR, alredy.

plugins/modules/rule.py
lgetwan: Thanks for fixing this in your PR, alredy.

plugins/modules/rule.py
lgetwan: Thanks for fixing this in your PR, alredy.

plugins/modules/rule.py
lgetwan: Thanks for fixing this in your PR, alredy.

plugins/modules/rule.py
lgetwan: Thanks for fixing this in your PR, alredy.

plugins/modules/rule.py
lgetwan: I don't think so. "location" is a parameter inside of the rule, not a top level one.

tests/integration/targets/rule/vars/main.yml
lgetwan: I think, it's not a good idea to use rulesets in our tests that have different data structures from one (tested) Checkmk version to another.

lgetwan:
Again, it's complicated when modifying an existing rule (by providing an ID) to decide if conditions should be set or not. Querying the values of an existing rule and re-applying them leads to problems.

msekania:
> Again, it's complicated when modifying an existing rule (by providing an ID) to decide if conditions should be set or not. Querying the values of an existing rule and re-applying them leads to problems.

@lgetwan can you provide some examples

robin-checkmk:
My initial tests show, that both for Checkmk 2.2.0p25 and 2.3.0b6 the idempotent is still completely broken without specifying the `rule_id`.

Representative example rule definition:
```yaml
  - ruleset: "checkgroup_parameters:cpu_utilization_os"
    rule:
      location:
        folder: "/"
        position: "bottom"
      properties: {
        "comment": "Set by Ansible",
        "disabled": false
      }
      value_raw: "{'core_util_time': (100.0, 300, 900), 'core_util_time_total': (100.0, 300, 900)}"
      conditions:
        host_labels: []
        service_labels: []
        host_tags: []
```
Accompanying task:
```yaml
- name: "Create Rules."
  checkmk.general.rule:
    site: "{{ site }}"
    ruleset: "{{ item.ruleset }}"
    rule: "{{ item.rule }}"
    state: "present"
  loop: "{{ rules_global }}"
  delegate_to: localhost
```

I am still investigating and will update this post, if I find something else.

msekania:
Thanks @robin-checkmk, for the example,

I'll have some tests.

Concerning `rule_id`, problem is that it's not existing when rule is created, therefor it's useless in bare create task.
If `rule_id` will remain the only solution for the moment, then working example would be:
- search if the rule with desired properties already exists, and if yes get `rule_id`,
- incorporate `rule_id` if found in the task which creates rule, otherwise omit this filed entirely.

msekania:
@robin-checkmk, 

In current rue module setting any value in description lest say
```yml
  - ruleset: "checkgroup_parameters:cpu_utilization_os"
    rule:
      location:
        folder: "/"
        position: "bottom"
      properties: {
        "description": "test"
        "comment": "Set by Ansible",
        "disabled": false
      }
      value_raw: "{'core_util_time': (100.0, 300, 900), 'core_util_time_total': (100.0, 300, 900)}"
      conditions:
        host_labels: []
        service_labels: []
        host_tags: []
```

solves idempotency issues. I'll check new version next, but this point might be essential.

msekania:
@robin-checkmk, @lgetwan,

I have a fix for the problem, namely issue is `description` field in `properties` in `DESIRED_DEFAULTS` line 353 https://github.com/Checkmk/ansible-collection-checkmk.general/blob/39f143e456ccd2e40431830deb72c6b4843ca915/plugins/modules/rule.py#L353, may be also https://github.com/Checkmk/ansible-collection-checkmk.general/blob/39f143e456ccd2e40431830deb72c6b4843ca915/plugins/modules/rule.py#L365,
`description` should not be the part of `properties`  in `DESIRED_DEFAULTS` (at least for `pre_230`)

robin-checkmk:
> In current rue module setting any value in description let's say [...] solves the idempotency peoblem.
I can confirm this for 2.2. Setting `"description": "test"`, makes virtually all my example rules idempotent again. In 2.3 this seems not to help though.

> I have a fix for the problem, namely issue is description field in properties in DESIRED_DEFAULTS line 353 [...], may be also [365], description should not be the part of properties in DESIRED_DEFAULTS (at least for pre_230)
I pushed a change commenting out these lines. But I want @lgetwan to get a chance to verify, if this is a sensible fix before merging it.

lgetwan:
I will check that tomorrow.

lgetwan:
The last commit solves the problem for 2.2.0, but I still don't get why it doesn't work for 2.3.0...

lgetwan:
@robin-checkmk & @msekania,
That was a hard piece of work considering that the diff is quite small, but now I'm pretty sure, that the idempotency in 2.3.0 is as "good" as in 2.2.0. Can you run your tests, again, as well, please?

msekania:
@lgetwan,

there are some more improvements see #602 , I could not test 2.3.0 yet, but similar fixes might be due for 2.3.0 as well.

I also have a question concerning your last changes.
What if I intentionally provide empty list just to clean up already set one?
I think this changes should also go similar to what I have done for comment and description in #602

msekania:
@lgetwan, 

I have an Idea how to incorporate logic of your changes in what I wrote above. I'll need hour or two at most.
Of course you can drop them if you do not like them, no problem.

msekania:
@lgetwan 

> @lgetwan,
> 
> I have an Idea how to incorporate logic of your changes in what I wrote above. I'll need hour or two at most. Of course you can drop them if you do not like them, no problem.

DONE

lgetwan:
Ok, now idempotency looks quite good with 2.2.0 and 2.3.0, at least for my own tests and the integration tests.
Thanks again for #602, @msekania!
Can you run your tests against the newest version of this PR?

robin-checkmk:
I can confirm idempotency as well for 2.2 and 2.3.
The only hiccup is, that `"comment": "{{ ansible_date_time.iso8601 }} Set by Ansible"` will not work, as you always get a different comment. But I think that is fine. In that case one needs to use the `rule_id`.

The only challenges that remain from my point of view are:
1. Verify that the module documentation is up-to-date with the recent changes.
2. Decide whether this is a breaking change or not and how to release it.

PR #585, opened 2024-03-28
Feature/add cluster managment to host module

msekania:
Adds management of cluster hosts

## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Issue Number: #518

## What is the new behavior?

What is the new behavior?

host module now can mange cluster hosts

- Create/Modify/Delete Cluster Host
- Modify attributes of the cluster host similar to normal host
- add, remove, change nodes of the cluster
- rename host or cluster hosts (**currently still not functioning because of strange REST API behavior**)


robin-checkmk:
I love the elegance of this approach! No boolean or a different key for clusters. Just defining nodes makes a host a cluster host. Love it! Pending a code review I love to add this to the next possible release. :muscle:

plugins/modules/host.py
```diff
@@ -73,6 +73,32 @@
         description: Allow extended functionality instead of the expected REST API behavior.
         type: bool
         default: true
+    # new_name:
+    #     description: The new name of the managed host.
+    #     required: false
+    #     type: str
```
robin-checkmk, line 79: What's with this commented out section?

msekania: I have not managed to get "Rename Host Part" to work so far.
It could be a problem with the REST API, or maybe I misunderstood how to use the dedicated API request.

Help is always welcome!

plugins/modules/host.py
```diff
@@ -256,6 +327,39 @@ def __init__(self, module):
...
+                    if el not in self.desired["nodes"]
+                ]
+            elif self.current.get("cluster_nodes"):
+                self.desired["nodes"] = self.current.get("cluster_nodes") + [
+                    el
+                    for el in self.params.get("add_nodes")
+                    if el not in self.current.get("cluster_nodes")
+                ]
+            else:
+                self.desired["nodes"] = self.desired.get("add_nodes")
```
lgetwan, line 347: Combining two lists and removing duplicate elements can be done with
`unique_list = list(set(my_list_01 + my_list_02))`

plugins/modules/host.py
```diff
@@ -316,12 +420,81 @@ def _build_move_endpoint(self):
...
+                or len(
+                    [
+                        el
+                        for el in desired_parameters.get("nodes")
+                        if el not in current_nodes
+                    ]
+                )
+                > 0
+            ):
+                changes.append("nodes")
```
lgetwan, line 480: There's also a simple way to do an XOR for two lists:
```
if set(current_nodes).symmetric_difference(desired_parameters.get("nodes", [])):
    changes.append("nodes")
```

lgetwan (commented):
Thanks for adding this feature and the corresponding tests!
I just added two comments where you could simplify the code a bit.

plugins/modules/host.py
msekania: OMG, sure there are ... set-s.

thanks again, I'll carry out the corresponding changes

plugins/modules/host.py
msekania: Yes of course, thanks a lot Lars, I always had the feeling that I was overdoing it here!

lgetwan (approved):
Looks perfect now.

Issue #571, opened 2024-02-22
[Feed] folder lookup plugin does not return the title of the folder


msekania:
The current implementation of the folder lookup plugin does not return the `title` of the looked up folder, it only returns  `extensions`

**Describe the bug**

It is rather a feature request!

**Component Name**

Component Name: lookup folder

**Ansible Version**

**Checkmk Version and Edition**

**Collection Version**

lgetwan:
@msekania,
 I fixed that in PR #589. Please check if that fits your needs. Please mind that this is a breaking change, because the structure of the returned data has changed.

msekania:
@lgetwan,

yes, it's fine!
thanks!

robin-checkmk:
Fixed with `5.0.0`.

PR #589, opened 2024-04-10
Fix folder lookup plugin title

lgetwan:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #571 
The folder lookup plugin didn't return the folder title, but only the extensions.

## What is the new behavior?

- Return the complete folder information, not only the extensions.

## Other information

People using this lookup plugin have to adjust the usage of the variables in the reply.
### Before:
```
- name: "Get attributes of folder."
  ansible.builtin.debug:
    msg: "Criticality of /tests is {{ folder.attributes.tag_criticality }}"
  vars:
    folder: "{{ lookup('checkmk.general.folder',
                    '/tests',
                    server_url=server_url,
                    site=site,
                    validate_certs=False,
                    automation_user=automation_user,
                    automation_secret=automation_secret)
              }}"
```
### After:
```
- name: "Get attributes of folder."
  ansible.builtin.debug:
    msg: "Criticality of /tests is {{ folder.extensions.attributes.tag_criticality }}"
  vars:
    folder: "{{ lookup('checkmk.general.folder',
                    checkmk_folder.path,
                    '/tests',
                    site=site,
                    validate_certs=False,
                    automation_user=automation_user,
                    automation_secret=automation_secret)
              }}"
```

### DIFF
--- a/changelogs/fragments/lookup_folder.yml
+++ b/changelogs/fragments/lookup_folder.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - Folder lookup module - Return the complete folder information, not only the extensions.
--- a/plugins/lookup/folder.py
+++ b/plugins/lookup/folder.py
@@ -165,6 +165,6 @@ class LookupModule(LookupBase):
                         response.get("msg", ""),
                     )
                 )
-            ret.append(response.get("extensions"))
+            ret.append(response)
 
         return ret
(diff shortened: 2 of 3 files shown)


robin-checkmk:
@lgetwan so from what I read, this is a breaking change, as users using the module have to adapt their configuration, right? So if they do not act, things break. If that is correct, I would line this up for `5.0.0` and depending on how urgent it is, push to get that release out sooner. But if it is not that critical, I would focus on `4.4.0` for now and release `5.0.0` when we finalized Checkmk 2.3.0 support.

robin-checkmk:
@lgetwan I think the examples of the module are out of sync with these changes. I looked at the changelog fragment and wanted to update it to be `breaking_changes` and then realized, that I do not know where to point them to for the necessary changes. Can you double-check and potentially create a pull request based on `devel` to fix the docs/examples? Thanks!

lgetwan:
Good point. See #607.

PR #596, opened 2024-04-26
Fix host being always initially created in main folder.

lgetwan:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Hosts are always created in main folder if previous state was `absent`.

## What is the new behavior?

- Hosts are created in the folder that is provided in the parameters

### DIFF
--- a/changelogs/fragments/fix_host_initial_folder.yml
+++ b/changelogs/fragments/fix_host_initial_folder.yml
@@ -1,2 +1,2 @@
 bugfixes:
- - Rule module - Fix empty rule conditions.
+ - Host module - Fix hosts being always created in the main directory.
--- a/changelogs/fragments/fix_host_initial_folder.yml
+++ b/changelogs/fragments/fix_host_initial_folder.yml
@@ -1,0 +1,2 @@
+bugfixes:
+ - Rule module - Fix empty rule conditions.
--- a/plugins/modules/host.py
+++ b/plugins/modules/host.py
@@ -320,12 +320,7 @@ class HostAPI(CheckmkAPI):
         # Get the current host from the API and set some parameters
         self._get_current()
 
-        if self.state == "present":
-            if (
-                self.params.get("folder")
-                and self.current["folder"] != self.params["folder"]
-            ):
-                self.desired["folder"] = self.params["folder"]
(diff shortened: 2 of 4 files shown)


changelogs/fragments/fix_host_initial_folder.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+ - Rule module - Fix empty rule conditions.
```
robin-checkmk, line 2: This looks like the wrong changelog fragment to me. :)

robin-checkmk:
Fix confirmed in Checkmk 2.2.0p25. Now testing 2.3.0b6.

msekania:
@lgetwan, 

thanks for quick bug fix!

may be it would better also to include test of this behavior in test suite. So that the future changes do not miss this point.

robin-checkmk:
> may be it would better also to include test of this behavior in test suite. So that the future changes do not miss this point.

I like that, but have no time to implement it. If you are up to the task, please go ahead.
I will merge this PR and release a bug fix release soonish though.

robin-checkmk (approved):

msekania:
@robin-checkmk,

I have quickly extended the tests, see #599.

robin-checkmk:
Apparently the new tests in #599 do not work with Checkmk 2.0.0. I have not started digging yet, but as we will be dropping support for 2.0.0 from the collection soon, I would not invest too much time. @msekania if you can quickly see what is wrong, that would be great. Otherwise, I might have to remove testing for 2.0.0 here, so we can get the bug fix release on the road.

So by the end of tomorrow, we either have a solution, or will remove 2.0.0 from the testing scope.

msekania:
@robin-checkmk, 

Yes I see the errors.

Unfortunately, I do not have a quick fix to the tests, except skipping it for version 2.0.0.

I am not quite sure were the problem is. From the automatic test output one can see that the hosts are created in the correct folders. I am not sure though, that the host lookup plugin extracts folder field correctly for version 2.0.0.

robin-checkmk:
I disabled the 2.0.0 tests for the `host` module for now.

Issue #593, opened 2024-04-19
[BUG] Error in the Agent role, Firewall seems to be only installed if checkmk_agent_server is an ip address


hasselk:
**Describe the bug**

It seems the Firewall Rules (Redhat and Debian) are only created if the checkmk_agent_server is an Ip address and skipped when not, which seems a bit Odd since you need an FQDN if you want to use https and check the cert.

**Component Name**

Component Name: roles/agent/tasks/ **Debian.yml and Redhat.yml**

**Ansible Version**

```console
ansible [core 2.14.5]
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.11/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.11.8 (main, Feb 19 2024, 22:58:08) [GCC 12.2.1 20220924] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True

```

**Checkmk Version and Edition**

```console
2.2.0p24 (CRE)
```

**Collection Version**

```console
Collection                    Version
----------------------------- -------
amazon.aws                    5.4.0
ansible.netcommon             4.1.0
ansible.posix                 1.5.2
ansible.utils                 2.9.0
ansible.windows               1.13.0
arista.eos                    6.0.1
awx.awx                       21.14.0
azure.azcollection            1.15.0
check_point.mgmt              4.0.0
chocolatey.chocolatey         1.4.0
cisco.aci                     2.6.0
cisco.asa                     4.0.0
cisco.dnac                    6.7.1
cisco.intersight              1.0.27
cisco.ios                     4.5.0
cisco.iosxr                   4.1.0
cisco.ise                     2.5.12
cisco.meraki                  2.15.1
cisco.mso                     2.4.0
cisco.nso                     1.0.3
cisco.nxos                    4.3.0
cisco.ucs                     1.8.0
cloud.common                  2.1.3
cloudscale_ch.cloud           2.2.4
community.aws                 5.4.0
community.azure               2.0.0
community.ciscosmb            1.0.5
community.crypto              2.12.0
community.digitalocean        1.23.0
community.dns                 2.5.3
community.docker              3.4.3
community.fortios             1.0.0
community.general             6.6.0
community.google              1.0.0
community.grafana             1.5.4
[...]

PR #597, opened 2024-04-26
Feature/agent adv fw mgmt

robin-checkmk:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: #593 

## What is the new behavior?

- A new variable is introduced: `checkmk_agent_server_ips`. I can be set to a list of IPs, which will be allowed in the firewall.
- This will **not** modify the agent's own whitelisting mechanisms!
- The `checkmk_agent_server` will automatically be added to the above list, but only if it is an IP address.

## Other information

This collection has an implicit dependency for the Python module `netaddr`. This is not new, but apparently never was an issue. This PR adds a hint to the README about this.

### DIFF
--- a/roles/agent/tasks/Debian.yml
+++ b/roles/agent/tasks/Debian.yml
@@ -47,17 +47,18 @@
 - name: "{{ ansible_os_family }} Derivatives: Configure Firewall for Agent."
   when: checkmk_agent_configure_firewall | bool and "ufw.service" in ansible_facts.services
   block:
-    - name: "{{ ansible_os_family }} Derivatives: Check if checkmk_agent_server is an IP address."
+    - name: "{{ ansible_os_family }} Derivatives: Add Checkmk Server to Firewall Whitelist if it is an IP address."
+      when: checkmk_agent_server | ansible.utils.ipaddr()
       ansible.builtin.set_fact:
-        checkmk_agent_server_ip: "{{ checkmk_agent_server }}"
-      when: checkmk_agent_server_ip is not defined and checkmk_agent_server | ansible.utils.ipaddr()
+        checkmk_agent_server_ips: "{{ checkmk_agent_server_ips + [checkmk_agent_server] }}"
 
     - name: "{{ ansible_os_family }} Derivatives: Allow Checkmk services access to the agent."
+      when: checkmk_agent_server_ips is defined
       community.general.ufw:
         rule: allow
         proto: tcp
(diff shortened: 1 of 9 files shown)


roles/agent/tasks/Debian.yml
```diff
@@ -47,17 +47,18 @@
...
     - name: "{{ ansible_os_family }} Derivatives: Allow Checkmk services access to the agent."
+      when: checkmk_agent_server_ips is defined
       community.general.ufw:
         rule: allow
         proto: tcp
-        src: "{{ checkmk_agent_server_ip }}"
+        src: "{{ item }}"
         port: '6556'
         comment: Allow Checkmk
-      when: checkmk_agent_server_ip is defined
```
mueller-ma, line 62: This should be marked as breaking change as it removes the variable `checkmk_agent_server_ip`. I set this variable to an IP address, so that the ufw rule can be added.

roles/agent/tasks/Debian.yml
```diff
@@ -47,17 +47,18 @@
 - name: "{{ ansible_os_family }} Derivatives: Configure Firewall for Agent."
   when: checkmk_agent_configure_firewall | bool and "ufw.service" in ansible_facts.services
   block:
-    - name: "{{ ansible_os_family }} Derivatives: Check if checkmk_agent_server is an IP address."
+    - name: "{{ ansible_os_family }} Derivatives: Add Checkmk Server to Firewall Whitelist if it is an IP address."
+      when: checkmk_agent_server | ansible.utils.ipaddr()
```
mueller-ma, line 51: Maybe try to resolve `checkmk_agent_server` here when it's a DNS name?

robin-checkmk: Funny you should say that. I spent a lot of time implementing that, but it introduces complexity and more dependencies. Eventually I dropped the attempt and kept it simple.

roles/agent/tasks/Debian.yml
robin-checkmk: Thanks for noticing! See, the thing is, that the variable was never intended to be used by users, so technically I would not consider it a breaking change. But we might not have made that clear enough. If you can give me a brief feedback, if the changes in here work for you, I will merge it prior to the `5.0.0` release and mark it as breaking.

mueller-ma: I tested this on a Debian 12 client and it works 👍

PR #599, opened 2024-04-30
Extend host module tests

msekania:
Add tests to host module

## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

## What is the new behavior?

- test for default folder resolution for hosts
- test for default folder resolution for cluster hosts

### DIFF
--- a/tests/integration/targets/host/tasks/test.yml
+++ b/tests/integration/targets/host/tasks/test.yml
@@ -40,6 +40,22 @@
   delegate_to: localhost
   run_once: true  # noqa run-once[task]
 
+- name: "{{ outer_item.version }} - {{ outer_item.edition | upper }} - Verify host folder."
+  ansible.builtin.assert:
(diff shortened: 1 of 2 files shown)


robin-checkmk:
@msekania maybe I am misunderstanding, but you basically added a host without an explicit folder, right?
Shouldn't we also add a host with a specific folder and verify, that both of these things yield the correct results?

msekania:
@robin-checkmk, 

> maybe I am misunderstanding, but you basically added a host without an explicit folder, right?

yes, right!

> Shouldn't we also add a host with a specific folder and verify, that both of these things yield the correct results?

There are already hosts with specific folder, but no check that they are created in the requested ones.
I'll add that check too.

msekania:
@robin-checkmk, 

I have included the folder test for hosts with given folder parameter, via host lookup plugin

robin-checkmk:
Awesome, thanks man!

PR #607, opened 2024-05-29
Folder lookup module - Update and improve the examples.

lgetwan:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

The examples in the module aren't up-to-date.

## What is the new behavior?

- The examples respect the new dictionary structure
- There's an example on how to query multiple folders, which results in a list of dictionaries.

### DIFF
--- a/changelogs/fragments/lookup_folder_examples.yml
+++ b/changelogs/fragments/lookup_folder_examples.yml
@@ -1,0 +1,2 @@
+breaking_changes:
+  - Folder lookup module - Update and improve the examples.
--- a/changelogs/fragments/lookup_folder_examples.yml
+++ b/changelogs/fragments/lookup_folder_examples.yml
@@ -1,2 +1,0 @@
-breaking_changes:
-  - Folder lookup module - Update and improve the examples.
--- a/plugins/lookup/folder.py
+++ b/plugins/lookup/folder.py
@@ -88,24 +88,23 @@ DOCUMENTATION = """
 """
 
 EXAMPLES = """
-- name: Get the attributes of folder /tests
+- name: Get the attributes of folders /tests and /snmp
   ansible.builtin.debug:
-    msg: "Attributes of folder /network: {{ attributes }}"
-  vars:
-    attributes: "{{
-                    lookup('checkmk.general.folder',
-                        '~tests',
-                        server_url=my_server_url,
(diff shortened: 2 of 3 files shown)


changelogs/fragments/lookup_folder_examples.yml
```diff
@@ -0,0 +1,2 @@
+breaking_changes:
+  - Folder lookup module - Update and improve the examples.
```
robin-checkmk, line 2: The original changelog fragment said `Return the complete folder information, not only the extensions.`. I think this here is misleading, as it says, what you did in this PR, but we actually want to explain here, what is the breaking change and how to update their configuration. Maybe you have a good idea, otherwise I can propose something. We should also update the existing fragment, rather than adding a new one.

plugins/lookup/folder.py
```diff
@@ -88,24 +88,23 @@
...
+                  automation_user=myuser,
+                  automation_secret=mysecret,
+                  validate_certs=False
+                  )
+         }}"
 
 - name: "Use variables outside the module call."
   ansible.builtin.debug:
-    msg: "Attributes of folder /network: {{ attributes }}"
+    msg: "Extended attributes of folder /network: {{ attributes.extensions }}"
```
robin-checkmk, line 107: Is this the only thing that we need to communicate to users?
"Update `{{ attributes }}` to `{{ attributes.extensions }}` and you are good to go"? :slightly_smiling_face:

robin-checkmk (changes requested):

plugins/lookup/folder.py
lgetwan: Yes, more or less, that's what makes this change _breaking_. It's a rather soft break. ;-)
Previously, the module returned the green part, now it returns the blue part:
image

changelogs/fragments/lookup_folder_examples.yml
lgetwan: I tried to improve the changelog.

plugins/lookup/folder.py
robin-checkmk: Great, thanks for confirming and updating the changelog!

robin-checkmk (approved):
Looking good, ready to merge!

PR #615, opened 2024-06-06
Improve idempotency of the registration tasks in the agent

robin-checkmk:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Both registrations, agent and updater were carried out by the `agent` role every time causing a change, even if the host was already successfully registered. While this does not break anything (given the configuration is sound), it is an inconvenience.

## What is the new behavior?

The `agent` role now checks, whether any of the registrations actually need to be performed.

## Other information

This change should be reviewed carefully, as the approaches chosen might not be ideal, or at least not without risks.
However, the worst case perceivable would be an unnecessary registration, we should not see anything actually break.


roles/agent/tasks/Linux.yml
```diff
@@ -87,6 +87,18 @@
...
+  become: true
+  ansible.builtin.command: cat /var/lib/cmk-agent/registered_connections.json
+  register: __checkmk_agent_registered_connections
+
+- name: "{{ ansible_system }}: Read Updater State."  # noqa no-changed-when
+  become: true
+  ansible.builtin.command: cat /var/lib/check_mk_agent/cache/plugins_cmk-update-agent.cache
+  register: __checkmk_agent_updater_state
+  changed_when: checkmk_agent_registration_server + '/' + checkmk_agent_registration_site in __checkmk_agent_updater_state.stdout
+  failed_when: __checkmk_agent_updater_state.rc == "255"
```
mueller-ma, line 100: I don't think `noqa no-changed-when` isn't required for these two tasks. To the first you can add `changed_when: false`, because it never changes something, the second already has a `changed_when` set.

Overall I really like to see this change. Currently we need to limit the agent role to a few hosts in serial, otherwise the load is too high on the CheckMK server caused by redudant tasks.

robin-checkmk: Thanks for the review!

I am always struggling with these kinds of tasks: On the one hand, they obviously do not change anything on the managed system, **but** technically a shell command is issued, which in my mind is kind of a change.
Ultimately, I will probably lean towards what you are saying and even simplify the logic for the second task, which really is not necessary.

Can you test this branch by the way and let me know, how it behaves out there in the real world?

mueller-ma: > but technically a shell command is issued, which in my mind is kind of a change.

In the end most Ansible tasks run commands. I guess (but haven't checked) that e.g. the `apt` module runs some `apt-get` command to check if a package needs to be installed.


I tested this branch on a Debian 12 target using Ansible 2.17.0:

````
$ ansible-playbook --version
ansible-playbook [core 2.17.0]
````

These tasks are changed:
````
TASK [checkmk.general.agent : Linux: Download host-specific CEE Agent.]
TASK [checkmk.general.agent : Linux: Read Agent Controller State.]
TASK [checkmk.general.agent : Linux: Read Updater State.]
TASK [checkmk.general.agent : Fetch fresh monitoring data from host.]
TASK [checkmk.general.agent : Update monitored services and labels on host.]
````

And with the latest stable version of the collection these tasks are changed:
````
TASK [checkmk.general.agent : Linux: Download host-specific CEE Agent.]
TASK [checkmk.general.agent : Linux: Register Agent for automatic Updates using Automation Secret.]
TASK [checkmk.general.agent : Linux: Register Agent for TLS.] 
TASK [checkmk.general.agent : Fetch fresh monitoring data from host.]
TASK [checkmk.general.agent : Update monitored services and labels on host.]
````

With the changes regarding `changed_when` two tasks are idempotent now 👍 
Maybe it's possible to make the other three tasks idempotent as well?

robin-checkmk: I pushed the changes pertaining to the registrations as discussed. I follow your logic.
For the remaining three tasks I will say this: They all are technically correct in being changed.
Now for download we might be able to check in advance, if a download is necessary, but that takes some time and testing (feel free to get on started on it, based on this branch). The latter two tasks by definition cannot be "ok", as they both actively do something. Especially fetching data. We could argue, that that is the same as above, but eventually the update task cannot be idempotent, as we have no means to compare the current versus the new state.

I really do appreciate this discussion!
However, I would like to keep this PR simple, get it mainlined and then maybe revisit and discuss the open points.
Is that alright with you @mueller-ma? :slightly_smiling_face:

mueller-ma: I don't know the CheckMK server api well enough, but maybe it could be enhanced by returning "nothing changed" when fetching new data and services? I'm fine with the current state of this PR and that the other tasks are kind of off-topic.

robin-checkmk: Great, then I will move forward with this PR, and we can revisit the open items at a later point. Feel free to open an issue for that. :v:

PR #624, opened 2024-06-13
Rebuild tests using Checkmk containers

robin-checkmk:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

We currently use the default container with `ansible-test` and install Checkmk into it on every test run.
This is not only lengthy, but has repercussions, e.g., just recently the OS of that container was bumped to a new version, breaking some of our test.

## What is the new behavior?

We use the GitHub Actions `services:` keyword, to start native Checkmk containers to run our tests against.

## Other information

This is a work in progress!

ToDo:
- Figure out how to handle managed edition -> Docker Login Timeout issue -> Download free?!
- https://github.com/orgs/community/discussions/128570


tests/integration/targets/activation/tasks/main.yml
```diff
@@ -1,13 +1,10 @@
 ---
 - name: "Include Global Variables."
-  ansible.builtin.include_vars: /root/ansible_collections/checkmk/general/tests/integration/files/includes/vars/global.yml
-
-- name: "Run Preparations."
-  ansible.builtin.include_tasks: /root/ansible_collections/checkmk/general/tests/integration/files/includes/tasks/prep.yml
```
lgetwan, line 6: To me, it looks like the prep.yml is not included anywhere, anymore. 
Is it still executed from somewhere? If not, why is it still in tests/integration/files/includes/tasks?

lgetwan (commented):
Looks like a lot of work!
I think I didn't get how the whole thing went into containers, based on the information I have here (changed files).

tests/integration/targets/activation/tasks/main.yml
lgetwan: After reading your description, I think, I got it! :man_facepalming: 
It's handled by the service section in the workflow definition, ok.

But still, I can't figure out where the prep.yml is called from, now.

robin-checkmk: We do not need the generic `prep.yml` anymore, as the containers are spun up from Checkmk images, which already contain the Checkmk site we need and `prep.yml` did nothing else.

lgetwan: Ok, that sounds good to me!

lgetwan (approved):

PR #629, opened 2024-06-21
Option create_only added to module host.py

hadu-iteratio:
Option create_only added. This is important for environments in which the initial installation and registration is done via Ansible, but the host is then to be changed manually or via a CMDB interface.
- [x ] Feature


msekania:
Hi @hadu-iteratio,

could please be more specific about what do you want achieve with this?

I might get your statement wrongly, but just to create a host with name the following task would be sufficient

```yaml
    - name: Create host test
      checkmk.general.host:
        server_url: "{{ my_checkmk_server }}"
        site: "{{ my_checkmk_site }}"
        automation_user: "{{ my_checkmk_automation_user }}"
        automation_secret: "{{ my_checkmk_automation_secret }}"
        name: "test"
        state: "present"
```
```

**Comment by hadu-iteratio** [comment]
Hi @msekania 

of course adding hosts already works today. However, today the hosts are always reset to their original state in Checkmk when the playbook or task is executed again. The option "create_only: True“ should prevent this behavior

Example: 

- The Checkmk agent package is installed on HostA, the HostA is created in Checkmk in the /new-agents/ folder and also  registered for agent updates. This is all done with the Ansible Collections .
- The HostA is then moved to another folder by an employee or it is manually given additional labels.
- Weeks later, the Linux is updated to a new OS release and all Ansible playbooks are executed against the update hosts, also the checkmk playbook
- As a result, the hosts is moved back to the /new-agent/ folder or all manual given attributes are removed.

This should be prevented by the "create_only" option. The host is only created in Checkmk if it does not exist. If it does exist, no changes are made to it. Currently its attributes/lables/tags would be reset or it would be moved back to its original folder.

Important:
This default behavior is the same as before because the default value for the create_only is false, so all existing playbooks or tasks out there are working as before.

**Comment by msekania** [comment]
> Currently its attributes/lables/tags would be reset or it would be moved back to its original folder.

Actually not!

use example above, then change attributes manually and use it again. all manually set features should remain.

which version of checkmk ansible collection are you using?

Can it be that problem is not hot plugin but rather agent?

**Comment by hadu-iteratio** [comment]
@msekania 

for your simple example this is true, but as soon as you add a folder or attributes, the behavior is different. Normally host are not created in the root folder or without any attributes. 
Following example

# Create a host.
  tasks:
  - name: "Create a host."
    checkmk.general.host:
      server_url: "{{ my_server }}"
      site: "{{ my_site }}"
      automation_user: "{{ my_automation_user }}"
      automation_secret: "{{ my_automation_secret }}"
      name: "{{ inventory_hostname_short }}"
      folder: "/new-agents/"
      state: "present"
      attributes:
        alias: "{{ ansible_fqdn }}"


Initially created

TASK [Create a host.] ********************************************************************************************************************************************************************
changed: [rocky9] => {"changed": true, "msg": "200 - Host created"}


Example: Host moved manually to another folder (/linux/):

TASK [Create a host.] ********************************************************************************************************************************************************************
changed: [rocky9] => {"changed": true, "msg": " (0), 200 - Host modified. Changed: folder (200), 200 - Host moved. Moved to: /new-agents (200)"}

Example: added tag

TASK [Create a host.] ********************************************************************************************************************************************************************
changed: [rocky9] => {"changed": true, "msg": " (0), 200 - Host modified. Changed: attributes: {\"alias\": \"rocky9.hdudomain.local\"} (200)"}

**Comment by robin-checkmk** [comment]
Hi @hadu-iteratio and thank you for your contribution!

Let me say up front, that I would have appreciated, if you stuck to the PR template, which would have made it way easier to understand, what you are trying to achieve.

I have one main concern with the proposed changed: It is basically against the Ansible philosophy. Ansible is a configuration management tool, aiming for idempotency. That means you want to be able to run Ansible (and it's modules) any number of times, and you want the result to always be the same. With your approach this breaks.

A module should not have a magic switch that subverts the expected behavior. Especially as we have several safeguards in place today, where you e.g., cannot modify attributes and the folder a host is in at the same time.

I think the solution you are looking for should be in the inventory, playbook, or maybe - just maybe - the role.
Let me elaborate: I would either use a variable or group in your inventory to create the desired behavior.
It is also perceivable to me, that it could be done at playbook level. Or just run the `agent` role with `checkmk_agent_add_host: false` and add the host beforehand. Just to lay out some ideas.

Of course, I am open for discussion, and if you can make a compelling case, I am of course open to implement the desired functionality.

**Comment by hadu-iteratio** [comment]
Hi @robin-checkmk,

Regarding the PR template:
You are absolutely right and I promise to be better next time

Regarding idempotency: 
Here again, you are of course basically right. But unfortunately the "pure doctrine" does not always correspond to reality. For many of our customers, hosts properties/attributes/labels are changed after the initial installation and registration.
And your module doesn't actually work like this all the time either.
The simple example of @msekania without folder and attribute specification would not reset the host to its original state, so idempotency is already broken
And the "safeguard" you mentioned is actually also inconsistent. 
Why should a host with changed attributes that is located in a different folder than when it was installed not be reset to its original state? Technically it would then be 2 API calls. But wouldn't that then correspond to the idempotency principle?

Nevertheless, I understand your comment and have decided to first check via the lookup modules of the collection whether a host exists and to register or not based on this.

**Comment by msekania** [comment]
@hadu-iteratio,

concerning my example, it does exactly what it should. It should not reset the folder at all. For that one should specify folder attribute explicitly.

I think there is a huge misunderstanding what module should do and what you intend to accomplish with it.

In your report it's task behaves exactly hoe the module was designed!

As @robin-checkmk, already mentioned, there are other ways to achieve the behavior you want (at least how I understood it).
In general more elaborate example with all ansible related and manually carries out changes, and explicit formulation what one tries to accomplish would be of great help, because it might be that we are talking about two different things.

**Comment by robin-checkmk** [comment]
Thanks @hadu-iteratio and @msekania for the lively and respectful discussion!
Above everything else, I want to make sure this project is a healthy and welcoming environment.

I have not double-checked all the points you made in your last comment @hadu-iteratio you might be right, that idempotency is not perfect. Regardless, in this case I am pretty certain, the functionality - even if it is not breaking anything - would create more confusion down the road.

However, I always appreciate insight into real-world scenarios and I want to make sure we accommodate every perceivable use-case, while keeping things lean and efficient. That being said: I think your use-case would be solved best by thinking of the agent role as a means for initial deployment. So I would run the agent role only once against a new system, installing the agent, adding it to Checkmk and registering the agent for updates. Then do not run the role again, as the agent updater handles the rest. Or course, there are many in-between solutions.

One tangible this I can see is this: Are the `defaults.yml` of the `agent` role maybe the source of your issue?
```yaml
checkmk_agent_folder: "{{ checkmk_var_folder_path | default('/') }}"
```
Because we tried to find sane defaults, but if you do not omit them, I can see them causing your troubles.
Setting them to be omitted for example or something else could help.

I will close the PR for now, but we can gladly continue the discussion!

PR #632, opened 2024-07-02
Fix removal of downtime when specifying only one service

maroessler:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When supplying only a single service in the downtime module, the downtimes are not correctly removed. Instead it says that there is no downtime found:

```
ok: [localhost] => {"changed": false, "msg": "'test-host/test-service' has no downtimes with comment 'Created by Ansible'."}
```

## What is the new behavior?
Downtimes are now correctly found and removed.

## Other information
Minimal example playbook:

```
- hosts: localhost
  tasks:
    - name: Set DT
      downtime:
        server_url: "{{ my_checkmk_server }}"
        site: "{{ my_checkmk_site }}"
        automation_user: "{{ my_checkmk_automation_user }}"
        automation_secret: "{{ my_checkmk_automation_secret }}"
        host_name: test-host
        state: present
        end_after:
          minutes: 10
        service_descriptions:
          - test-service

    - name: Pause
      ansible.builtin.pause:
        prompt: 'Check'

    - name: Remove DT
      downtime:
        server_url: "{{ my_checkmk_server }}"
        site: "{{ my_checkmk_site }}"
        automation_user: "{{ my_checkmk_automation_user }}"
        automation_secret: "{{ my_checkmk_automation_secret }}"
        host_name: test-host
        state: absent
        service_descriptions:
          - test-service
```


maroessler:

maroessler:
recheck

robin-checkmk:
@maroessler thanks for your contribution!
Someone will review the code eventually, after you fixed the linting finding.
Can I also ask you to add a test, that would detect this issue in the future?

maroessler:
Hi @robin-checkmk, sorry, this is my first time contributing, so I need to wrap around my head around everything.

You mean in tests/integration/targets/downtime/tasks/test.yml? I'm wondering about what should be the desired output here? When I take a look at those, I see some green and some yellow: https://github.com/Checkmk/ansible-collection-checkmk.general/actions/runs/9762644623/job/26946827206?pr=632

For example, this here: https://github.com/Checkmk/ansible-collection-checkmk.general/actions/runs/9762644623/job/26946827206?pr=632#step:6:344
It creates a downtime in the past and in the next step it removes it again which obviously fails. It seems like Checkmk does not keep this downtine since it's in the past anyway. At least, my tests seem to show that.

For "my" contribution, I would simple add the two tasks from my example playbook in there, would that be okay? The expected output would be that both tasks should cause a change.

robin-checkmk:
> Hi @robin-checkmk, sorry, this is my first time contributing, so I need to wrap around my head around everything.
You are good, what you already did looks very clean and it is a good PR!

> For "my" contribution, I would simply add the two tasks from my example playbook in there, would that be okay? The expected output would be that both tasks should cause a change.
I did not expect anything else. The idea is, that a task should fail, if the result is not expected. So OK and CHANGED are perfectly fine most of the time. In your case, it might be necessary to use a lookup to verify, if the downtime was properly created/deleted. If that is to complex for now, we can follow up later, no problem.

> It creates a downtime in the past and in the next step it removes it again which obviously fails. It seems like Checkmk does not keep this downtine since it's in the past anyway. At least, my tests seem to show that.
This is a good finding, we probably need to fix that with a similar approach as above. Feel free to add another PR to fix it, or open an issue about the situation, so we can fix it down the road.

robin-checkmk:
You could also look at the `assert` module and our lookup plugins. But that might be a little too advanced. As long as it works, I am fine with it.

maroessler:
I created the tests now with a simple "failed_when".

I will open an issue for the problem you mentioned here: https://github.com/Checkmk/ansible-collection-checkmk.general/pull/632#issuecomment-2203412153 tomorrow. A quick fix would probably to set the downtimes far into the future instead of in the past. But maybe someone else has a better idea.

maroessler:
Ah, I see now that I misaligned the register and failed_when clause. It should be fixed now.

Max-checkmk (approved):
Thanks for the PR!

Issue #547, opened 2024-02-01
host_tag_group Endpoint changed the parameters id/ident


lgetwan:
**For our Awareness**

[Werk 16364](https://checkmk.com/werk/16364) unifies the parameters of the host tag group endpoints

**Component Name**

Module tag_group

**Checkmk Version and Edition**

Checkmk < 2.2.0p??: There's a mixed usage of "id" and "ident"
2.2.0p?? < Checkmk < 2.4.0b1 : API always accepts both, "id" and "ident"
Checkmk > 2.4.0b1: Only "id" is accepted.

**Expected behavior**

No matter what Checkmk version the module contacts, it should always use the right parameter name.

**Actual behavior**

The current module version will stop working with Checkmk >= 2.4.0b1

robin-checkmk:
Implemented in #640.

PR #640, opened 2024-07-17
Taggroup update 2.4

Max-checkmk:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

- When creating or updating a taggroup we use the parameter "ident".
-> Issue Number: #547 

## What is the new behavior?

- With Werk 16364 we added the parameter "id". Beginning with 2.4.0 "ident" will be removed, so this plugin uses "id" if the version is 2.4 or higher.

## Other information
Added version check in the module and also made 1 class out of the 4 which existed before.

### DIFF
--- a/plugins/modules/tag_group.py
+++ b/plugins/modules/tag_group.py
@@ -130,10 +130,9 @@ from ansible_collections.checkmk.general.plugins.module_utils.types import RESUL
 from ansible_collections.checkmk.general.plugins.module_utils.utils import (
     result_as_dict,
 )
-
-# from ansible_collections.checkmk.general.plugins.module_utils.version import (
-#     CheckmkVersion,
-# )
+from ansible_collections.checkmk.general.plugins.module_utils.version import (
+    CheckmkVersion,
+)
 
 # We count 404 not as failed, because we want to know if the taggroup exists or not.
 HTTP_CODES_GET = {
@@ -142,7 +141,7 @@ HTTP_CODES_GET = {
 }
 
 
-def normalize_data(raw_data):
+def normalize_data(raw_data, ver):
     data = {
         "title": raw_data.get("title", ""),
         "topic": raw_data.get("topic", ""),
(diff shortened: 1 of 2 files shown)


robin-checkmk:
Fix confirmed, pending code review we are good to release this. :partying_face:

plugins/modules/tag_group.py
```diff
@@ -177,29 +191,28 @@ def post(self):
             return result
 
         else:
-            data = normalize_data(self.params)
-            data["ident"] = self.params.get("name")
+            data = normalize_data(self.params, self.ver)
+            if self.ver < CheckmkVersion("2.4.0"):
+                data["ident"] = self.params.get("name")
+            else:
+                data["id"] = self.params.get("name")
```
lgetwan, line 198: To me, this looks like it could be part of the "normalize_data()" function, as well.

plugins/modules/tag_group.py
```diff
@@ -286,35 +287,33 @@ def run_module():
...
+        if taggroup.current.http_code == 200:
             # If tag group has changed then update it.
-            if changes_detected(module, json.loads(current.content.decode("utf-8"))):
-                taggroupupdate = TaggroupUpdateAPI(module)
-                taggroupupdate.headers["If-Match"] = current.etag
-                result = taggroupupdate.put()
+            if changes_detected(
+                module, json.loads(taggroup.current.content.decode("utf-8"))
+            ):
+                taggroup.headers["If-Match"] = taggroup.current.etag
```
lgetwan, line 298: I would pot this line into the put() method, as well.

plugins/modules/tag_group.py
```diff
@@ -142,7 +141,7 @@
 }
 
 
-def normalize_data(raw_data):
+def normalize_data(raw_data, ver):
```
lgetwan, line 144: This could become a method of the TaggroupAPI class. Then, it would also have instant access to self.ver and self.data.

plugins/modules/tag_group.py
```diff
@@ -286,35 +287,33 @@ def run_module():
...
-            result = taggroupdelete.delete()
+        if taggroup.current.http_code == 200:
+            result = taggroup.delete()
 
             time.sleep(3)
-        elif current.http_code == 404:
+        elif taggroup.current.http_code == 404:
             result = RESULT(
                 http_code=0,
                 msg="Taggroup doesn't exist.",
```
lgetwan, line 319: "Taggroup already absent." would explain the state a bit better.

lgetwan (changes requested):
Looks like you did some nice cleanup work, as well. :-)
I suggested minor changes. They are only cosmetics but would complete the cleanup.

plugins/modules/tag_group.py
Max-checkmk: I wouldn't integrate this into the function, as this is only needed when creating a new taggroup

plugins/modules/tag_group.py
Max-checkmk: Line for the If-Match header is now in the method!

plugins/modules/tag_group.py
Max-checkmk: Changed

plugins/modules/tag_group.py
Max-checkmk: Done!

lgetwan (approved):
Passt!

PR #654, opened 2024-08-20
Implement module "site" for configuring distributed monitoring

lgetwan:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

There was no module for managing distributed monitoring sites.

## What is the new behavior?

Now there is.

## Other information

Thanks to the usage of IDs in the site management REST API, it was possible to make this module idempotent. If you run into problems, please let us know.


robin-checkmk (commented):
Can we discuss the module name?
The UI calls it "Distributed monitoring", but the REST API "Site Management".
My issue here is, that the latter sounds **a lot** like what our `server` role does.
We also follow the UI in other places IIRC, as the REST API made some interesting naming choices.

lgetwan:
I also disliked the module name tbo, so I changed it to _distmon_, which is short and recognizable. 
Also removed Python Typing to make it Python2.7 compatible.
Changelog was also added lately.

plugins/doc_fragments/distmon_options.py
```diff
@@ -0,0 +1,332 @@
...
+                - The settings of the site.
+            type: dict
+            suboptions:
+                login_data:
+                    description:
+                        - The authentication data for a configuration connection.
+                        - Only required when the O(state) is V(login).
+                    type: dict
+                    suboptions:
+                        remote_username:
```
robin-checkmk, line 31: I think just `user` would suffice, right? We explain, that the user needs admin privileges in the description anyway. And it should be obvious, that this user has to be from the remote site.

plugins/doc_fragments/distmon_options.py
```diff
@@ -0,0 +1,332 @@
...
+                            description:
+                                - A user with administrative permissions.
+                            type: str
+                        remote_password:
```
robin-checkmk, line 35: I think just `password` would suffice, right?

plugins/doc_fragments/distmon_options.py
```diff
@@ -0,0 +1,332 @@
...
+            description:
+                - The site ID to manage.
+            required: true
+            type: str
+        site_connection:
+            description:
+                - The settings of the site.
+            type: dict
+            suboptions:
+                login_data:
```
robin-checkmk, line 25: Nitpick: Maybe `authentication`?

plugins/doc_fragments/distmon_options.py
```diff
@@ -0,0 +1,332 @@
...
+                            suboptions:
+                                connection:
+                                    description:
+                                        - When connecting to sites on remote servers, please
+                                        - make sure that Livestatus over TCP is activated there.
+                                        - You can use UNIX sockets to connect to foreign sites on
+                                        - localhost.
+                                    type: dict
+                                    suboptions:
+                                        socket_type:
```
robin-checkmk, line 58: Maybe just `type`? Looking at the structure it would be `site_config.status_connection.connection.type`. I think that is clear enough.

plugins/doc_fragments/distmon_options.py
```diff
@@ -0,0 +1,332 @@
...
+                                            description:
+                                                - The IP or domain name of the host
+                                                - running the remote site.
+                                            type: str
+                                        path:
+                                            description:
+                                                - When the connection name is unix,
+                                                - this is the path to the unix socket.
+                                            type: str
+                                proxy:
```
robin-checkmk, line 85: Maybe this should actually be `liveproxy`, just to avoid confusion?

plugins/doc_fragments/distmon_options.py
```diff
@@ -0,0 +1,332 @@
...
+                                    description:
+                                        - The Livestatus Proxy Daemon configuration attributes.
+                                    type: dict
+                                    suboptions:
+                                        use_livestatus_daemon:
```
robin-checkmk, line 90: I am not happy with this bulky option. Maybe we abstract this like `enabled` with the values `true` and `false`, resulting in `with_proxy` and `direct` respectively. Not sure if that makes enough sense, but feels cleaner.

plugins/doc_fragments/distmon_options.py
```diff
@@ -0,0 +1,332 @@
...
+                                                    type: bool
+                                                    default: true
+                                connect_timeout:
+                                    description:
+                                        - The time that the GUI waits for a connection to the site
+                                        - to be established before the site is considered to be
+                                        - unreachable.
+                                    type: int
+                                    default: 2
+                                persistent_connection:
```
robin-checkmk, line 178: `persistent` should suffice.

plugins/doc_fragments/distmon_options.py
```diff
@@ -0,0 +1,332 @@
...
+                                        - I(is set to) V(with_proxy)!
+                                    type: dict
+                                    suboptions:
+                                        status_host_set:
+                                            description:
+                                                - enabled for 'use the following status host' and
+                                                - disabled for 'no status host'
+                                            type: str
+                                            choices: [enabled, disabled]
+                                            default: disabled
```
robin-checkmk, line 207: Shouldn't we be able to determine this from the provided values? No values mean `disabled`, values for `host` and `site` mean, it is enabled. Or am I missing something here?

plugins/doc_fragments/distmon_options.py
```diff
@@ -0,0 +1,332 @@
...
+                                        site:
+                                            description:
+                                                - The site ID of the status host.
+                                            type: str
+                                        host:
+                                            description:
+                                                - The host name of the status host.
+                                            type: str
+                                disable_in_status_gui:
```
robin-checkmk, line 216: Maybe `disable` is good enough here, as the details are in the description.

plugins/doc_fragments/distmon_options.py
```diff
@@ -0,0 +1,332 @@
...
+                                        - be shown in the status GUI. The replication is not affected
+                                        - by this, however.
+                                    type: bool
+                                    default: false
+                        configuration_connection:
+                            description:
+                                - A site's configuration connection.
+                            type: dict
+                            suboptions:
+                                enable_replication:
```
robin-checkmk, line 228: Just `replication`.

plugins/doc_fragments/distmon_options.py
```diff
@@ -0,0 +1,332 @@
...
+                                    description:
+                                        - Replication allows you to manage several monitoring sites
+                                        - with a logically centralized setup.
+                                        - Remote sites receive their configuration
+                                        - from the central sites.
+                                    type: bool
+                                    default: false
+                                url_of_remote_site:
```
robin-checkmk, line 236: Just `url`.

plugins/doc_fragments/distmon_options.py
```diff
@@ -0,0 +1,332 @@
...
+                                    default: true
+                                ignore_tls_errors:
+                                    description:
+                                        - This might be needed to make the synchronization accept
+                                        - problems with SSL certificates when using an SSL secured
+                                        - connection. We encourage you to always understand TLS issues
+                                        - and fix them, though!
+                                    type: bool
+                                    default: false
+                                direct_login_to_web_gui_allowed:
```
robin-checkmk, line 262: Maybe `login_to_web_gui`?

plugins/doc_fragments/distmon_options.py
```diff
@@ -0,0 +1,332 @@
...
+                                                - Sync with ldap connections.
+                                            type: str
+                                            choices: ['ldap', 'all', 'disabled']
+                                            default: all
+                                        ldap_connections:
+                                            description:
+                                                - A list of ldap connections to synchronize.
+                                            type: list
+                                            elements: str
+                                replicate_event_console:
```
robin-checkmk, line 291: Just `event_console`.

plugins/doc_fragments/distmon_options.py
```diff
@@ -0,0 +1,332 @@
...
+                                    description:
+                                        - This option enables the distribution of global settings and
+                                        - rules of the Event Console to the remote site. Any change in
+                                        - the local Event Console settings will mark the site as needing
+                                        - to sync. A synchronization will automatically reload
+                                        - the Event Console of the remote site.
+                                    type: bool
+                                    default: true
+                                replicate_extensions:
```
robin-checkmk, line 300: Just `extensions`.

plugins/doc_fragments/distmon_options.py
```diff
@@ -0,0 +1,332 @@
...
+                                        - activation of changes MKPs that are installed on your central site
+                                        - and all other files below the C($OMD_ROOT/local/) directory will be
+                                        - transferred to the remote site. All other MKPs and
+                                        - files below C($OMD_ROOT/local/) on the remote site will be removed.
+                                    type: bool
+                                    default: true
+                        secret:
+                            description:
+                                - The shared secret used by the central site to authenticate with the
+                                - remote site for configuring Checkmk.
```
robin-checkmk, line 312: This seems to be REST API specific, at least I cannot find it in the UI. Do we even need this here?

plugins/doc_fragments/distmon_options.py
```diff
@@ -0,0 +1,332 @@
...
+                                        - The alias of the site.
+                                    type: str
+                                customer:
+                                    description:
+                                        - The customer of the site (Managed Edition - CME only).
+                                    type: str
+                                site_id:
+                                    description:
+                                        - The site ID.
+                                    type: str
```
robin-checkmk, line 331: This seems redundant as we already have a `site_id` higher up, but it is probably technically not possible to merge them, right?

plugins/doc_fragments/distmon_options.py
```diff
@@ -0,0 +1,332 @@
...
+                                url_of_remote_site:
+                                    description:
+                                        - URL of the remote site including C(/check_mk/).
+                                        - This URL can be the same as the URL prefix of the status
+                                        - connection, but with C(/check_mk/) appended.
+                                        - Here it must always be an absolute URL, though.
+                                        - Unfortunately, this field is required by the REST API,
+                                        - even if there is no configuration connection enabled.
+                                    type: str
+                                    default: http://localhost/nonexistant/check_mk/
```
robin-checkmk, line 245: Do we set this intentionally to an unlikely hostname, or could we also go with a nicer name?

robin-checkmk (changes requested):
I am getting the overall feeling, that it might make sense, to abstract the module more from the REST API. The structure of the module could probably be flatter than that of the API. That is not to say, that this is not a good first version, we just might want to keep it in mind for a later review.

plugins/doc_fragments/distmon_options.py
lgetwan: Actually, we should open an internal ticket to fix this behavior of the REST API. I hope, they fix this quickly, so I suggest to just keep this "nonexistant" site.

plugins/doc_fragments/distmon_options.py
robin-checkmk: That makes sense. Then let's stick with approach 1 for the initial version, and we will see, what the feedback looks like. We can still do an iteration, if we learn, that the current approach does not work for users.

plugins/doc_fragments/distmon_options.py
robin-checkmk: "we" as in you will, or "we" as in I should do it? :wink:

lgetwan: Done.

plugins/doc_fragments/distmon_options.py
lgetwan: Done.

plugins/doc_fragments/distmon_options.py
lgetwan: Changing it to "username" was easier (because then, it's the same field name as used in the REST API).

plugins/doc_fragments/distmon_options.py
lgetwan: I will stick with the field name used by the REST API (see other comments).

plugins/doc_fragments/distmon_options.py
lgetwan: I will stick with the field name used by the REST API (see other comments).

plugins/doc_fragments/distmon_options.py
lgetwan: I will stick with the field name used by the REST API (see other comments).

plugins/doc_fragments/distmon_options.py
lgetwan: When doing a GET and then using the data gathered with a POST, this field should be used. Otherwise the central site will no longer be logged in to that remote site.
Maybe, this parameter is not needed, yet. But once, we provide a lookup module, it should be there. So, I suggest to keep it.

robin-checkmk: Maybe we can extend the documentation a little then? With what you just wrote it makes sense, while before I did not get the full picture.

lgetwan:
Currently, the integration tests are not correct. I have to refine them. But from my point of view, the module itself is ready for some first manual tests.

plugins/doc_fragments/distmon_options.py
lgetwan: To be honest, I didn't find a (currently supported) use-case that included setting this secret. Thus, I removed the parameter. The module only uses the secret internally to determine if Checkmk is logged in to a remote site.

plugins/doc_fragments/distmon_options.py
lgetwan: I will stick with the field name used by the REST API (see other comments).

plugins/doc_fragments/distmon_options.py
lgetwan: I will stick with the field name used by the REST API (see other comments).

plugins/doc_fragments/distmon_options.py
lgetwan: I will stick with the field name used by the REST API (see other comments).

plugins/doc_fragments/distmon_options.py
lgetwan: I will stick with the way the REST API handles this (see other comments).
Unfortunately, "status_host_set" is a mandatory field, even with config replication turned off.

plugins/doc_fragments/distmon_options.py
lgetwan: I will stick with the field name used by the REST API (see other comments).

plugins/doc_fragments/distmon_options.py
lgetwan: I will stick with the field name used by the REST API (see other comments).

lgetwan:
Looks like the integration tests runs into this problem: https://github.com/actions/runner-images/issues/6680
Probably, the tests take too long?

robin-checkmk:
> Looks like the integration tests runs into this problem: [actions/runner-images#6680](https://github.com/actions/runner-images/issues/6680) Probably, the tests take too long?

I'll investigate, but I doubt, it is a timeout issue, we have/had tests, which run longer. It might be resource exhaustion.

lgetwan:
It was indeed resource exhaustion. I reduced the number of remote sites to be created. Now github is happy.

robin-checkmk (approved):
I am happy with the changes. Pending a code review, we are good to merge.

changelogs/fragments/site.yml
```diff
@@ -0,0 +1,3 @@
+major_changes:
+  - Site module - Added a module for distributed monitoring.
+    Refer to the module documentation for further details.
```
robin-checkmk, line 3: I just realized, that we should document, that this module cannot accept the remote site's livestatus certificate.
At least in the changelog under "known_issues", but probably a note in the module documentation itself might make sense.

robin-checkmk: I added a hint, let me know, what you think.

robin-checkmk:
The linting failures are irrelevant to this PR. I will fix them in the `devel` branch.

changelogs/fragments/site.yml
lgetwan: That's an important advise for the user. Good idea.

mike1098:
I tested the following without issues against 2.2.0p33cee:

- Creating a site connection
- Loging in to a site
- Changing an attribute of a site connection
- Logout from a site
- Remove a site

Will discuss with my colleague as soon as he is back from vacation but I would say its working good.
Thank you for all the effort spent.

PR #656, opened 2024-09-06
Implementing inventory plugin for the Checkmk collection

Max-checkmk:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- It's not possible to use Checkmk as source for an ansible inventory

## What is the new behavior?

- This PR introduces the first version of the inventory plugin

## Other information

This first version of the plugin has the possibility to get tags and sites as groups.
Please comment or open an issue if you have suggestions for enhancements or ideas for future extensions of this plugin.


lgetwan:
I already did some tests. :-) 
Will continue tomorrow.

plugins/inventory/checkmk.py
```diff
@@ -0,0 +1,285 @@
...
+EXAMPLES = """
+# Getting all hosts incl. hosttags and sites as groups
+plugin: checkmk.general.checkmk
+server_url: "http://hostname/"
+site: "sitename"
+user: "cmkadmin"
+secret: "******"
+validate_certs: False
+groupsources: ["hosttags", "sites"]
+"""
```
lgetwan, line 67: Is it possible to extend the example to make it clear that calling
```
ansible-inventory -i inventory/checkmk.yml --list     # or
ansible-inventory -i inventory/checkmk.yml --list -vvvv
```
is the thing to do? I had a hard time figuring out that the way I called the command was correct, but it failed because of some misconfiguration in my site. To me, it looked like the inventory plugin wasn't used at all. Calling it with "-vvvv" would have shown me that it actually _is_ used.

.github/workflows/ans-unit-test-inventory.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+env:
+  NAMESPACE: checkmk
+  COLLECTION_NAME: general
+  MODULE_NAME: checkmk
+  TESTPATH: tests/unit/plugins/inventory/test_checkmk.py
+
+# concurrency:
+#   group: ${{ github.workflow }}-${{ github.ref }}
+#   cancel-in-progress: true
```
robin-checkmk, line 11: This can be enabled.

plugins/inventory/checkmk.py
```diff
@@ -0,0 +1,287 @@
...
+        automation_secret:
+            description: Automation secret for the REST API access.
+            required: true
+            type: string
+
+        validate_certs:
+            description: Whether or not to validate TLS certificates.
+            type: boolean
+            required: false
+            default: True
```
robin-checkmk, line 41: Did you try to use `extends_documentation_fragment: [checkmk.general.common]` like in the normal modules here? I know it does not work for lookup plugins, but maybe for inventories?

plugins/inventory/checkmk.py
robin-checkmk: I am also missing the explanation, how to exactly use the plugin, or is that documented elsewhere?

tests/unit/plugins/inventory/conftest.py
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env python
+
+import sys
+import importlib
+
+fake_lookupapi = importlib.import_module(
```
robin-checkmk, line 6: Maybe we want to use the term "mock" here (and probably elsewhere), to make clear, what is happening here?

robin-checkmk (changes requested):

changelogs/fragments/checkmk.yml
```diff
@@ -0,0 +1,2 @@
+major_changes:
+  - Inventory module - Added module for creating an dynamic inventory with Checkmk as source.
```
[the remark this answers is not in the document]
Max-checkmk: Done

.github/workflows/ans-unit-test-inventory.yaml
Max-checkmk: Enabled and already gone through the test

tests/unit/plugins/inventory/conftest.py
Max-checkmk: I think both words match and at least other unit tests (docker inventory as example) use "fake_api" too.
But i wouldn't care if we decide to change it here.

robin-checkmk: I can live with it, especially, if others do it the same way. :+1:

plugins/inventory/checkmk.py
```diff
@@ -0,0 +1,287 @@
...
+
+        plugin:
+            description: Name of the plugin. Should always be C(checkmk.general.checkmk)
+            type: string
+            required: true
+            choices: ['checkmk.general.checkmk']
+
+        groupsources:
```
robin-checkmk, line 25: When not declaring this option, we get an error `argument of type 'NoneType' is not iterable`.
Can we define a default value of e.g., `site` in the plugin, so users get at least some output and do not assume something is broken entirely? Especially, as this is marked as `required: false`. Or we need to make it required.

plugins/inventory/checkmk.py
```diff
@@ -0,0 +1,287 @@
...
+            type: string
+
+        validate_certs:
+            description: Whether or not to validate TLS certificates.
+            type: boolean
+            required: false
+            default: True
+
+        plugin:
+            description: Name of the plugin. Should always be C(checkmk.general.checkmk)
```
robin-checkmk: The naming looks odd, but I assume others do is the same way as well? IIRC Amazon uses e.g., `ec2` as the plugin name. Should we consider something like `checkmk.general.hosts`? That would even enable us to create inventory plugins for other elements at some point. Just a thought.

plugins/inventory/checkmk.py
robin-checkmk: I updated the EXAMPLES. What do you guys think, does that make enough sense?

plugins/inventory/checkmk.py
robin-checkmk: I implemented this and nothing complains, so I think we are good.

plugins/inventory/checkmk.py
lgetwan: Yes, that's exactly the part that I was missing.

plugins/inventory/checkmk.py
Max-checkmk: Just added a check for that.

plugins/inventory/checkmk.py
Max-checkmk: Like just discussed, we stay on checkmk.

robin-checkmk (approved):

lgetwan (approved):

Issue #663, opened 2024-09-26
[BUG] Create rule that already exists doesn't provide ruleid back


jupiler:
**Provide your feedback**
Create rule that already exists doesn't provide ruleid back.
It would be nice if the ruleid would be returned so it can be used in the future as a reference for the rules to be manipulated.

**Component Name**
Component Name: general.rule

**Ansible Version**

```
ansible-playbook [core 2.17.4]
  config file = ~/ansible.cfg
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/user/.venv/lib/python3.12/site-packages/ansible
  ansible collection location = /opt/user/.venv/collections
  executable location = /opt/user/.venv/bin/ansible-playbook
  python version = 3.12.3 (main, Sep 11 2024, 14:17:37) [GCC 13.2.0] (/opt/user/.venv/bin/python)
  jinja version = 3.1.4
  libyaml = True
```

**Collection Version**
```
collections:
  # https://github.com/Checkmk/ansible-collection-checkmk.general/tree/main
  - name: checkmk.general
    version: 5.2.1
```
**Environment**
Checkmk version 2.2.0p27
local os ubuntu 22.04

**Additional context**
The tasks in question.
```
- name: State_present | Create rule in {{ checkmk_rule_ruleset }}
  checkmk.general.rule:
    server_url: "{{ checkmk_rule_host }}"
    site: "{{ checkmk_rule_site }}"
    automation_user: "{{ checkmk_rule_user }}"
    automation_secret: "{{ checkmk_rule_pass }}"
    ruleset: "{{ checkmk_rule_ruleset }}"
    rule:
      rule_id: "{{ checkmk[rule_id_filename].ruleid |default('') }}"
      conditions:
        host_labels: "{{ rule.host_labels | default([]) }}"
        host_name:
          match_on: "{{ rule.host_names }}"
          operator: "{{ rule.host_labels | default('one_of') }}"
        host_tags: "{{ rule.host_tags | default([]) }}"
        service_labels: "{{ rule.service_labels | default([]) }}"
      properties:
        comment: Created with Ansible pipeline
        description: "{{ rule.host_names[0] }} - {{ rule.description | default('no description specified')}}"
        disabled: "{{ rule.disabled | default('False') | bool}}"
      value_raw: "{{ rule.value_raw | string }}"
      location:
[...]

PR #665, opened 2024-09-26
Fix: return rule id for existing rule

msekania:
returns content (includes rule id) field for existing and not modified rules

## Pull request type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #663

## What is the new behavior?

- return `content`, `http_code` for the case when rule exists and changes are not required


lgetwan (approved):
That looks good to me. I also successfully tested the new function.
Thanks for implementing this, Michael!

Issue #684, opened 2024-10-22
[BUG] 401 unauthorized using ansible collection


jgotteswinter:
Verify first that your issue is not already reported [here](https://github.com/Checkmk/ansible-collection-checkmk.general/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc).
Where possible also test if the latest release and main branch are affected too.
Complete all sections as described!

**Describe the bug**

When trying to use the MK Ansible collection i always end up in a 401 unauthorized. Calling the same url from my browser (using the automation secret) works fine, no auth issues. Using curl also works fine.

When using the Ansible collection it always ends up with a 401. This is a fresh installation i am currently working on.

The sites apache shows "AH01614: client used wrong authentication scheme:"  when trying to do something with the ansible collection.

**Component Name**
Component Name: activation

**Ansible Version**
```console
ansible [core 2.17.2]
  config file = None
  configured module search path = ['/Users/jgotteswinter/.ansible/plugins/modules', '/opt/local/share/ansible/plugins/modules']
  ansible python module location = /opt/local/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/ansible
  ansible collection location = /Users/jgotteswinter/.ansible/collections:/opt/local/share/ansible/collections
  executable location = /opt/local/bin/ansible
  python version = 3.12.7 (main, Oct  5 2024, 01:39:55) [Clang 16.0.0 (clang-1600.0.26.3)] (/opt/local/Library/Frameworks/Python.framework/Versions/3.12/bin/python3.12)
  jinja version = 3.1.4
  libyaml = True
```

**Checkmk Version and Edition**
```console
2.3.0p17 RAW
```

**Collection Version**
```console
Collection                               Version
---------------------------------------- -------
ansible.posix                            1.5.4
ansible.utils                            4.1.0
checkmk.general                          5.3.0
community.general                        8.6.0
```

**To Reproduce**

Fresh installation on Ubuntu 24.04, get the automation user secret and try a random task with the ansible collection

robin-checkmk:
Thanks for filing this issue @jgotteswinter! This feels like an issue with the system Apache on the server. Do you have any particular configuration there? A redirect or something else? Maybe the logs of said Apache contain a hint as well.

jgotteswinter:
[...]

PR #706, opened 2024-12-31
docs: clarify params server_url & site

HarisDotParis:
This PR clarifies docs that the parameters `server_url` & `site` are being concatenated for all API requests and that `site` is meant NOT to be part of the parameter `server_url`. This PR might be (un)related to #684.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently the docs don't specify that `server_url` must NOT contain the site appendix/directory as its part. This leads at least in my case to long troubleshootings of the modules with the error lying in this misunderstand. The creator @jgotteswinter of issue #684 might have also suffered from this docs shortcomming.

## What is the new behavior?

Docs are adjusted slightly to mention the parameters being concatenated and that `server_url` shall not contain the `site` value.


HarisDotParis:
recheck

HarisDotParis:

plugins/doc_fragments/common.py
```diff
@@ -7,11 +7,11 @@ class ModuleDocFragment(object):
     DOCUMENTATION = r"""
     options:
         server_url:
-            description: The base url of your Checkmk server including the protocol.
+            description: The base url of your Checkmk server including the protocol but excluding the site directory.
```
robin-checkmk, line 10: I would use 'site name' instead of 'site directory'.

plugins/doc_fragments/common.py
```diff
@@ -7,11 +7,11 @@ class ModuleDocFragment(object):
...
             required: true
             type: str
         site:
-            description: The site you want to connect to.
+            description: The site directory you want to connect to. This will be appended to the server_url as part of the API request url.
```
robin-checkmk, line 14: As above: 'directory' is not the correct terminology.

robin-checkmk (changes requested):

plugins/doc_fragments/common.py
HarisDotParis: I chose the word directory as the site is a directory of the URL path. If you don't agree with that word, I would rather omit any such word and just say `site` to avoid confusion. Which way do you prefer?

robin-checkmk (approved):

PR #714, opened 2025-01-16
Enable management of MKP packages with server role

ra2xfael:
This pull request tries to enable the installation and management of MKP packages specified by site. Packages can be downloaded or provided on the ansible controller and will be installed, enabled, disabled or removed from the sites. 

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

CheckMK MKP packages installation and management is not possible with the server role.

## What is the new behavior?

- Download of mkp packages using HTTP(S)
- Copying of mkp packages from the ansible controller
- MKP packages can be installed, enabled, disabled and removed from specific sites.


robin-checkmk:
Thanks for this contribution @ra2xfael, looks pretty neat! :muscle: 
Please try to satisfy the tests, as a review on our end might take some time.
But rest assured, we will get back to you when we can. :v:

roles/server/README.md
```diff
@@ -82,8 +82,17 @@ Note: this is not a recommended procedure, and will not be supported for enterpr
...
+          - name: 'mypackage'
+            version: 1.0
+            src: 'path'
+            url: 'download'
+            checksum: 'md5:download checksum'
+            installed: true
+            enabled: true
 
 A dictionary of sites, containing the desired version, admin password and state.
+MKP packages can be listed to be installed on the specific site. A source can be set on the ansible controller. Alternatively a url can be specified to download the mkp package from.
```
robin-checkmk, line 95: Not sure, if this is getting to convoluted here, but I got no better idea, so I am just mentioning it.

roles/server/defaults/main.yml
```diff
@@ -56,3 +56,5 @@ checkmk_server_epel_gpg_check: 'true'
 
 checkmk_server_cleanup: 'false'
 checkmk_server_no_log: 'true'
+
+checkmk_server_mkp_staging: '/tmp'
```
robin-checkmk, line 60: I think we should add a global tmp dir variable, like in the `agent` role: `__checkmk_agent_host_tmp_dir`.
That should obviously be done in a dedicated PR ahead of this one. Would you want to take that on @ra2xfael?

But why make it configurable in the first place?

robin-checkmk (commented):
So I have this problem: In distributed monitoring, MKPs will be replicated from a central to the remote sites upon activating changes. So MKP management would only ever make sense on a central site in this scenario.
Now your changes do enable this, as MKP management can be configured on a per-site basis. I am just uncertain at this point, if we need to build safeguards or if providing documentation and maybe a warning will suffice.

roles/server/defaults/main.yml
ra2xfael: You are right. I wanted to make it configurable and was unsure where to place the variable.
I can take a look to use the global variable in other tasks in another pull request

roles/server/README.md
ra2xfael: Its a bit verbose but I think it is easy to understand that way.
Make it easy to use >> Make it pretty ?

robin-checkmk: Oh, I agree: "easy to use" should be the priority for sure! I am just not sure, if that is the case here.
Maybe you take a peek at the argument_specs.yml, to see if that might be better suited to document this stuff and only hint on it here in the README?
We have to check and probably adapt argument_specs.yml anyway, so this might be a good chance.

I do not expect you to everything here, I am just discussing options. Of course, I appreciate, if you can look into them yourself. :)

robin-checkmk:
@ra2xfael I took the liberty to rebase this PR on `devel` after your other PRs were merged. Please double-check, if your changes are still good, then we can proceed to final review.

ra2xfael:
Thanks for the rebase. The changes are still valid so I'd welcome the review.

roles/server/README.md
robin-checkmk: So while adding integration tests to this new feature in #723 (cannot base the PR on this PR), I realized, that the documentation definitely needs improving.
For example: What exactly would `url` be?
What happens, if `src` and `url` are provided at the same time?

lgetwan (approved):
I like the idea to handle MKPs with the server role.
Assuming that the tests are successful, I'd approve this PR.

roles/server/README.md
ra2xfael: I see the problem. The documentation is lacking.
My idea was that either a download link or a local path on the ansible controller can be specified as a source for the package. In case of a download link the checksum can be used to verify the authenticity of the package (for example if the download happens from the CheckMK Exchange).
The checksum uses the underlying implementation of the ansible.builtin.get_url module, so it should work.

ra2xfael: Sure, I am currently running tests with the checksum and report back once I have verified them working.
You are welcome to improve the documentation. I think it is important to mention that the checksum only works with the url directive. Maybe it is necessary to rename the variables to make this clearer?

robin-checkmk: I'll look into it as soon as I can and let you know. Thanks!

ra2xfael: I had no issue installing (and verifying) mkp packages with the checksum set with this test configuration
```
    mkp_packages:
      - name: SSL-Certificates
        version: 8.9.1
        url: https://exchange.checkmk.com/packages/sslcertificates/1570/sslcertificates-8.9.1.mkp
        checksum: "md5:d759c6be90fdbe2bb3a27e15fe2488b0"
```
The formality of the checksum is odd but it is working flawlessly according to the [get_url module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/get_url_module.html#parameter-checksum)

robin-checkmk: Ah, there is my issue: I assumed I only need the "root URL" to the package and that the "final" URL would be generated by the tasks. I'll circle back to the tests and let you know then.

robin-checkmk:
@lgetwan and @ra2xfael I am not happy with the changes. Would y'all take a last peek at the README and defaults?
Just in case, something still feels inconsistent. But I am now very much with this feature. :rocket: 

Thanks again for your contribution @ra2xfael! :hugs:

ra2xfael:
Hi @robin-checkmk 
the changes look good, I think the Readme is precise and the defaults are a good example. The molecule tests also seem sound. Thanks for the support!

Issue #709, opened 2025-01-13
[BUG] `discovery` module uses deprecated parameter `mode`


robin-checkmk:
**Describe the bug**

The formerly deprecated Checkmk REST API parameter `mode` for the bulk service discovery is removed in Checkmk 2.4.0, hence the newly introduced `options` parameter has to be used by the `discovery` module.

**Component Name**

Component Name: discovery

**Ansible Version**

```console
$ ansible --version

```

**Checkmk Version and Edition**

```console
2.4.0
```

**Collection Version**

```console
$ ansible-galaxy collection list

```

**To Reproduce**
1. Take a look at [this GitHub Actions run](https://github.com/Checkmk/ansible-collection-checkmk.general/actions/runs/12748129717/job/35535604997).

**Expected behavior**

The module should detect the Checkmk version and use `mode` and `options` accordingly.

**Actual behavior**

The module fails.

**Minimum reproduction example**

See the linked GitHub Actions run above.

**Additional context**

PR #716, opened 2025-01-24
Make discovery module compatible to Checkmk 2.4.0.

lgetwan:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: #709 

## What is the new behavior?

- In versions 2.4.0 and newer, the modes tabula_rasa and refresh are no longer available,
 in that case, we perform a add/remove all services and labels, instead.

## Other information

The API parameter mode is gone for the bulk discovery endpoint in Checkmk 2.4.0. In 2.3.0 it was already deprecated. Instead, the parameter "mode" was introduced. Thus, we use this mapping:
| State                       | monitor_undecided_services | remove_vanished_services | update_service_labels | update_host_labels |
| --------------------------- | -------------------------- | ------------------------ | --------------------- | ------------------ |
| new                         | X                          |                          |                       | X                  |
| fix_all                     | X                          | X                        |                       | X                  |
| monitor_undecided_services  | X                          |                          |                       |                    |
| refresh                     | X                          | X                        | X                     | X                  |
| tabula_rasa                 | X                          | X                        | X                     | X                  |
| only_service_labels         |                            |                          | X                     |                    |
| only_host_labels            |                            |                          |                       | X                  |
| remove                      |                            | X                        |                       |                    |

### DIFF
--- a/.github/workflows/ans-int-test-discovery.yaml
+++ b/.github/workflows/ans-int-test-discovery.yaml
@@ -81,7 +81,7 @@ jobs:
         ports:
           - 5323:5000
         env:
-          CMK_SITE_ID: "stable_cme"
+          CMK_SITE_ID: "old_cme"
           CMK_PASSWORD: "Sup3rSec4et!"
       stable_cre:
         image: checkmk/check-mk-raw:2.4.0-daily
--- a/plugins/modules/discovery.py
+++ b/plugins/modules/discovery.py
@@ -240,13 +240,23 @@ class newBulkDiscoveryAPI(CheckmkAPI):
             "update_host_labels": False,
         }
 
-        if self.params.get("state") in ["new", "fix_all", "monitor_undecided_services", "refresh"]:
+        if self.params.get("state") in [
+            "new", 
+            "fix_all", 
+            "monitor_undecided_services", 
+            "refresh"
+        ]:
             options["monitor_undecided_services"] = True
         if self.params.get("state") in ["remove", "fix_all", "refresh"]:
             options["remove_vanished_services"] = True
         if self.params.get("state") in ["only_service_labels", "refresh"]:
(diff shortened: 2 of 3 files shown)


lgetwan:
Looks like we can merge this now, @robin-checkmk. :+1:

plugins/modules/discovery.py
```diff
@@ -366,8 +372,8 @@ def run_module():
 
     ver = discovery.getversion()
 
-    if single_mode and ver < CheckmkVersion("2.1.0"):
-        discovery = oldDiscoveryAPI(module)
+    # if single_mode and ver < CheckmkVersion("2.1.0"):
+    #     discovery = oldDiscoveryAPI(module)
```
robin-checkmk, line 376: Can we remove these lines completely, or does the comment serve a deeper purpose?

.github/workflows/ans-int-test-discovery.yaml
```diff
@@ -81,7 +81,7 @@ jobs:
         ports:
           - 5323:5000
         env:
-          CMK_SITE_ID: "stable_cme"
+          CMK_SITE_ID: "old_cme"
```
robin-checkmk, line 84: Canyou fix this? It should read "stable_cme".

robin-checkmk (changes requested):

plugins/modules/discovery.py
lgetwan: Yes, we can!
And I did.

.github/workflows/ans-int-test-discovery.yaml
lgetwan: I'd keep it this way, as it resolves the previous errors in the integration test.

PR #734, opened 2025-02-21
Trying to handle the changes that came with werk #16466.

lgetwan:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The discovery module is not handling the changes that came with Werk [#16466](https://checkmk.com/werk/16466)

## What is the new behavior?

The discovery module works with all supported Checkmk versions up to to 2.4.0

### DIFF
--- a/plugins/module_utils/discovery_240.py
+++ b/plugins/module_utils/discovery_240.py
@@ -74,7 +74,7 @@ class ServiceDiscoveryAPI(CheckmkAPI):
 class ServiceBulkDiscoveryAPI(CheckmkAPI):
     def post(self):
         mode = self.params.get("state")
-        if mode not in COMPATIBLE_MODES or mode == "refresh":
+        if mode not in COMPATIBLE_MODES:
             return generate_result(
                 msg="State %s is not supported with this Checkmk version." % mode
             )
@@ -86,13 +86,30 @@ class ServiceBulkDiscoveryAPI(CheckmkAPI):
             "update_host_labels": False,
         }
 
-        if self.params.get("state") in ["new", "fix_all", "monitor_undecided_services"]:
+        if self.params.get("state") in [
+            "new",
+            "fix_all",
+            "refresh",
+            "tabula_rasa",
+            "monitor_undecided_services",
+        ]:
             options["monitor_undecided_services"] = True
-        if self.params.get("state") in ["remove", "fix_all"]:
+        if self.params.get("state") in ["remove", "fix_all", "refresh", "tabula_rasa"]:
             options["remove_vanished_services"] = True
-        if self.params.get("state") in ["only_service_labels"]:
+        if self.params.get("state") in [
+            "only_service_labels",
+            "fix_all",
+            "refresh",
+            "tabula_rasa",
+        ]:
             options["update_service_labels"] = True
-        if self.params.get("state") in ["new", "fix_all", "only_host_labels"]:
+        if self.params.get("state") in [
+            "new",
+            "fix_all",
+            "only_host_labels",
+            "refresh",
+            "tabula_rasa",
+        ]:
             options["update_host_labels"] = True
 
         data = {
--- a/plugins/module_utils/discovery_240.py
+++ b/plugins/module_utils/discovery_240.py
@@ -38,6 +38,31 @@ COMPATIBLE_MODES = [
     "monitor_undecided_services",
 ]
 
+MONITOR_UNDECIDED_SERVICES = [
+    "new",
+    "fix_all",
+    "refresh",
+    "tabula_rasa",
+    "monitor_undecided_services",
+]
+
+REMOVE_VANISHED_SERVICES = ["remove", "fix_all", "refresh", "tabula_rasa"]
+
+UPDATE_SERVICE_LABELS = [
+    "only_service_labels",
+    "fix_all",
+    "refresh",
+    "tabula_rasa",
+]
+
+UPDATE_HOST_LABELS = [
+    "new",
+    "fix_all",
+    "only_host_labels",
+    "refresh",
+    "tabula_rasa",
+]
+
 SUPPORTED_VERSIONS = {
     "min": "2.4.0",
     "max": "2.4.0p99",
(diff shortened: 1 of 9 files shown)


lgetwan:
The "Wait for completion" mechanism changed with the 2.4.0.
We have to adapt the changes in the discovery module.

lgetwan:
Ok, the integration tests are now also successful. And they are quite extensive for the discovery module.
Who can review the PR from a coding perspective, @Max-checkmk, @msekania (even though it was a complete rewrite, It still contains some lines of code from you. ;-)), @robin-checkmk?

msekania:
@lgetwan,

I will review the code today, tomorrow at the latest.

plugins/module_utils/discovery_240.py
```diff
@@ -0,0 +1,241 @@
...
+            "update_host_labels": False,
+        }
+
+        if self.params.get("state") in [
+            "new",
+            "fix_all",
+            "refresh",
+            "tabula_rasa",
+            "monitor_undecided_services",
+        ]:
```
msekania, line 95: I think it's better to introduce the following variable (and few similar of this type, below)

```
MONITOR_UNDECIDED_SERVICES = [
    "new",
    "fix_all",
    "refresh",
    "tabula_rasa",
    "monitor_undecided_services",
]
```

plugins/module_utils/discovery_240.py
```diff
@@ -0,0 +1,241 @@
...
+            options["monitor_undecided_services"] = True
+        if self.params.get("state") in ["remove", "fix_all", "refresh", "tabula_rasa"]:
+            options["remove_vanished_services"] = True
+        if self.params.get("state") in [
+            "only_service_labels",
+            "fix_all",
+            "refresh",
+            "tabula_rasa",
+        ]:
```
msekania, line 104:
```
UPDATE_SERVICE_LABELS = [
    "only_service_labels",
    "fix_all",
    "refresh",
    "tabula_rasa",
]
```

plugins/module_utils/discovery_240.py
```diff
@@ -0,0 +1,241 @@
...
+        if self.params.get("state") in [
+            "new",
+            "fix_all",
+            "refresh",
+            "tabula_rasa",
+            "monitor_undecided_services",
+        ]:
+            options["monitor_undecided_services"] = True
+        if self.params.get("state") in ["remove", "fix_all", "refresh", "tabula_rasa"]:
+            options["remove_vanished_services"] = True
```
msekania, line 98:
```
REMOVE_VANISHED_SERVICES = [
    "remove",
    "fix_all",
    "refresh",
    "tabula_rasa",
]
```

plugins/module_utils/discovery_240.py
```diff
@@ -0,0 +1,241 @@
...
+            "tabula_rasa",
+        ]:
+            options["update_service_labels"] = True
+        if self.params.get("state") in [
+            "new",
+            "fix_all",
+            "only_host_labels",
+            "refresh",
+            "tabula_rasa",
+        ]:
```
msekania, line 112:
```
UPDATE_HOST_LABELS = [
    "new",
    "fix_all",
    "only_host_labels",
    "refresh",
    "tabula_rasa",
]
```

plugins/module_utils/discovery_240.py
```diff
@@ -0,0 +1,241 @@
...
+        result = self.discovery_api.post()
+        if result.failed:
+            return result
+
+        if self.wait_for_completion:
+            if self.single_mode and result.http_code != 200:
+                result = self._wait_for_completion("current")
+            elif self.bulk_mode:
+                job_id = json.loads(result.content).get("id")
+                result = self._wait_for_completion("current", job_id=job_id)
```
msekania, line 239: again, it might timeout here, but it does not mean that the process will not be finished successfully after some time without intervention from the user side.

plugins/module_utils/discovery_240.py
```diff
@@ -0,0 +1,241 @@
...
+                failed=False,
+                changed=True,
+            )
+
+        while True:
+            now = time.time()
+            if now > deadline:
+                return generate_result(
+                    msg="Timeout reached while waiting for %s discovery" % what
+                )
```
msekania, line 204: I understand the idea of `timeout` and it might be appealing but I see couple of issues which should be clarified:
- when it timeouts here it does not mean that process is still not running in background,
- it does not return failed state and next call of rediscovery might fail.

If one still decides to go with timeout concept, then one should also make clear in documentation that in essence it might take up to twice the `timeout` before plugin returns timeout, but it also does not mean that the process will not be finished successfully after some time without intervention from the user side.

plugins/modules/discovery.py
```diff
@@ -58,11 +62,25 @@
...
+        type: bool
+        default: True
+    wait_for_previous:
+        description: If true, wait for previously running discovery jobs to finish.
+        type: bool
+        default: True
+    wait_timeout:
+        description: The time in seconds to wait for (previous/current) completion.
+        type: int
+        default: 15
```
msekania, line 76: In order to qualify as a non-breaking change, the default value of this parameter should be "infinite".

In essence consider `timeout` only if `wait_timeout` was explicitly set (no default value), otherwise treat unset `wait_timeout` as wait what it takes.

msekania (commented):
@lgetwan,

First of all, I like the idea to split module in version dependent parts!
The logic of the code is clearer and easier to follow in this form!

I commented on `discovery_240.py` file only for the moment in order to clarify the logic change introduced by the "wait" concept.
Otherwise, the comments would be very similar to other version-parts of the code.

plugins/module_utils/discovery_240.py
lgetwan: I like that idea!

plugins/module_utils/discovery_240.py
lgetwan: I like that idea!

plugins/modules/discovery.py
lgetwan: Done.

plugins/module_utils/discovery_240.py
lgetwan: The use-cases I had in mind, were:
1. I only want to trigger the discovery, and I don't mind if it finishes successfully. I might check later in the playbook
2. I want to wait for the discovery to finish, and the play should only continue if the discovery was successful.
3. A mix: 99% of my discoveries finish within 30s. I don't want to wait longer in my playbook. The 1% that fails (or succeeds in background), I will check manually, later or in another playbook. (~ Pareto principle. ;-))

After implementing the default "infinite", these should be possible, with "2." being the default, to make the change non-breaking.

Does that make sense to you? Are there other use-cases that you can think of?

plugins/module_utils/discovery_240.py
lgetwan: That's true, and that was by intention (see my use-case "1." in the other comment).

plugins/module_utils/discovery_240.py
msekania: > Does that make sense to you? Are there other use-cases that you can think of?

Yes! I think that's it!

I will check the changed implementation more closely today, with a quick glance something is wrong.

msekania: @lgetwan,

v2.2 and v2.1 have not yet received the updates (`wait_time: -1` treatment), right?

lgetwan: Yes, 2.2 and 2.1 were missing, but I had to leave into Feierabend. ;-)
Fixed that this morning.

msekania:
@lgetwan,

It's fine from my point of view!

lgetwan:
Thanks a lot, @msekania!

@robin-checkmk,
Looks like we can merge this PR. :-)

PR #745, opened 2025-03-15
Add possibility to provide URL/URI download timeout

LennertMertens:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Currently we are facing issues when downloading the agent files, especially on Windows Server 2025. It takes a little longer than 30 seconds, so the task fails prior to completing.

## What is the new behavior?

- The possibility of providing an agent download timeout gives us the flexibility to ensure the agent is properly installed without having to do some hacky pre-tasks.

## Other information

This would not break anything since this variable is omitted by default, you CAN provide a timeout but you don't have to. By default the modules defaults will be used.


LennertMertens:

roles/agent/README.md
```diff
@@ -182,6 +182,10 @@ If you are using the Checkmk Cloud Edition (CCE) with an agent in `push` mode, y
 
 Whether to log sensitive information like passwords, Ansible output will be censored for enhanced security by default. Set to `false` for easier troubleshooting. Be careful when changing this value in production, passwords may be leaked in operating system logs.
 
+    checkmk_agent_download_timeout: 30
```
robin-checkmk, line 185: The default value for the `ansible.builtin.get_url` module is 10 seconds (https://docs.ansible.com/ansible/latest/collections/ansible/builtin/get_url_module.html). Interestingly for `ansible.windows.win_get_url` it is 30 seconds (https://docs.ansible.com/ansible/latest/collections/ansible/windows/win_get_url_module.html).

I was about to propose setting the value to the default of the modules, but those do not match.
The idea of suggesting there is a default in the README does not sit well with me. At the same time, we never had this before, usually there is a default.

We could do something complicated like `checkmk_agent_download_timeout: "{% if ansible_system == 'Win32NT' %}30{% else %}10{% endif %}"`.

What do you think @LennertMertens?

robin-checkmk (commented):
In your initial description you said the download takes long on a Windows server. As the default timeout for a Windows server is 30 seconds, and you set it to 30 seconds here, I am curious, if the issue matches the solution.

I am not saying this feature makes no sense, I am just trying to understand the extent and impact.

roles/agent/README.md
LennertMertens: @robin-checkmk what you propose seems like a good solution to me, however at this time I just ignore the timeouts completely if the `checkmk_agent_download_timeout` is ommited, so the modules fall back to their default values. So I'm not entirely sure if an entry in the README is required then. If so, I would go with what you suggest!

robin-checkmk: I mean, we need to document the variable in the README, as other users might be facing the same challenge as you did. But I want to make sure, it is well understood and not used lightly.

Do you want to update the PR, or should I take a swing, and you can review it afterwards, if it is still in line with your proposal?

LennertMertens: I'm glad to review your proposed changes!

robin-checkmk: @LennertMertens for some reasons I cannot push to this PR right now. Would you mind reviewing and implement the following changes?

```
diff --git a/roles/agent/README.md b/roles/agent/README.md
index 0cc042d5..463a63b4 100644
--- a/roles/agent/README.md
+++ b/roles/agent/README.md
@@ -182,9 +182,11 @@ If you are using the Checkmk Cloud Edition (CCE) with an agent in `push` mode, y
 
 Whether to log sensitive information like passwords, Ansible output will be censored for enhanced security by default. Set to `false` for easier troubleshooting. Be careful when changing this value in production, passwords may be leaked in operating system logs.
 
-    checkmk_agent_download_timeout: 30
+    checkmk_agent_download_timeout: "{% if ansible_system == 'Win32NT' %}30{% else %}10{% endif %}"
+
+This setting can be used to increase the timeout in seconds for downloading the Checkmk agent from the Checkmk server. Only use this, if you encounter
+issues with the agent download. There is no role default, the module defaults will be used.
 
-The timeout interval that is respected when trying to download the Checkmk agent. This is an optional variable and not required by default. If you omit this, the module's default values will be used.
 
 ## Tags
 
diff --git a/roles/agent/defaults/main.yml b/roles/agent/defaults/main.yml
index aab780f6..0d93805e 100644
--- a/roles/agent/defaults/main.yml
+++ b/roles/agent/defaults/main.yml
@@ -39,6 +39,12 @@ checkmk_agent_host_attributes:
 checkmk_agent_mode: 'pull'
 checkmk_agent_no_log: 'true'
 
+# The following setting can be used to increase the timeout for downloading
+# the Checkmk agent from the Checkmk server. Only use this, if you encounter
+# issues with the agent download. There is no role default, the module defaults
+# will be used. Value in seconds.
+# checkmk_agent_download_timeout: "{% if ansible_system == 'Win32NT' %}30{% else %}10{% endif %}"
+
 # If you trust your local hostnames, you could also use the following
 # to use the local hostname instead of the inventory hostname:
 # checkmk_agent_host_name: "{{ hostvars[inventory_hostname]['ansible_fqdn'] }}"
```

LennertMertens: Pushed the changes! Thanks for adding this!

robin-checkmk (approved):

roles/agent/README.md
robin-checkmk: Thanks, then we are all clear now! I will merge this request, but the release might be a little while out still. :slightly_smiling_face:

PR #750, opened 2025-04-02
Make gpg key download url configurable

ra2xfael:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

When installing the CheckMK server on a backend system with no access to the internet via a local mirror, the downloaded package can not be verified via the gpg key. The role tries to download the gpg key from https://download.checkmk.com/checkmk even if another mirror is used. This leads to the necessary deactivation of `checkmk_server_verify_setup` if `checkmk_server_download_url` is overwritten. 

## What is the new behavior?

- Introduction of the variable `checkmk_server_gpg_download_url` which is set to `{{ __checkmk_server_base_url }}/Check_MK-pubkey.gpg`
- If the `checkmk_server_download_url` needs to be modified the `checkmk_server_gpg_download_url` can be updated too.

## Other information

This change makes the *verified* installation on systems without internet access possible.

### DIFF
--- a/changelogs/fragments/gpg.yml
+++ b/changelogs/fragments/gpg.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - Server role - Make GPG key download URL configurable.
--- a/roles/server/tasks/main.yml
+++ b/roles/server/tasks/main.yml
@@ -54,7 +54,7 @@
 
     - name: "Download Checkmk GPG Key."
       ansible.builtin.get_url:
-        url: "https://download.checkmk.com/checkmk/Check_MK-pubkey.gpg"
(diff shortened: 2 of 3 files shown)


ra2xfael:
I just needed the feature for my own work, so no circumstances there.

What type of test are you thinking about? Spin up a separate container with a mirror or let the installation fail on purpose?

robin-checkmk:
That is actually a good question. You know what? I will approve this request as is, because I like the approach and it makes sense. If you have a good idea for a test case, I appreciate another PR and if I can come up with something meaningful, I will add it as well.

robin-checkmk (approved):

PR #542, opened 2024-01-30
[Draft] Add module aux_tag

muehlings:
## Pull request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

There is no module for creating, editing or deleting aux tags.

## What is the new behavior?

[Draft] Module for creating, editing or deleting aux tags.

## Other information

Third try of generating Ansible modules from OpenAPI definition for Checkmk. This code is 99% generated from the checkmk.yaml file in SwaggerUI. Manually changed parts are around the [...] brackets. Module is based on Checkmk 2.2.0p20-CME YAML. Now with a class for the managed object/endpoint.

PR #752, opened 2025-04-03
Add aux_tags suboption for tags in tag_group module

meni2029:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The aux_tags suboption for the tags in the tag_group module is not available, while it was possible to declare it in previous collection version (i.e. 3.4.0).

## What is the new behavior?

- Added the aux_tags suboption for the tags in the tag_group module
- The suboption is optional and default value is an empty list ([])
- The aux_tags is now used in the `changes_detected` function

## Other information

It has been tested on Checkmk 2.3.0p27. It might not fully work with 2.2 as, if I remember correctly, the API returns no "aux_tags" item if there's none selected. At least in 2.3, the API always returns it with empty list if none selected, so default => []:
```
"topic": "Notifications",
    "tags": [
        {
            "id": "yes",
            "title": "of course",
            "aux_tags": []
        }
```


robin-checkmk:
Hi @meni2029 and thank you for your contribution! Looks good to me so far.
Do you think you can add one or two integration tests, which verify that the feature works as intended?
That would be highly appreciated! If not, I can take care of it as well.

meni2029:
Hi @robin-checkmk ,
I added integration tests. For the tests I create the aux_tags with direct API call, as there's no module for aux_tags.
Hopefully this is ok.
Cheers

meni2029:
Hi @robin-checkmk , @lgetwan 
Thanks for looking into that.
Regarding the integration tests using direct API calls, to me it is not much of an issue, as it is for the tests, not inside the actual module.
Of course it would be better to have a module for `aux_tags`, but not easy to create it. One challenge is the dependency between `tag_group` and `aux_tags`, when creating and also supressing.
@muehlings made a proposition a year ago: https://github.com/Checkmk/ansible-collection-checkmk.general/pull/542. I don't really have the time to dig into it.
On our side at the moment we handle `aux_tags` with direct API calls, and use `tag_group` module.
I think it makes sense to do this small step forward and merge this MR ;)

robin-checkmk (approved):

PR #758, opened 2025-04-09
User role: No longer allow to change the user that is used for API authorization.

lgetwan:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

From 2.4.0 on, the etag of a user changes with every API call, if the user is also used for API auth.
This made it impossible to e.g. change the cmkadmin user when authorizing against the API as cmkadmin.

## What is the new behavior?

The user module now prevents from using the same user for auth that you are actually changing.

### DIFF
--- a/changelogs/fragments/user.yml
+++ b/changelogs/fragments/user.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - User role - No longer allow to change the user that is used for API authorization.
+  - User role - No longer allow to modify the user that is used for API authorization.
--- a/changelogs/fragments/user.yml
+++ b/changelogs/fragments/user.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - User role - No longer allow to change the user that is used for API authorization.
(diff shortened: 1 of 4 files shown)


plugins/modules/user.py
```diff
@@ -507,6 +507,18 @@ def run_module():
 
     module = AnsibleModule(argument_spec=module_args, supports_check_mode=False)
 
+    if module.params.get("automation_user") == module.params.get("name"):
+        result = RESULT(
+            http_code=0,
+            msg="Can't modify the user that is used for API authorization."
```
robin-checkmk, line 513: "You cannot modify the user that is used for API authorization" would be a bit nicer here, right? It is a user-facing message, rather than a technical one, right?

changelogs/fragments/user.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - User role - No longer allow to change the user that is used for API authorization.
```
robin-checkmk, line 2: "modify" instead of "change" is probably better and in line with the wording in the module.

Issue #764, opened 2025-05-05
[FEED] incompatible and incomplete environment variables support for connection credentials


penguincustodian:
**Provide your feedback**
Currently only the lookup plugins from this collection support passing the connection credentials via environment variables, inventory and module plugins don't:

```
export ANSIBLE_LOOKUP_CHECKMK_SERVER_URL="https://myserver"
export ANSIBLE_LOOKUP_CHECKMK_SITE=mysite
export ANSIBLE_LOOKUP_CHECKMK_AUTOMATION_USER=automation
export ANSIBLE_LOOKUP_CHECKMK_AUTOMATION_SECRET=mysecret
export ANSIBLE_LOOKUP_CHECKMK_VALIDATE_CERTS=False
```

This was implemented in https://github.com/Checkmk/ansible-collection-checkmk.general/pull/546.

However, there is a problem with this implementation, when using AWX/Ansible Automation Platform: It is disallowed to create a custom Credential Type with "ANSIBLE_*" env-vars injector configurations: 

```
Environment variable ANSIBLE_LOOKUP_CHECKMK_AUTOMATION_SECRET may affect Ansible configuration so its use is not allowed in credentials.
```

Which makes sense, as ANSIBLE_* env-vars are "traditionally" reserved for ansible-core configurations and not for third party collections. Judging from several other third party collections, the usual naming scheme seems to be just a prefix of the managed product/namespace, e.g. "CHECKMK_". For a transition period I would suggest adding env-vars with a compatible naming schema (e.g. CHECKMK_SERVER_URL, CHECKMK_SITE, CHECKMK_AUTOMATION_SECRET, etc) and keep the current names as fallbacks. 

Additionally it would make sense to implement the support for the env-vars in inventory plugins and modules as well. Plugins all use the same way to parse enviroment variables for parameters (via DOCUMENTATION docstring, also able to externalize that via doc_fragments), modules use the argspec fallback option with env_fallback, as for example seen here: https://github.com/ansible-collections/community.vmware/blob/main/plugins/module_utils/_argument_spec.py

**Collection Version**
v5.8.0

robin-checkmk:
Hi @penguincustodian and thanks for bringing this to our attention!
Do you think you are up to the task of changing this yourself and creating a pull request?
We are usually pretty pressed for resources and have little time to work on these quality-of-life improvements.
So feel free to get started on a PR, otherwise someone will pick this up at some point.

[...]

PR #766, opened 2025-05-09
Move module args to utils

robin-checkmk:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently, each module has to query the same module parameters over and over again.

## What is the new behavior?

This PR moves the most basic module parameters, e.g., authentication, to the collection utils.

## Other information

This PR enables work on extending environment variable support as requested in #764.


plugins/module_utils/utils.py
```diff
@@ -13,6 +13,16 @@
 from ansible_collections.checkmk.general.plugins.module_utils.types import RESULT
```
lgetwan, line 15: I'd do it a bit simpler:
```
base_argument_spec = dict(
        server_url=dict(type="str", required=True),
        site=dict(type="str", required=True),
    )
```
And then use it in the modules like this:
```
argument_spec = base_argument_spec
```

plugins/modules/site.py
```diff
@@ -300,9 +300,265 @@ def werk16722(site_config):
 
 
 def run_module():
-    # define available arguments/parameters a user can pass to the module
+    argument_spec = base_argument_spec()
+    argument_spec.update(
+        state=dict(
```
lgetwan, line 305: In that particular case, I'd like to keep the whole argument spec in plugins/module_utils/site.py.

So, we could keep this in plugins/modules/site.py:
```
    argument_spec = base_argument_spec
    argument_spec.update(site_argument_spec)
```

And in plugins/module_utils/site.py:
```
site_argument_spec = dict(
    state=dict(
...
```

lgetwan (commented):
It feels good to have this duplicate code eliminated! :-)

I added two minor comments.

PR #813, opened 2025-07-14
Add variable to configure firewall rule comment

mueller-ma:
Add a variable to configure firewall rule comment. Other comments may be in a fixed schema and "Allow Checkmk" might not match that.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Hardcoded to "Allow Checkmk"

## What is the new behavior?

- Made it variable


robin-checkmk:
@mueller-ma I am afraid, you did not see the breaking change, which is in preparation now. Does not make your PR invalid, quite contrary, but I want to merge that ahead of your change. So while it is trivial, it will need a week or so, until I can release this.

But the contribution is valued as always! :pray:

Issue #824, opened 2025-07-23
[BUG] MKP install fails on 2.4.0p8.cre


gurubert:
I have this in my site definition:

```
    mkp_packages:
      - name: 'sslcertificates'
        version: 9.2.1
        url: 'https://exchange.checkmk.com/packages/sslcertificates/1928/sslcertificates-9.2.1.mkp'
        checksum: 'md5:0bdaa05f53877a23dbc4fe65d30f0b54'
        installed: true
        enabled: true
```

Running the playbook for the checkmk.general.server role yields:

```
TASK [checkmk.general.server : Include MKP management tasks.] ****************************************************************************************************************************
included: /home/kurs/.ansible/collections/ansible_collections/checkmk/general/roles/server/tasks/mkp.yml for client01 => (item={'name': 'cephmon', 'version': '2.4.0p8', 'update_conflict_resolution': 'abort', 'state': 'started', 'admin_pw': '__omit_place_holder__ed88aad7d43ba668a3a66944aeadccc4c0c23404', 'omd_auto_restart': 'false', 'omd_config': [{'var': 'AUTOSTART', 'value': 'on'}], 'mkp_packages': [{'name': 'sslcertificates', 'version': '9.2.1', 'url': 'https://exchange.checkmk.com/packages/sslcertificates/1928/sslcertificates-9.2.1.mkp', 'checksum': 'md5:0bdaa05f53877a23dbc4fe65d30f0b54', 'installed': True, 'enabled': True}]})

TASK [checkmk.general.server : Download MKP Packages.] ***********************************************************************************************************************************
changed: [hostname] => (item={'name': 'sslcertificates', 'version': '9.2.1', 'url': 'https://exchange.checkmk.com/packages/sslcertificates/1928/sslcertificates-9.2.1.mkp', 'checksum': 'md5:0bdaa05f53877a23dbc4fe65d30f0b54', 'installed': True, 'enabled': True})

TASK [checkmk.general.server : Transfer MKP Packages.] ***********************************************************************************************************************************
skipping: [hostname] => (item={'name': 'sslcertificates', 'version': '9.2.1', 'url': 'https://exchange.checkmk.com/packages/sslcertificates/1928/sslcertificates-9.2.1.mkp', 'checksum': 'md5:0bdaa05f53877a23dbc4fe65d30f0b54', 'installed': True, 'enabled': True}) 
skipping: [hostname]

[...]

PR #827, opened 2025-07-23
Bugfix/issue824

gurubert:
## Pull request type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

When using playbooks that already use `become_user` the tasks to install MKPs do not work.

Issue Number: #824

## What is the new behavior?

Installation of MKPs works as the `mkp` call now runs as site user.

## Other information

I do not think that this special case violates Ansible's recommendation to use `become_user`. This cannot be used twice as reported in issue #824.

### DIFF
--- a/roles/server/tasks/mkp.yml
+++ b/roles/server/tasks/mkp.yml
@@ -28,9 +28,7 @@
     - manage-mkp-packages
 
 - name: "Install mkp packages."
-  become: true
-  become_user: "{{ __site.name }}"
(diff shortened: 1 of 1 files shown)


robin-checkmk:
Hi @gurubert and thanks for the contribution!
Please use the template for PRs, so we have all the relevant information easily accessible (you can copy the description from another PR and edit yours).
Additionally, I am hesitant to merge this without explanation, why your approach is better.
I actually seem to remember that your proposal is discouraged by Ansible in favor of the `become` mechanism.

PR #854, opened 2025-08-22
Add the DCD module donated by Atruvia AG

lgetwan:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

There was no DCD module

## What is the new behavior?

- There is a DCD module

## Other information
There is no PUT endpoint in the REST API for the DCD, so it's not possible to update existing DCD connectors.

### DIFF
--- a/plugins/modules/dcd.py
+++ b/plugins/modules/dcd.py
@@ -252,6 +252,15 @@ class DCDAPI(CheckmkAPI):
         if "comment" not in self.desired or self.desired["comment"] is None:
             self.desired["comment"] = ""
 
+        # Completely remove parameters that contain empty lists
+        if "connector" in self.desired and isinstance(self.desired["connector"], dict):
+            for rule in self.desired["connector"].get("creation_rules", []):
+                if len(rule.get("matching_hosts", [])) == 0:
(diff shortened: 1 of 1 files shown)


lgetwan:
Next week, we'll get the integration tests running, as well. ;-)

plugins/modules/dcd.py
```diff
@@ -0,0 +1,503 @@
...
+                                type: bool
+                                default: false
+                            host_attributes:
+                                description: Additional host attributes to set on created hosts.
+                                type: dict
+                    discover_on_creation:
+                        description: Discover services on host creation.
+                        type: bool
+                        default: true
+                    restrict_source_hosts:
```
robin-checkmk, line 73: The documentation does not state, that this option is required, but the module fails without it.
I am torn between seeing this as an error, as it is technically a valid use case to not restrict source hosts, but the collection being stricter here, might help users avoid errors.

lgetwan: I left it "optional", but have set the default to an empty list, now.

lgetwan:
We should trigger the integration tests...

PR #892, opened 2025-10-10
Add aux_tag module

nicoske:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Auxiliary tags can only be referenced in the tag_group module but cannot be managed.

## What is the new behavior?

- Added new aux_tag module to manage auxiliary tags

## Other information
This module fills a gap to manage auxiliary tags creation and management before being referenced in tag groups.

### DIFF
--- a/.github/workflows/ans-int-test-aux_tag.yaml
+++ b/.github/workflows/ans-int-test-aux_tag.yaml
@@ -14,7 +14,7 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
-name: Ansible Integration Tests for Activation Module
+name: Ansible Integration Tests for Aux Tag Module
 on:
   workflow_dispatch:
   schedule:


plugins/modules/aux_tag.py
```diff
@@ -0,0 +1,244 @@
...
+            auxtagupdate.headers["If-Match"] = result.etag
+            result = auxtagupdate.put()
+
+            time.sleep(3)
+
+        elif result.http_code == 404:
+            auxtagcreate = AuxTagCreateAPI(module)
+            result = auxtagcreate.post()
+
+            time.sleep(3)
```
lgetwan: Same here.

lgetwan (changes requested):
Thanks a lot for creating this module!
I added some comments, of which the one in the delete method might be the most crucial one.

nicoske:
I'm afraid I delete my local fork by mistake while cleaning up my repositories. Let me know if you'd prefer that I re-open a PR or if we can update this one.

PR #921, opened 2025-11-14
Bump ansible-core version

robin-checkmk:
This PR bumps the `ansible-core` versions.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

We are testing against a EOL version of `ansible-core`. while a publicly released one is not explicitely tested.

## What is the new behavior?

- EOL `ansible-core` is dropped from testing.
- In turn, the minimum required `ansible-core` version to run this collection was bumped to `2.18`. This is still not a hard technical requirement, older versions will still probably work, we just do not test them anymore.
- Stable `ansible-core` `2.20` is added to the testing matrix.
- `ansible-core` is removed from `pyproject.toml`, to unblock testing and as it never made sense there in the first place.

## Other information

ToDo: Check open PRs, so they are in line with these changes!
* #869 
* #922

### DIFF
--- a/.github/workflows/ans-int-test-activation.yaml
+++ b/.github/workflows/ans-int-test-activation.yaml
@@ -41,16 +41,19 @@ jobs:
       fail-fast: false
       matrix:
         ansible:
-          - stable-2.17
           - stable-2.18
           - stable-2.19
+          - stable-2.20
           - devel
         python:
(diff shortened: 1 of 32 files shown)


lgetwan (approved):
Looks good to me!

PR #922, opened 2025-11-15
Add aux_tag module

nicoske:
This PR replaces #892 which could not be reopened after the fork repository was recreated.

All review feedback from @lgetwan has been addressed:
- Consolidated four API classes into single AuxTagAPI class
- Added idempotency check to only update when changes detected  
- Removed time.sleep() calls from module execution
- Fixed delete endpoint to use POST on /actions/delete/invoke

## Pull request type
- [x] Feature

Auxiliary tags can only be referenced in the tag_group module but cannot be managed.

## What is the new behavior?
- Added new aux_tag module to manage auxiliary tags

## Other information
This module fills a gap to manage auxiliary tags creation and management before being referenced in tag groups.

### DIFF
--- a/.github/workflows/ans-int-test-aux_tag.yaml
+++ b/.github/workflows/ans-int-test-aux_tag.yaml
@@ -55,35 +55,35 @@ jobs:
 
     services:
       ancient_cre:
-        image: checkmk/check-mk-raw:2.2.0p45
+        image: checkmk/check-mk-raw:2.2.0p47
         ports:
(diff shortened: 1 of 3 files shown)


robin-checkmk:
Hi @nicoske and thanks for re-opening this! Can you fix the issues detected by the tests, or do you need support with that? We are happy to help. :)

lgetwan (approved):
I like it!
It works like a charm and looks nice and compact.

Issue #923, opened 2025-11-17
[BUG] Inventory plugin broken in 3.6.0


nick-pike:
**Describe the bug**
Inventory plugin throws an `unexpected keyword argument 'user'` error

**Component Name**

Component Name: checkmk.general.checkmk

**Ansible Version**

```console
$ ansible --version
ansible [core 2.17.14]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.11.2 (main, Apr 28 2025, 14:11:48) [GCC 12.2.0] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```

**Checkmk Version and Edition**

```console
2.4.0p12 (CRE)
```

**Collection Version**

```console
$ ansible-galaxy collection list
# /root/.ansible/collections/ansible_collections
Collection                               Version
---------------------------------------- -------
ansible.utils                            6.0.0  
checkmk.general                          6.3.0  

# /usr/lib/python3/dist-packages/ansible_collections
Collection                               Version
---------------------------------------- -------
amazon.aws                               8.2.1  
ansible.netcommon                        6.1.3  
ansible.posix                            1.6.2  
ansible.utils                            4.1.0  
ansible.windows                          2.5.0  
arista.eos                               9.0.0  
awx.awx                                  24.6.1 
azure.azcollection                       2.7.0  
check_point.mgmt                         5.2.3  
chocolatey.chocolatey                    1.5.3  
cisco.aci                                2.10.1 
cisco.asa                                5.0.1  
cisco.dnac                               6.25.0 
cisco.intersight                         2.0.20 
cisco.ios                                8.0.0  
cisco.iosxr                              9.0.0  
cisco.ise                                2.9.6  
cisco.meraki                             2.18.3 
cisco.mso                                2.9.0  
cisco.nxos                               8.1.0  
cisco.ucs                                1.14.0 
cloud.common                             3.0.0  
cloudscale_ch.cloud                      2.4.0  
community.aws                            8.0.0  
community.ciscosmb                       1.0.9  
community.crypto                         2.22.3 
community.digitalocean                   1.27.0 
community.dns                            3.1.0  
[...]

PR #930, opened 2025-11-19
Trigger integration and unit tests on collection API changes

robin-checkmk:
When something is changed on the collection-internal APIs, this could go undetected up until now.
And it actually happened and led to #923. 

## Pull request type

- [x] Bugfix
- [ ] Feature
- [x] Build related changes

Tests for modules would not be triggered by changes to the collection-internal API.

## What is the new behavior?

Changes to the collection-internal APIs will trigger all relevant tests.

## Other information

ToDo: Check open PRs, so they are in line with these changes!
* #869 
* #922

### DIFF
--- a/.github/workflows/ans-int-test-aux_tag.yaml
+++ b/.github/workflows/ans-int-test-aux_tag.yaml
@@ -29,8 +29,9 @@ on:
     paths:
       - '.github/workflows/ans-int-test-aux_tag.yaml'
       - 'plugins/modules/aux_tag.py'
-      - 'tests/integration/files/includes/'
-      - 'tests/integration/targets/aux_tag/'
+      - 'plugins/module_utils/**'
+      - 'tests/integration/files/includes/**'
+      - 'tests/integration/targets/aux_tag/**'
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


lgetwan (commented):
To me, it looks like adding a directory doesn't work as expected.
No integration tests were triggered, when I changed the api.py.
When adding single files in that directory instead, it worked fine:
```  push:
    paths:
      - '.github/workflows/ans-int-test-activation.yaml'
      - 'plugins/modules/activation.py'
      - 'plugins/module_utils/api.py'
      - 'plugins/module_utils/differ.py'
      - 'plugins/module_utils/logger.py'
      - 'plugins/module_utils/lookup_api.py'
      - 'plugins/module_utils/utils.py'
      - 'plugins/module_utils/types.py'
      - 'plugins/module_utils/version.py'
      - 'tests/integration/files/includes/'
      - 'tests/integration/targets/activation/'
```
Can you please double-check?

lgetwan:
Ah, it works like this:
```
  push:
    paths:
      - '.github/workflows/ans-int-test-activation.yaml'
      - 'plugins/modules/activation.py'
      - 'plugins/module_utils/**'
      - 'tests/integration/files/includes/**'
      - 'tests/integration/targets/activation/**'
```
The double asterisk also includes sub-directories, a single one only the files inside the given directory.
There are no sub-directories in  the plugins/module_utils directory, yet, but who knows what will come.

lgetwan (approved):
Looks good now.

Issue #905, opened 2025-10-31
[FEED] Rename authentication variables


robin-checkmk:
**Describe the bug**

In #852 we introduced two new variables and decided to prefix them with `api_`.
We also agreed, that we want to use the name naming scheme for all authentication related variables.
As that will be a breaking change, this issue serves as a reminder to check the impact of this change and implement it, if reasonable.

**Component Name**

Component Name: all

**Collection Version**

```console
$ ansible-galaxy collection list

```

**Additional context**

PR #941, opened 2025-11-25
Breaking: Rename authentication variables

robin-checkmk:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring
- [ ] Build related changes

Issue: #905 

The authentication variables used in the collection have intentionally diverted, as we realized good naming of these variables needs a different approach than what was done in the early days of the collection.

With #852 we introduced two new authentication-related variables: `api_auth_type` and `api_auth_cookie`. We prefixed them with `api_`, as that is what they relate to: The Checkmk REST API.

From the early days of the collection, we used `automation_user` and `automation_secret` for authentication against Checkmk. But again, these two variables define how we authenticate against the Checkmk REST API.

## What is the new behavior?

With this pull request, we re-align all authentication-related variables and make them consistently use the `api_` prefix.

### DIFF
--- a/plugins/lookup/ldap_connection.py
+++ b/plugins/lookup/ldap_connection.py
@@ -54,8 +54,8 @@ EXAMPLES = """
   vars:
     checkmk_var_server_url: "http://myserver/"
     checkmk_var_site: "mysite"
-    checkmk_var_automation_user: "myuser"
-    checkmk_var_automation_secret: "mysecret"
+    checkmk_var_api_user: "myuser"
+    checkmk_var_api_secret: "mysecret"
     checkmk_var_validate_certs: false
     attributes: "{{ lookup('checkmk.general.ldap_connection', 'my_ldap_connection') }}"
 """


plugins/modules/dcd.py
```diff
@@ -114,8 +114,8 @@
     server_url: "http://myserver/"
     site: "mysite"
     automation_auth_type: "bearer"
```
kumaran-repo: Inconsistent naming - should be `api_auth_type` instead of `automation_auth_type` to match the api_ prefix convention used in this PR.

plugins/modules/dcd.py
```diff
@@ -140,8 +140,8 @@
...
```
kumaran-repo: Inconsistent naming - should be `api_auth_type` instead of `automation_auth_type` to match the api_ prefix convention used in this PR.

plugins/modules/ldap.py
```diff
@@ -527,8 +533,8 @@
     server_url: "http://myserver/"
     site: "mysite"
     automation_auth_type: "bearer"
```
kumaran-repo: Inconsistent naming - should be `api_auth_type` instead of `automation_auth_type` to match the api_ prefix convention used in this PR.

plugins/lookup/ldap_connection.py
adam-cmk: Line `#51`: This should use api_user and secret

lgetwan (approved):
Perfect!

PR #953, opened 2025-12-04
Fix site module test ports to follow a proper concept

robin-checkmk:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

The ports for the `site` module tests were following concepts of a plan, but that got messed up quickly.

## What is the new behavior?

A defined guideline is implemented and helps choosing proper ports for the future.

### DIFF
--- a/tests/integration/targets/site/vars/main.yml
+++ b/tests/integration/targets/site/vars/main.yml
@@ -26,7 +26,7 @@ checkmk_var_test_sites:
               use_livestatus_daemon: "with_proxy"
               global_settings: false
               tcp:
-                port: 6663
+                port: 4013
                 only_from: []
                 tls: true
               params:
--- a/tests/integration/files/includes/tasks/prep.yml
+++ b/tests/integration/files/includes/tasks/prep.yml
@@ -38,9 +38,25 @@
     creates: "/omd/sites/{{ item.site }}"
   loop: "{{ checkmk_var_test_sites }}"
 
-- name: "Configure Sites."  # noqa no-changed-when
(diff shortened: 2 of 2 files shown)


lgetwan (changes requested):
I love having such a concept!

tests/integration/targets/site/vars/main.yml
```diff
@@ -1,24 +1,32 @@
...
+              port: 6013
               encrypted: true
               host: "localhost"
               verify: false
             proxy:
               use_livestatus_daemon: "with_proxy"
               global_settings: false
               tcp:
-                port: 6664
+                port: 6663
```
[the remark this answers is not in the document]
robin-checkmk: Good catch! I fixed it, thanks!

tests/integration/targets/site/vars/main.yml
```diff
@@ -1,24 +1,32 @@
...
+              port: 6013
               encrypted: true
               host: "localhost"
               verify: false
             proxy:
               use_livestatus_daemon: "with_proxy"
               global_settings: false
               tcp:
-                port: 6664
+                port: 4013
```
lgetwan: The site module will indeed create a livestatus proxy config for the remotesite that is also accessible on this port on the central site.
But we're not using/testing the connection to this port, so it can be a random number. I still like the fact of having it well-defined.

robin-checkmk: Works for me. We could probably add a test, but I cannot see much value in it right now. But configuring it makes a lot of sense for sure.

PR #974, opened 2025-12-31
Improve environment variable testing with improved isolation

robin-checkmk:

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

There is a somewhat awkward hack in place to test lookup modules with environment variables.

The new approach is still a compromise, but it is more sturdy and more resilient than before.


lgetwan (approved):
That looks obvious to me, and seeing the action results, it seems to work well.

PR #977, opened 2026-01-02
Simplify integration tests for lookup plugins

robin-checkmk:

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

There are several lookup plugins who exist twice for good reason.
Currently, we have integration tests for every single one of them.
To me, it makes sense to simplify this and have one integration test per set.
This reduces overall complexity and overhead significantly, while the risk of having flaky tests or false positives is rather small.

## What is the new behavior?
Integration tests now are combined for lookup module "pairs".
While merging the existing tests, improvements to the overall test logic were implemented as well.
The testing scope should at least be the same as before, potentially better.

## Other information
Once this has been merged into `main`, we can also remove duplicate labels. Check the changed files for the auto labelers.

ToDo: Can we improve the tests `lookup_rules` and `lookup_rulesets` as well, or are they fine as they are?

### DIFF
--- a/.github/labels-issues.yml
+++ b/.github/labels-issues.yml
@@ -70,38 +70,26 @@ lookup:bakery:
 
 lookup:folder:
   - 'Component Name: lookup_folder'
-
-lookup:folders:
   - 'Component Name: lookup_folders'
 
 lookup:host:
   - 'Component Name: lookup_host'
-
-lookup:hosts:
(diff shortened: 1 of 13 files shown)


lgetwan (changes requested):
I like the idea of simplifying these!

Added two comments about missing tests.

PR #1015, opened 2026-02-13
LDAP module: Now compatible with Checkmk 2.5.0

lgetwan:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The LDAP module wasn't able to handle the user parameter "main_menu_icons" that was named "mega_menu_icons" in versions older than 2.5.0.

## What is the new behavior?

- The LDAP module is now correctly handling the new parameter and allows the usage of the old name as an alias.


robin-checkmk (approved):
Looks good to me apart from the typo. :)

PR #1035, opened 2026-03-06
Simplify CI

robin-checkmk:

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

Integration Tests combine several Checkmk version in a single run.
This does not only create unnecessary load on the runners, but makes it hard to read results as well.

- Integration tests were completely redesigned and are now both more accessible and easier on the runners.
- Site module - Set default for ``configuration_connection.is_trusted`` to ``false``.
    This is also the default in Checkmk, but was not reflected in the collection yet.
    On rare occasion, this could cause failures. This is fixed now.
    Double-check your configuration and refer to the official user guide for details on the option.
- Bakery module - Checkmk 2.5 uses UUIDs instead of simple integers for key IDs.
    While this change is backwards-compatible, the collection needs to support both formats.
    With this change, both integer and UUID key IDs can be used.
- LDAP module - Fix usage of ``mega_menu_icons`` and ``main_menu_icons`` for Checkmk 2.5.


lgetwan (approved):
I have mainly checked the files below the plugins folder. Looks good to me!

Issue #1016, opened 2026-02-13
[FEATURE] Create host folder if it does not exist


gurubert:
**Describe the festure**
The variable `checkmk_agent_folder` contains a folder (path) where a new host should be created.
Currently this fails when the folder does not exist.

It would be very nice if the agent role would just create folders (without any special settings).

**Component Name**
Component Name: `agent`

**Ansible Version**

```console
$ ansible --version
ansible [core 2.20.2]
  config file = /home/anarchy/r.sander/git/kunden/bbaw/ceph-deployment/ansible.cfg
  configured module search path = ['/home/r.sander/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/anarchy/r.sander/venv-ansible/lib/python3.13/site-packages/ansible
  ansible collection location = /home/r.sander/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/anarchy/r.sander/venv-ansible/bin/ansible
  python version = 3.13.5 (main, Jun 25 2025, 18:55:22) [GCC 14.2.0] (/home/anarchy/r.sander/venv-ansible/bin/python3)
  jinja version = 3.1.6
  pyyaml version = 6.0.3 (with libyaml v0.2.5)

```

**Collection Version**
```console
$ ansible-galaxy collection list
Collection                               Version
---------------------------------------- -------
ansible.posix                            2.1.0  
ansible.utils                            6.0.1  
ansible.windows                          3.3.0  
checkmk.general                          7.1.0  
community.general                        12.3.0 
```

lgetwan:
I'd say: 
The ansible.builtin.copy module can't create directories, either.
But the ansible.builtin.file module can create directories including their parents.

Thus, the checkmk.general.folder module should be able to create a folder _and_ its parents.

lgetwan:
Check out the Draft PR #1081

PR #1081, opened 2026-04-20
Folder module: recursively create parent folders.

lgetwan:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

When creating a folder, the parent folders have to exist (see #1016).

## What is the new behavior?

- The parent folders will automatically be created.
- The titles of the parent folder(s) will be taken from their path. Please provide them in lowercase, with no special characters except "_".

### DIFF
--- a/plugins/modules/folder.py
+++ b/plugins/modules/folder.py
@@ -513,7 +513,7 @@ class FolderAPI(CheckmkAPI):
             logger.debug("-> missing")
             grandparent, parent = self._normalize_path(parent)
             result = self._ensure_parent_exists(grandparent)
-            if result.http_code != 200:
+            if not self.module.check_mode and result.http_code != 200:
                 return result
 
             logger.debug("Creating parent folder %s" % parent)
@@ -635,7 +635,6 @@ def run_module():
     module = AnsibleModule(argument_spec=argument_spec, supports_check_mode=True)
 
     logger.set_loglevel(module._verbosity)
-    logger.set_loglevel(2)
 
     _exit_if_missing_pathlib(module)
 
--- a/plugins/modules/folder.py
+++ b/plugins/modules/folder.py
@@ -513,7 +513,7 @@ class FolderAPI(CheckmkAPI):
             logger.debug("-> missing")
             grandparent, parent = self._normalize_path(parent)
             result = self._ensure_parent_exists(grandparent)
-            if not self.module.check_mode and result.http_code != 200:
+            if result.http_code != 200:
                 return result
 
             logger.debug("Creating parent folder %s" % parent)
@@ -537,7 +537,7 @@ class FolderAPI(CheckmkAPI):
         logger.debug("Will create the folder")
         data = self.desired.copy()
         result = self._ensure_parent_exists(data.get("parent"))
-        if result.http_code != 200:
+        if not self.module.check_mode and result.http_code != 200:
             exit_module(
                 self.module,
                 msg="Failed to create parent folder(s).",
--- a/plugins/modules/folder.py
+++ b/plugins/modules/folder.py
@@ -30,7 +30,7 @@ options:
             - The full path to the folder you want to manage.
               Pay attention to the leading C(/) and avoid trailing C(/).
               Special characters apart from C(_) are not allowed!
-              Be aware, that the parent folder has to to exist.
+              Missing parent folders will be created.
         required: true
         type: str
     name:
@@ -105,7 +105,7 @@ EXAMPLES = r"""
     name: "My Folder"
     state: "present"
 
-- name: "Create a nested folder."  # Be advised, that the parent folder must exist
+- name: "Create a nested folder."
   checkmk.general.folder:
     server_url: "https://myserver/"
     site: "mysite"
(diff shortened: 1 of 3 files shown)


plugins/modules/folder.py
```diff
@@ -485,8 +497,55 @@ def _check_output(self, mode):
...
+        )
+
+        if result.http_code == 200:
+            logger.debug("-> exists")
+            return result
+        else:
+            logger.debug("-> missing")
+            grandparent, parent = self._normalize_path(parent)
+            result = self._ensure_parent_exists(grandparent)
+            if result.http_code != 200:
```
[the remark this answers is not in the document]
lgetwan: The GET on the "/" folder returns HTTP 200, so we're safe.

tests/integration/targets/folder/tasks/test.yml
```diff
@@ -140,6 +140,27 @@
...
+- name: "{{ outer_item.version }}.{{ outer_item.edition }} - Create folders recursively."
+  folder:
+    server_url: "{{ checkmk_var_server_url }}"
+    site: "{{ outer_item.site }}"
+    api_user: "{{ checkmk_var_api_user }}"
+    api_secret: "{{ checkmk_var_api_secret }}"
+    path: "{{ item.path }}"
+    name: "{{ item.name | default(omit) }}"
+    state: "present"
+  loop: "{{ checkmk_var_folders_recursive }}"
```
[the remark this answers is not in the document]
lgetwan: We want to keep the tests simple in the first step.

plugins/modules/folder.py
```diff
@@ -597,7 +664,7 @@ def run_module():
         if desired_state in ("present"):
             result = current_folder.create()
 
-    module.exit_json(**result_as_dict(result))
+    exit_module(module, result=result, logger=logger)
```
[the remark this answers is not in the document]
lgetwan: Updated the RETURN documentation.

tests/integration/targets/folder/vars/main.yml
```diff
@@ -20,6 +20,11 @@ checkmk_var_folders:
   - path: /foo/bar1
     name: Digital
 
+checkmk_var_folders_recursive:
+  - path: /the/folder/should/be/inside/this
+    name: Finally
+    top: /the
```
robin-checkmk, line 26: I fail to understand the idea behind this `top` thingy and the accompanying test.
I have ideas, but would like to understand your intention.

plugins/modules/folder.py
```diff
@@ -485,8 +497,57 @@ def _check_output(self, mode):
...
+                return result
+
+            # Create folder
+            logger.info("Creating parent folder %s" % parent)
+
+            data = {
+                "name": parent,
+                "title": parent,
+                "parent": grandparent,
+            }
```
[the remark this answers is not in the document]
robin-checkmk: @lgetwan is this still relevant?

Maybe slightly off-topic: I also considered, how the names of parent folders should be derived.
In my testing I saw them just being extracted from the path. Are there any safeguards?
And would it be possible to have a naming template? Like `title: $PATH.firstLetterCapital`.
It is probably too complex and if users require it, they should just create the parents directly.
I am just curious, what you think.

I wanted to add my personal preference, that by default the `title` should be the `path` element but with the first letter capitalized. Not sure if that is, what the majority would like, or if it is better to keep it simple. Probably the letter. But if the two of us like one approach better, I see no harm in following it.

plugins/modules/folder.py
```diff
@@ -30,7 +30,7 @@
             - The full path to the folder you want to manage.
               Pay attention to the leading C(/) and avoid trailing C(/).
               Special characters apart from C(_) are not allowed!
-              Be aware, that the parent folder has to to exist.
+              Missing parent folders will be created.
```
robin-checkmk: We should probably note here, how the parent folders are handled (see discussion below).

robin-checkmk (commented):
Looks good overall, my tests were successful so far. But a few things to discuss (see detailed comments).

tests/integration/targets/folder/vars/main.yml
lgetwan: That was the easiest way to delete the complete tree after creating the complete tree recursively.

plugins/modules/folder.py
lgetwan: I would keep it simple. Recursive folder creation is just a comfort feature, and regarding the title of the folders, users have to either compromise or create the folders with separate tasks or in a loop (which from a performance perspective should be similar than with our recursive creation).

And yes, I should write that into the DOCUMENTATION part. :+1:

lgetwan: The alternative would be a breaking change: The user could use the titles in the path parameter, e.g. "MUC/Infrastructure/Cisco Switches", and we let Checkmk determine the path ("muc/infrastructure/cisco_seitches"), just like in the UI.

tests/integration/targets/folder/vars/main.yml
robin-checkmk: Oh. I overlooked the `state: absent`. :facepalm: 
Can you update the task title then?

plugins/modules/folder.py
robin-checkmk: Yeah, no I definitely dislike that idea.

Let's keep it simple, as we both agree that that is the best way forward.

plugins/modules/folder.py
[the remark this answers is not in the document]
robin-checkmk: See discussion below.

robin-checkmk (approved):
Looks good to me now. 👍 
Thanks for the feature! :)

PR #1089, opened 2026-05-04
Make the `discovery` module ready for Checkmk 3.0

robin-checkmk:
Enable the `discovery` module to interact with the upconfing Checkmk 2.6.

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The `discovery` module fails, when running against Checkmk 3.0

The module now supports the upcoming Checkmk 3.0.

### DIFF
--- a/tests/integration/targets/discovery/tasks/test.yml
+++ b/tests/integration/targets/discovery/tasks/test.yml
@@ -70,7 +70,7 @@
       loop: "{{ checkmk_var_hosts }}"
       register: __checkmk_var_result_labels
       failed_when: "'State only_service_labels is not supported with this Checkmk version' not in __checkmk_var_result_labels.msg"
-      when: "not '2.3.0' in outer_item.version and not '2.4.0' in outer_item.version and not '2.5.0' in outer_item.version"
+      when: "(outer_item.version | regex_replace('p.*', '')) is version('2.3.0', '<')"
 
     - name: "{{ outer_item.version }}.{{ outer_item.edition }} - Monitor undecided services. (Should fail < 2.3.0)"
       discovery:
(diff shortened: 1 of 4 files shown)


robin-checkmk:
@lgetwan ready for review. :)

Here is the diff of the modules:
```bash
$ diff plugins/module_utils/discovery_250.py plugins/module_utils/discovery_300.py 
67,68c67,68
<     "min": "2.5.0",
<     "max": "2.5.0p99",
---
>     "min": "2.6.0",
>     "max": "3.0.0p99",
170c170
< class Discovery250(Discovery):
---
> class Discovery300(Discovery):
223c223
<                 # For single mode, there's a forwarding, but _fetch_url() doesn't support that.
---
>                 # For single mode, there's a forwarding, but _fetch() doesn't support that.

```

I'll create a follow-up PR for the one thing that changed in Checkmk 3.0 so far: #1103

relrod:
@robin-checkmk I'd advocate merging this. I have not reviewed it extensively, but the current behavior is that the collection cannot be used against master since the version bump has already happened there. :)

tests/integration/targets/discovery/tasks/test.yml
```diff
@@ -70,7 +70,7 @@
       loop: "{{ checkmk_var_hosts }}"
       register: __checkmk_var_result_labels
       failed_when: "'State only_service_labels is not supported with this Checkmk version' not in __checkmk_var_result_labels.msg"
-      when: "not '2.3.0' in outer_item.version and not '2.4.0' in outer_item.version and not '2.5.0' in outer_item.version"
+      when: "(outer_item.version | regex_replace('p.*', '')) is version('2.3.0', '<')"
```
lgetwan, line 73: Shouldn't this be 
`when: "(outer_item.version | regex_replace('p.*', '')) is version('3.0.0', '<')"`

lgetwan (commented):
I'm not sure what version('2.3.0', '<') does, but maybe that should be '3.0.0'?

tests/integration/targets/discovery/tasks/test.yml
robin-checkmk: Nope, because I inverted the logic: Instead of listing all newer versions, we now basically match on "older than". It is more future-proof and less maintenance-intensive. Does that make sense?

lgetwan: Now that I read the complete Task -- yes. ;-)
:picardfacepalm

lgetwan (approved):
Looks good to me!

PR #1091, opened 2026-05-11
Enable cleaner server role output and fix secret masking

robin-checkmk:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Due to security concerns, a contributor introduced censoring task output, so passwords would not be printed to the terminal or logs. This mechanism made it hard to read the output of some tasks and ultimately was flawed to begin with.

This PR explores a different way of solving the issue, while producing cleaner task output.
It does so by removing the blacklist approach and actively filtering for the values we want to print (namely the `name` attribute of sites and MKPs).

Additionally, the conditionals involved were improved and gaps were filled, where secrets could have been revealed.

### DIFF
--- a/changelogs/fragments/server.yml
+++ b/changelogs/fragments/server.yml
@@ -4,6 +4,14 @@ minor_changes:
     ``omd_config`` defined was started first, only to be stopped again
     immediately to apply the configuration. Now the configuration is applied
     while the site is still stopped and the site is started only once.
+  - Server role - Use the site or package name as the loop label in the site
+    and MKP management tasks for cleaner output. The ``no_log`` logic was
+    also improved to be as little obstructive as possible.
 bugfixes:
   - Server role - Only update the site admin password when it actually
     differs from the configured one.
+security_fixes:
+  - Server role - The MKP management tasks no longer show the whole package
+    dictionary, including a potentially configured ``download_password``,
+    as the loop label. Additionally, the MKP download task now honors
+    ``checkmk_server_no_log`` when a ``download_password`` is configured.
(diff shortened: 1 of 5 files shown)


robin-checkmk:
@Bbo80 I hope you do not mind me pinging you directly, but as I am touching your code, the least I can do is to check in. :)

I realized a while back, that with the security-related fix you build, the security issue was fixed, but the role output also became somewhat unreadable. This PR aims to solve that. It switches from a blacklist approach (forbid the `admin_pw` filed to be printed), to a whitelist approach (only print the site name/MKP name). I am not yet certain, how to handle the `no_log` part, but want to get your feedback early on. What do you think?

kumaran-repo (approved):
Overall looks good, legit security fix, no blocker.

One small nit (non-blocking): lines 183 and 206 in sites.yml still use the old blanket no_log: "{{ checkmk_server_no_log | bool }}", while the rest of the PR moved to the more scoped item.admin_pw is defined and checkmk_server_no_log | bool pattern. Currently harmless since the when: on those tasks already restricts to admin_pw is defined, but for consistency (and in case that condition ever changes) it might be worth aligning these two with the new pattern.

robin-checkmk:
Thank you for the review!

> One small nit (non-blocking): lines 183 and 206 in sites.yml still use the old blanket no_log: "{{ checkmk_server_no_log | bool }}", while the rest of the PR moved to the more scoped item.admin_pw is defined and checkmk_server_no_log | bool pattern. Currently harmless since the when: on those tasks already restricts to admin_pw is defined, but for consistency (and in case that condition ever changes) it might be worth aligning these two with the new pattern.

This was done intentionally, as the two mentioned tasks always deal with a password and embed it in a shell command, but also will not run, if no `admin_pw` is defined. 

So while the nit is a good and sensible finding, it is intentional.

PR #1098, opened 2026-05-18
Bugfix  default auth_type for user module

robin-checkmk:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Since 7.1.0, the parameter `auth_type` defaulted to `password`, so every edit of an existing user silently reset LDAP- or SAML-authenticated users to local authentication.

The default has been removed and the authentication option is now only changed when `auth_type`, `password` or ``enforce_password_change`` is set explicitly.

**This is a breaking change and needs a dedicated release.**


lgetwan (approved):
Looks good at first sight. If the tests are successful, go for it.

PR #1114, opened 2026-06-08
Feature rpm gpg fix for rhel 10

scrat14:
Red Hat Enterprise Linux 10 and derivates (AlmaLinux, Rocky Linux) use PQC GPG keys, which causes Ansible module rpm_key to fail in Ansible version < 2.16.18 - see https://github.com/ansible/ansible/blob/stable-2.16/changelogs/CHANGELOG-v2.16.rst and https://github.com/ansible/ansible/issues/86157.

It is fixed in Ansible 2.16.18, but sadly RHEL, AlmaLinux and Rocky Linux only ship 2.16.16, so the fix from Ansible team is not available.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When installing Ansible from RHEL/Alma/Rocky sources and run Checkmk server role, it fails with the following error message:

```
fatal: [cmk-server]: FAILED! => {"changed": false, "msg": "gpg: WARNING: no command supplied.  Trying to guess what you mean ...\ngpg: packet(6) with unknown version 6\n"}
```

## What is the new behavior?

This fix adds a rescue section to use rpm --import as a workaround for EL 10 systems until they receive the latest Ansible 2.16 packages.


robin-checkmk:
> It is fixed in Ansible 2.16.18, but sadly RHEL, AlmaLinux and Rocky Linux only ship 2.16.16, so the fix from Ansible team is not available.

This is the operational sentence in this discussion: There is a bug, it was fixed upstream, but RedHat did not update the package to this day. So not only are we discussing an end of life software version, but there is a fix, which was not adopted downstream.

So while I very much appreciate any contribution, I fail to see how introducing more complex code here is more than a workaround for an edge case. Especially given, that the recommendation on how to install Ansible is virtually anything but the OS packages. Personally I prefer `uv` these days, but really any Python package manager such as `pip` works. Those give you up-to-date Ansible, which this collection tests against. And we just cannot take the burdon of supporting EOL Ansible (or Python) versions, we have to keep the supported configuration leen in order to be able to manage workload. And I think supporting all currently supported Ansible versions is a fair support statement.

Happy to discuss this of course!

PR #1100, opened 2026-05-19
[FEATURE] Inventory Plugin: Accept variables or env-vars for secrets

JDog1895:
> Provide a brief summary of your changes as a title in the textbox above

Inventory Plugin should use auth variables served by environment variables.
Added an Options to filter hosts by folder, search folders recursive, exclude tags from search and convert hostnames to lowercase.

Check the type of change your PR introduces:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (describe what kind of change you performed):

## What is the current behavior?
Current behavior is that checkmk inventory only uses hardcoded auth parameters from the yaml file, which is a not disired behavior when using version control systems. When using AWX/AAP to inject credentials at runtime, it doesn't work. Also the plugin retrieves every host from the server, with no possibillity to filter hosts.

Issue Number: https://github.com/Checkmk/ansible-collection-checkmk.general/issues/1090
As @robin-checkmk has suggested, I created an fork and tried to implement the changes. Used an LLM to create modifications. Code is tested succesfully in our environment.

## What is the new behavior?
Added optional parameter injection by environment variables and ansible variables. Definded most previous parameters as not required.
Added Options to filter hosts by folder, search folders recursive, exclude tags from search and convert hostnames to lowercase and add an option to map a suffix to the hostnames

-
-
-

PR #1117, opened 2026-06-11
Improve Inventory Module - implements #1100

robin-checkmk:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The `inventory` module only supports grouping hosts into site and host tag based groups.

- Groups can also be based on folders (both direct and recursive)
- Hosts can be excluded based on host tags
- Hostnames can be forced to be lowercase and a domain can be appended based on host tags


lgetwan (approved):
The inventory module itself looks quite complex, so I didn't check it's functionality from a global view, but the code that was added/changed looks fine to me.

PR #1129, opened 2026-06-26
Release 8.0.0

robin-checkmk:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Since 7.1.0, the parameter ``auth_type`` defaulted to ``password``, so every edit of an existing user silently reset LDAP- or SAML-authenticated users to local authentication.

- User module - Fix editing of users that authenticate via LDAP or SAML.
    The default has been removed and the authentication option is now only changed when
    ``auth_type``, ``password`` or ``enforce_password_change`` is set explicitly.
- User module - The parameter ``auth_type`` no longer defaults to ``password``.
    Playbooks that create a new user and rely on this default must now set ``auth_type`` explicitly.
    Existing playbooks that already pass ``auth_type`` are unaffected.


adam-cmk:
This makes sense to me. Having the authentication type reset every time is definitely not something I would want.

PR #1133, opened 2026-07-02
Reworked and extended downtime module.

### LINKED ISSUE #672: [FEED] Downtime cannot be changed
**Describe the bug**
When patching servers with ansible we set a quite long downtime in Checkmk. After the patching is done we would like to reduce the downtime window instead of removing it for auditing purpose.

Trying to do this with the ansible collection does not work: setting the endtime via the downtime module does not change the downtime.
 
**Component Name**
Component Name: downtime

**Ansible Version**
This is from a test wsl installation in Ubuntu 22.04. The issue is also present for different ansible versions
```console
$ ansible --version
ansible 2.10.8
  config file = None
  configured module search path = ['/home/stefan/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.10.12 (main, Sep 11 2024, 15:47:36) [GCC 11.4.0]
```

**Checkmk Version and Edition**

```console
Checkmk Raw Edition 2.3.0p18
```

**Collection Version**
```console
checkmk.general:5.3.0
```

**To Reproduce**
1. Set downtime
2. Try to change end_after or end_time
3. See that downtime has not changed

**Expected behavior**
When a new endtime is set it should be changed

**Actual behavior**
Nothing changes

**Minimum reproduction example**
```yaml
---
- name: "Test Downtimes"
  hosts: all
  connection: local
  gather_facts: false
  vars:
    checkmk_var_server_url: "http://localhost:8080"
    checkmk_var_automation_user: "cmkadmin"
    checkmk_var_automation_secret: "..."
    mysite: "cmk"
  tasks:
    - name: "Set Downtime"
      checkmk.general.downtime:
        comment: "Downtime to be changed"
        server_url: "{{ checkmk_var_server_url }}"
        site: "{{ mysite }}"
        automation_user: "{{ checkmk_var_automation_user }}"
        automation_secret: "{{ checkmk_var_automation_secret }}"
        host_name: "{{ inventory_hostname }}"
        end_after:
          hours: 2

    - name: "Change Downtime"
      checkmk.general.downtime:
        comment: "Downtime to be changed"
        server_url: "{{ checkmk_var_server_url }}"
        site: "{{ mysite }}"
        automation_user: "{{ checkmk_var_automation_user }}"
        automation_secret: "{{ checkmk_var_automation_secret }}"
        host_name: "{{ inventory_hostname }}"
        end_after:
          minutes: 1
```

Running this playbook already shows the second task does nothing:
```console
$ ansible-playbook -i 'demo.example.com,' test-downtimes.yml

PLAY [Test Downtimes] **************************************************************************************

TASK [Set Downtime] ****************************************************************************************
changed: [demo.example.com]

TASK [Change Downtime] *************************************************************************************
ok: [demo.example.com]

PLAY RECAP *************************************************************************************************
demo.example.com           : ok=2    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

lgetwan:
## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

The previous downtime module is not very powerful and doesn't support changing existing downtimes.

#672

## What is the new behavior?

- Idempotet
- Changing downtimes is now possible
- Lookup modules available


lgetwan:
Hi @ogaida,

I have finally rewritten the downtime module after postponing this for almost two years.
I involved Claude to implement some particular features and went through all the tests again and again until everything was green. The tests of the previous module version were also green.

It's now possible to change existing downtimes. We also uncovered some special behavior when the Nagios core is involved.

As you wrote the original module, I'd like to ask you if you could test the new module in your environment and provide the results. It's important to me that it provides all the original features that you have implemented back then.
Would that be okay?

Please take your time, as I'll be on vacation in the upcoming weeks.

Best regards
Lars

plugins/modules/downtime.py
```diff
@@ -13,95 +14,188 @@
 
 short_description: Manage downtimes in Checkmk
 
-# If this is part of a collection, you need to use semantic versioning,
-# i.e. the version is of the form "2.5.0" and not "2.4".
-version_added: "0.2.0"
+version_added: "6.7.0"
```
robin-checkmk: I am not certain if even a rewrite justifies changing this value. As from a user perspective, the module is there since collection version `0.2.0` I would keep it as is. But we can lose the comment above.

plugins/modules/downtime.py
```diff
@@ -1,8 +1,9 @@
 #!/usr/bin/python
 # -*- encoding: utf-8; py-indent-offset: 4 -*-
 
-# Copyright: (c) 2022, Oliver Gaida <ogaida@t-online.de>
+# Copyright: (c) 2025, Lars Getwan <lars.getwan@checkmk.com>
```
robin-checkmk: Maybe we can combite the copyright note like this?
`Copyright: (c) 2022, Oliver Gaida <ogaida@t-online.de> & 2025, Lars Getwan <lars.getwan@checkmk.com>`
We did that in e.g., the `user` module for co-authorship.

plugins/modules/downtime.py
```diff
@@ -13,95 +14,188 @@
...
 
 description:
-    - Manage downtimes within Checkmk.
+    - Create, update and delete host and service downtimes in Checkmk.
+    - An existing downtime can be updated (e.g. to shorten or extend its end time)
+      without deleting and recreating it. The downtime to act on can be identified
+      by its ID, by host name (and optionally service descriptions), or by a
+      Livestatus query.
+    - This module is idempotent. It only changes downtimes when the desired end
+      time or comment differs from the current state.
```
robin-checkmk: Claude tends to do this, but I would consider this comment redundant. We always strive for idempotence and usually only document, if it cannot be achieved.

plugins/modules/downtime.py
```diff
@@ -13,95 +14,188 @@
...
+        choices: ["host", "service"]
     comment:
         description:
-            - Remarks for the downtime. If omitted in combination with state = present, the
-              default 'Set by Ansible' will be used, in combination with state = absent, ALL downtimes of
-              a host or host/service will be removed.
+            - The comment of the downtime.
+            - When creating or matching a downtime by I(host_name), the comment is
+              part of the identity of the downtime. If omitted, C(Managed by Ansible)
+              is used.
```
robin-checkmk, line 93: Does it make sense to have a default value for the `comment`?
From the rest of the documentation it sounds like a comment is ususally not required.

plugins/modules/downtime.py
```diff
@@ -13,95 +14,188 @@
...
+        description:
+            - The host to schedule, update or delete a downtime for.
+            - Mutually exclusive with I(downtime_id) and I(query).
+        required: false
+        type: str
+    service_descriptions:
+        description:
+            - A list of service descriptions.
+            - If set together with I(host_name), the module acts on service
+              downtimes for these services. If omitted, it acts on host downtimes.
```
robin-checkmk: I struggle to understand this part (probably just a documentation issue).
So if I set this, I also need to set `host_name`, otherwise nothing happens, right?
Maybe we can clarify this to avoid confusion.

plugins/modules/downtime.py
```diff
@@ -13,95 +14,188 @@
...
+        description:
+            - A Livestatus query as a JSON string, written in terms of the
+              Livestatus C(downtimes) table (e.g. C(host_name),
+              C(service_description), C(comment)). See the Checkmk REST API
+              documentation for the query syntax.
+            - The same query columns are used for B(create), B(update) and
+              B(delete). For query-based B(create) the module automatically
+              translates the columns to the queried object's table (C(host_name)
+              becomes C(name) for hosts, C(service_description) becomes
+              C(description) for services), so one query works for all operations.
```
robin-checkmk: I feel like this goes a bit too deep into implementation details, I think we can remove it.

plugins/modules/downtime.py
```diff
@@ -13,95 +14,188 @@
...
         type: dict
         default: {}
-    start_time:
+    duration:
         description:
-            - The start datetime of the downtime. The format has to conform to the ISO 8601 profile I(e.g. 2017-07-21T17:32:28Z).
-              The built-in default is now.
+            - Duration in minutes. When set, the downtime does not begin
+              automatically at a nominated time, but when a non-OK status actually
+              appears for the host (flexible downtime).
```
robin-checkmk, line 138: I wante to suggest to link to the official guide here, but the explanation about flexible downtimes there is rather scarse as well. So maybe we just mention "Refer to the offical user guide" here? Or keep as is, that would be fine for me too.

plugins/modules/downtime.py
```diff
@@ -13,95 +14,188 @@
...
+            - Only relevant when creating a downtime.
+        required: false
+        type: int
+        default: 0
+    recur:
```
robin-checkmk: If there is not good reason to use this abbraviated `recur`, maybe we can use `recurring`?

plugins/modules/downtime.py
```diff
@@ -13,95 +14,188 @@
...
-      If this combination already exists as a downtime, the new downtime will not be created except using the B(force) argument.
-      The creation of service downtimes works accordingly, with hostname, service description and comment.
+    - Creating a downtime is possible via I(host_name) or via I(query) (using
+      I(downtime_type) to choose host or service). Updating and deleting can be
+      done via I(downtime_id), I(host_name) or I(query).
+    - Idempotency is based on the current end time and comment of the matching
+      downtimes. Absolute times (I(end_time)) are fully idempotent. Relative times
+      (I(end_after)) are recomputed on every run and will therefore usually trigger
+      an update.
+    - On the Checkmk Raw edition (renamed I(Community) in 2.5) the Nagios core
```
robin-checkmk: Let's only use `In Checkmk Community` here. We dropped the old edition names in most places, so to avoid having to clean this up, I would lose it right now.

plugins/modules/downtime.py
```diff
@@ -13,95 +14,188 @@
...
+    - plugin: checkmk.general.downtime
+      plugin_type: lookup
+    - plugin: checkmk.general.downtimes
+      plugin_type: lookup
 
 author:
-    - Oliver Gaida (@ogaida)
-    - Lars Getwan (@lgetwan)
+    - Oliver Gaida (@ogaida) -- Original implementation
+    - Lars Getwan (@lgetwan), with the help of Claude -- Modified version
```
robin-checkmk: We do not need the comments here. We never did that before and I would rather not start. Who is to say, when and which comments would need to be made at some point. Let's keep it simple. :)

plugins/modules/downtime.py
```diff
@@ -114,198 +208,139 @@
...
-- name: "Schedule a host downtime with a comment, starting now, ending in 2 hours."
-  checkmk.general.downtime:
-    server_url: "https://myserver/"
-    site: "mysite"
-    api_user: "myuser"
-    api_secret: "mysecret"
-    host_name: "myhost"
-    comment: "Managed by Ansible"
-    end_after:
-      hours: 2
```
robin-checkmk, line 135: I did not find a replacement, is there a reason to drop this example? There are a few more below.
If we can folde these dropped examples into others, I am all for it. But it feels like they were dropped, because the focus of Claude was on other topics.

plugins/modules/downtime.py
```diff
@@ -114,198 +208,139 @@
...
-
 - name: "Schedule a host downtime using absolute start and end times."
   checkmk.general.downtime:
     server_url: "https://myserver/"
     site: "mysite"
     api_user: "myuser"
     api_secret: "mysecret"
     host_name: "myhost"
-    comment: "Managed by Ansible"
+    comment: "Patch window"
```
robin-checkmk: Is there a good reason to change this?

Going through the rest of the documentation, this was changed in several places, but I see no good reason. This might be in conjunction with my comment about the `comment` field above, maybe I am missing something.

plugins/modules/downtime.py
```diff
@@ -114,198 +208,139 @@
...
     api_user: "myuser"
     api_secret: "mysecret"
     host_name: "myhost"
-    comment: "Managed by Ansible"
-    start_after:
-      minutes: 30
+    comment: "Patch window"
+    service_descriptions:
+      - "CPU utilization"
+      - "Memory"
```
robin-checkmk: Same question about the comment as above. Here it makes the example actually worse, as a downtime on a few distinct services makes no sense for a "Patch window".

plugins/modules/downtime.py
```diff
@@ -114,198 +208,139 @@
...
-    state: "absent"
+    query: '{"op": "=", "left": "service_description", "right": "Filesystem /"}'
+    downtime_type: "service"
+    comment: "Storage migration"
+    end_after:
+      hours: 1
 
 # ---------------------------------------------------------------------------
-# Service downtimes - scheduling
+# Updating an existing downtime (solves issue #672)
```
robin-checkmk: Remove the reference to the issue please.

plugins/modules/downtime.py
```diff
@@ -114,198 +208,139 @@
...
+    query: '{"op": "=", "left": "host_name", "right": "myhost"}'
+    end_after:
+      minutes: 30
 
-- name: "Remove service downtimes matching a specific comment."
+# On the host_name path the comment is part of the downtime's identity, so a
+# different comment is treated as a different downtime and a new one is created.
+# To change the comment of an existing downtime, select it by a query (e.g. by
+# its host and current comment) and set the new comment; the query matches the
+# downtime independently of the comment you are about to write.
```
robin-checkmk, line 326: I am going to refer one more time to my above comment about downtime comments. I feel like this is hard to grasp. Maybe that is just how it is, but I would like to improve it if possible at all.

tests/integration/targets/downtime/tasks/test.yml
```diff
@@ -1,10 +1,22 @@
...
   ansible.builtin.set_fact:
     checkmk_var_customer: "{{ 'provider' if outer_item.edition in ['managed', 'ultimatemt'] else None }}"
 
+# -----------------------------------------------------------------------------
+# Setup: hosts, discovery, activation
+# -----------------------------------------------------------------------------
+
 - name: "{{ outer_item.version }}.{{ outer_item.edition }} - Create hosts."
-  host:
+  checkmk.general.host:
```
robin-checkmk: Was the switch to FQCNs done intentionally? I think I know why, but want to make sure.

tests/integration/targets/downtime/tasks/test.yml
```diff
@@ -40,284 +52,678 @@
...
-  loop: "{{ checkmk_var_hosts }}"
 
-- name: "{{ outer_item.version }}.{{ outer_item.edition }} - Schedule downtime on services with relative times."
-  downtime:
+- name: "{{ outer_item.version }}.{{ outer_item.edition }} - Wait for the core to process the queued downtime command."
+  ansible.builtin.wait_for:
+    timeout: "{{ checkmk_var_dt_settle }}"
+
+# -----------------------------------------------------------------------------
+# A. Host downtimes: create, idempotency and modify-in-place (#672)
```
robin-checkmk: I am sensing, that there are a few too many references to the issue number. I do not mind it, if a test hits the specific topic of an issue. But I think for general integration tests, the reference should not be necessary.

tests/integration/targets/downtime/tasks/test.yml
```diff
@@ -40,284 +52,678 @@
...
-    start_after:
-      hours: 2
-      minutes: 30
-    end_after:
-      hours: 1
-      minutes: 30
-    service_descriptions:
-      - "Check_MK"
-      - "Check_MK Discovery"
+    comment: "Patch window"
```
robin-checkmk: See my comments about the `comment` field in the module itself.

robin-checkmk (changes requested):
I love the overall shape of this PR.
I cannot judge the code, but the integration tests look very extensive and complete. I had a few remarks about the module documentation and some smaller comments. More on the cleanup side of things, than a redesign.

plugins/modules/downtime.py
lgetwan: I would keep the default. This field is needed to provide idempotency.
If users are unhappy with "Managed by Ansible", they still can set it to something different.

plugins/modules/downtime.py
lgetwan: I changed that. Probably, this describes it better:

>         description:
>            - A list of service descriptions.
>            - Together with I(host_name), the module acts on service downtimes
>              for these services on that particular host. If omitted, it acts on
>              host downtimes.
>            - If you want to set a downtime on a particular service for I(all)
>              hosts, you have to use the I(query) parameter.

plugins/modules/downtime.py
lgetwan: I replaced it with this, which might be more helpful for the user:

       - As the column names are different, depending on the Livestatus table,
         Please use the column names as defined in the downtimes table, e.g.
         C(service_description) instead of C(description) and C(host_name)
         instead of C(name).

plugins/modules/downtime.py
lgetwan: The parameter is named "recur" in the REST API, and in the previous version of the module, it was also named "recur".
If we rename it, we should keep "recur" as an alias. 
What do you think?

plugins/modules/downtime.py
lgetwan: Claude suggests:
>            - Length of a flexible downtime in minutes. Instead of starting at
>              the scheduled time, the downtime starts when the host or service
>              first goes non-OK within the start/end window, then lasts this long.

Better, isn't it?

plugins/modules/downtime.py
lgetwan: We have another "starting now, ending in 2 hours" example, and we also have examples with comments, so that should be fine from my point of view.

plugins/modules/downtime.py
lgetwan: Do you mean improving the comment or the behavior?

tests/integration/targets/downtime/tasks/test.yml
lgetwan: This was not by intention. I will use the short name.

plugins/modules/downtime.py
robin-checkmk: This is way better, thank you!

plugins/modules/downtime.py
robin-checkmk: Alias it is, good idea!

plugins/modules/downtime.py
robin-checkmk: Yep, definitely better than a plain "Refer to the offical user guide". 👍

plugins/modules/downtime.py
robin-checkmk: Fair enough. 👍

plugins/modules/downtime.py
robin-checkmk: Yeah, I think that is better, thanks.

plugins/modules/downtime.py
```diff
@@ -13,308 +14,333 @@
...
-              a host or host/service will be removed.
+            - The comment of the downtime.
+            - When creating or matching a downtime by I(host_name), the comment is
+              part of the identity of the downtime. If omitted, C(Managed by Ansible)
+              is used.
+            - When updating a downtime by I(downtime_id) or I(query), the comment is
+              only changed if it is explicitly set here.
+            - When deleting by I(host_name), the deletion is limited to downtimes
+              with this comment if it is set, otherwise all matching downtimes are
+              removed.
```
robin-checkmk, line 98: Does this work, if the module defaults to "Managed by Ansible"?

plugins/modules/downtime.py
robin-checkmk: I am honestly not sure. It feels pretty complicated to me, but that is not the module, but Checkmk itself.
So I guess we can only document as good as possible. On that note, we should probably mention this in the documentation in  the top of this file, right?

plugins/lookup/downtimes.py
```diff
@@ -0,0 +1,180 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+DOCUMENTATION = """
+    name: downtimes
+    author: Lars Getwan (@lgetwan)
+    version_added: "6.7.0"
```
robin-checkmk: The `version_added` is off, we need to update it prior to release. Leaving this here as a reminder.

plugins/lookup/downtime.py
```diff
@@ -0,0 +1,135 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+DOCUMENTATION = """
+    name: downtime
+    author: Lars Getwan (@lgetwan)
+    version_added: "6.7.0"
```
robin-checkmk: The `version_added` is off, we need to update it prior to release. Leaving this here as a reminder.

robin-checkmk (commented):
Looking at the lookup modules - which look pretty neat to me - we have no integration tests for those, right?

Also, is the current state rebased on `devel`, as we just landed some relevant changes there.

plugins/modules/downtime.py
lgetwan: Yes, because the default is only used when state == present.

plugins/modules/downtime.py
lgetwan: Yes, done.

plugins/lookup/downtimes.py
lgetwan: Done.

plugins/lookup/downtime.py
lgetwan: Done.

robin-checkmk (approved):

PR #1136, opened 2026-07-08
Move OMD configuration management before state management

robin-checkmk:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

- Previously, a stopped site (e.g., freshly created or just updated) with `omd_config` defined was started first, only to be stopped again immediately to apply the configuration. Now the configuration is applied while the site is still stopped and the site is started only once.


kumaran-repo (approved):
Verified the task reorder against configure-site.yml's stop/config/start logic.

PR #1139, opened 2026-07-13
Release 8.1.0

robin-checkmk:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

- Server role - Apply the OMD site configuration before starting the site.
    Previously, a stopped site (e.g. freshly created or just updated) with
    ``omd_config`` defined was started first, only to be stopped again
    immediately to apply the configuration. Now the configuration is applied
    while the site is still stopped and the site is started only once.
- Server role - Use the site or package name as the loop label in the site
    and MKP management tasks for cleaner output. The ``no_log`` logic was
    also improved to be as little obstructive as possible.
- Server role - Only update the site admin password when it actually
    differs from the configured one.
- Server role - The MKP management tasks no longer show the whole package
    dictionary, including a potentially configured ``download_password``,
    as the loop label. Additionally, the MKP download task now honors
    ``checkmk_server_no_log`` when a ``download_password`` is configured.
- Remove the unused dependency on the ``ansible.utils`` collection.


roles/server/tasks/sites.yml
```diff
@@ -139,65 +152,64 @@
...
+- name: "Update Site Admin Password for Checkmk < 2.1."
   become: true
   ansible.builtin.shell: |
     set -o pipefail
-    echo '{{ item.admin_pw }}' | htpasswd -i /omd/sites/{{ item.name }}/etc/htpasswd cmkadmin
+    if ! echo '{{ item.admin_pw }}' | htpasswd -vi /omd/sites/{{ item.name }}/etc/htpasswd cmkadmin > /dev/null 2>&1
+    then
+      echo '{{ item.admin_pw }}' | htpasswd -i /omd/sites/{{ item.name }}/etc/htpasswd cmkadmin
+      echo "Password updated."
+    fi
```
lgetwan, line 180: Cool, I like that!

lgetwan (approved):
Looks fine to me!

PR #1143, opened 2026-07-15
Improve idempotence of the rule module

robin-checkmk:
This PR was LLM-assisted. We will test it thoroughly and possibly adapt it in the process.
Human review is definitely the decisive factor here, we will only merge it, if it is well understood and valid.

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: #186

- F1: rule module - Values are compared structurally, instead of rewriting
  parentheses to brackets, which silently ignored real changes to values
  containing brackets or parentheses inside strings, e.g. a command line.
  (``d731b032``)
- F2: rule module - ``host_labels`` and ``service_labels`` are translated to
  the label group format of Checkmk 2.3 and newer before comparing and sending,
  so the old syntax no longer creates a duplicate rule on every run.
  (``38cd1259``)
- F3: rule module - A nonexistent ``rule_id`` with ``state: present`` now fails
  as documented, instead of reporting an unchanged success. (``1f4cd5fb``)
- F4: rule module - Creating a rule at the default location no longer issues a
  superfluous move. (``f50c45cf``)
- F5: rule module - The examples use a value format that is valid on 2.3, 2.4
  and 2.5, and new notes cover the canonical value format, rules containing
  secrets and rule ordering. (``0994ce25``, ``6c3b9089``)
- F6: rule module - Normalizing a rule for comparison no longer modifies the
  desired and the current state in place. The data sent to the API is
  unchanged. (``fae197a9``)
- rule module - Regression tests for F1 to F4. Each of them was confirmed to
  fail against the pre-fix code. (``71d8c9ed``)

Two judgement calls to confirm during review:

- J1: Rules containing secrets have to reference the password store. Such a
  reference is byte-stable on 2.3, 2.4 and 2.5, while an explicit password
  never is: the API masks it, and 2.3 and 2.4 regenerate its id on every read.
  Its exact form depends on the version and on the ruleset, so the notes tell
  users to copy the form their site returns.
- J2: Relative positions cannot be idempotent in general, and ``any`` is
  unsuitable for rulesets that are evaluated in first match order. Both are
  documented, together with the pattern that works: anchor the first rule and
  chain the others with ``position: after`` and ``neighbour``.

### DIFF
--- a/changelogs/fragments/rule_idempotency.yml
+++ b/changelogs/fragments/rule_idempotency.yml
@@ -1,3 +1,4 @@
 bugfixes:
   - rule module - Fail as documented when the provided ``rule_id`` does not exist, instead of reporting an unchanged success.
   - rule module - Compare rule values structurally instead of rewriting parentheses to brackets, which silently ignored real changes to values containing parentheses or brackets inside strings.
+  - rule module - Translate ``host_labels`` and ``service_labels`` conditions to the label group format used by Checkmk 2.3 and newer, preventing a new rule from being created on every run when the old syntax is used.
--- a/changelogs/fragments/rule_idempotency.yml
+++ b/changelogs/fragments/rule_idempotency.yml
@@ -1,2 +1,3 @@
 bugfixes:
   - rule module - Fail as documented when the provided ``rule_id`` does not exist, instead of reporting an unchanged success.
+  - rule module - Compare rule values structurally instead of rewriting parentheses to brackets, which silently ignored real changes to values containing parentheses or brackets inside strings.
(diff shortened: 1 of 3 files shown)


lgetwan:
I reviewed the code changes of each commit, and the look fine!
Especially F3 was a nasty bug, obviously. :-( 
F1 will fix many idempotency problems for our users, and it looks like a rather simple solution. Not sure why I haven't done it like that beforehand. 

Regarding the judgement calls:
J1: I'd prefer to insist on the usage of the password safe, as this is good practice, anyway. 
J2: Using "any" will not work for "first match" rule sets. We have to mention this in the documentation.
Maybe a combination of "bottom/top/before/after" (when the playbook is executed for the first time) and "any" (for each upcoming execution) can do the job. This doesn't sound like Ansible best practice, though.

robin-checkmk:
@lgetwan thanks for the first look! I updated the PR based on your remarks and from my end it should be good to go. 
I also keep testing it in with different use-cases and so far the operational results look promising.
Can you do a final review of the code?

lgetwan (approved):
Looks good to me!

PR #1144, opened 2026-07-17
Fix privilege escalation for MKP management

robin-checkmk:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #824 

- Remove `become_user: "{{ __site.name }}"`.
- Prefix command block with `su - {{ __site.name }} -c ` .
- This avoids the problem discussed in the aforementioned issue.
- [Background information](https://docs.ansible.com/projects/ansible-core/2.20/playbook_guide/playbooks_privilege_escalation.html#risks-of-becoming-an-unprivileged-user)

### DIFF
--- a/changelogs/fragments/mkps.yml
+++ b/changelogs/fragments/mkps.yml
@@ -1,0 +1,4 @@
+bugfixes:
+  - Server role - Fix MKP management tasks failing when connecting as an
+    unprivileged user without ACL support, by using ``su`` to switch to the
+    site user instead of relying on ``become_user``.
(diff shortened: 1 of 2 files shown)


kumaran-repo (commented):
I tested this locally. This is a minor correctness and maintainability issue rather than a current functional bug.

'-c' specifies the command to execute, not the shell. Since '-c' is provided twice, the second value overwrites the first, so `/bin/bash` is not actually used to select Bash. The command currently works because the Checkmk site user already has Bash configured as its login shell.

Please use '-s /bin/bash' to select the shell explicitly:
```
ansible.builtin.command: >-
  su - {{ __site.name }} -s /bin/bash
  -c 'mkp add {{ __checkmk_server_tmp_dir }}/{{ __mkp.name }}-{{ __mkp.version }}.mkp'
```

kumaran-repo (approved):

PR #1151, opened 2026-07-23
add UID/GID as optional parameters for OMD create in server role

pwdebruyne:
## Pull request type
- [x] Feature

Add optional parameters in the server role to specify the UID and GID (cfr. omd create commandline parameters)

## What is the new behavior?
If uid and/or gid is specified in the variable checkmk_server_sites, the uid and/or gid will be used during the creation of the site.

### DIFF
--- a/changelogs/fragments/changelog_role_sitecreate.yml
+++ b/changelogs/fragments/changelog_role_sitecreate.yml
@@ -1,4 +1,2 @@
-release_summary: "Add support in server role for UID and GID"
-
 minor_changes:
   - "Add support in server role for optional parameter OMD create UID and GID"
--- a/changelogs/fragments/changelog_role_sitecreate.yml
+++ b/changelogs/fragments/changelog_role_sitecreate.yml
@@ -1,0 +1,4 @@
+release_summary: "Add support in server role for UID and GID"
+
+minor_changes:
+  - "Add support in server role for optional parameter OMD create UID and GID"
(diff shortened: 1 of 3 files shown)


pwdebruyne:

changelogs/fragments/changelog_role_sitecreate.yml
```diff
@@ -0,0 +1,4 @@
+release_summary: "Add support in server role for UID and GID"
```
robin-checkmk: Please remove the `release_summary`, that is something we will add only once for the full release.

roles/server/meta/argument_specs.yml
```diff
@@ -88,6 +88,14 @@ argument_specs:
               value:
                 type: "str"
                 description: "The value of the variable."
+          uid:
+            type: "str"
```
robin-checkmk: I did not check the reference, but shouldn't this be an integer?

roles/server/meta/argument_specs.yml
```diff
@@ -88,6 +88,14 @@ argument_specs:
...
+            description: "The UID of the OMD site user"
+            required: false
+          gid:
+            type: "str"
```
robin-checkmk: See above comment.

robin-checkmk (changes requested):

robin-checkmk:
Thank you for your contribution! Pending my few comments, this is good for the next release. :)

pwdebruyne:
Thanks for the feedback Robin. I updated according to your comments.

robin-checkmk (approved):
Looks good to me now, thanks! I will merge this prior to the next release.
Thank you for your contribution!

PR #1162, opened 2026-08-06
Normalise a trailing slash on `server_url`.

robin-checkmk:

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

- All modules, lookup plugins and the inventory plugin - Normalize a trailing
  slash on ``server_url``. The documentation consistently showed the
  ``https://myserver/`` form, which built request URLs with a doubled slash in
  front of the site name. The Checkmk REST API tolerates that, but other
  consumers of the same variable do not.

- Lookup plugins, Inventory plugin - ``CheckMKLookupAPI`` now takes
  ``server_url`` and ``site`` separately and joins them itself, instead of
  every caller building the site URL by hand. This only affects code importing
  the class directly from ``module_utils``, not playbooks or roles.

### DIFF
--- a/tests/unit/plugins/module_utils/test_api.py
+++ b/tests/unit/plugins/module_utils/test_api.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python
+# -*- encoding: utf-8; py-indent-offset: 4 -*-
+
+# Copyright: (c) 2026, Checkmk GmbH
+# GNU General Public License v3.0+
+# (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+from unittest.mock import MagicMock
+
+import pytest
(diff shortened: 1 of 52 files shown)


tests/unit/plugins/module_utils/test_api.py
lgetwan, line 1: Cool, there's even a unit test!

tests/unit/plugins/module_utils/test_lookup_api.py
lgetwan, line 1: ... and another one. :-)

lgetwan (approved):
Looks straight forward. Thumbs up!

tests/unit/plugins/module_utils/test_api.py
robin-checkmk: That's the french touch. ;)

tests/unit/plugins/module_utils/test_lookup_api.py
robin-checkmk: Even more french touch. ;)

PR #1163, opened 2026-08-07
Accept a list of terms in the singular lookup plugins

robin-checkmk:

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

- Activation, Folder, Folders, Host, LDAP Connection, Role and Site lookup
  modules - Terms can now be passed as a single list, not only as separate
  arguments. Previously `lookup('checkmk.general.host', my_host_list)`
  failed with `can only concatenate str (not "_AnsibleLazyTemplateList")`,
  and the Activation lookup module silently built an invalid URL instead.


lgetwan (approved):
Looks good to me from a Python perspective.

PR #1166, opened 2026-08-10
fix(agent): separate download and upload paths for agent package transfer

### LINKED ISSUE #741: [BUG] checkmk_agent_delegate_download to localhost for windows target won't work
**Describe the bug**
When `checkmk_agent_delegate_download` is set to `localhost` and the target is a Windows system, the download tasks use a Windows path (e.g., `C:\Windows\Temp\check-mk-agent_2.3.0p25.msi`). This causes the download to fail, and the issue goes unnoticed because the `failed_when` condition is set to false. A potential fix would be to introduce a new setting (e.g., `checkmk_agent_delegate_download_path`) that downloads the file into a Linux-compatible directory, which is then copied to the correct Windows location.

**Component Name**
role.agent > Win32NT-files.yml > Download Tasks

**Ansible Version**
```console
ansible [core 2.16.2]
  config file = None
  configured module search path = ['/home/xclemens/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/xclemens/.local/lib/python3.10/site-packages/ansible
  ansible collection location = /home/xclemens/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/xclemens/.local/bin/ansible
  python version = 3.10.12 (main, Feb  4 2025, 14:57:36) [GCC 11.4.0] (/usr/bin/python3)
  jinja version = 3.0.3
  libyaml = True

```

**Checkmk Version and Edition**
```console
Checkmk Enterprise Edition 2.3.0p25
```

**Collection Version**

```console
$ ansible-galaxy collection list
checkmk.general                      5.6.0  
```

**To Reproduce**
1. Set checkmk_agent_delegate_download to localhost on a Linux control node.
2. Target a Windows system.
3. Run the playbook to download the agent.
4. The agent download fails because the destination path (C:\...) does not exist on Linux.

**Expected behavior**
The agent should be downloaded to a Linux-compatible temporary directory (e.g., using a new setting like checkmk_agent_delegate_download_path) and then copied to the correct Windows location.

**Actual behavior**
Nothing is downloaded because the destination path is invalid on Linux, and the error is hidden due to failed_when being set to false.

**Additional context**
Affected Lines:
[Win32NT-files.yml Lines 177C1-L225C27](https://github.com/Checkmk/ansible-collection-checkmk.general/blob/2188a6dc031a021ba9148e77877b002dcb82b41d/roles/agent/tasks/Win32NT-files.yml#L177C1-L225C27)

**Possible Solution, not ready (!) just a idea for the direction**
```yaml
- name: "{{ ansible_system }}: Download GENERIC {{ checkmk_agent_edition | upper }} Agent (delegated)."
  ansible.builtin.uri:
    url: "{{ __checkmk_agent_agent.url.cee }}?\
          os_type={{ __checkmk_agent_files_mapping[ansible_os_family] }}&\
          agent_type=generic"
    validate_certs: "{{ checkmk_agent_server_validate_certs | bool }}"
    dest: "{{ checkmk_agent_delegate_download_path }}/{{ inventory_hostname }}-check-mk-agent_{{ checkmk_agent_version }}.msi"
    method: GET
    headers:
      Authorization: "Bearer {{ checkmk_agent_user }} {{ __checkmk_agent_auth }}"
      Accept: "application/octet-stream"
  when: |
    checkmk_agent_edition | lower != "cre"
    and not (checkmk_agent_host_specific | bool or checkmk_agent_folder_specific | bool)
  retries: 3
  delegate_to: "{{ checkmk_agent_delegate_download }}"
  tags:
    - download-package

- name: "{{ ansible_system }}: Transfer host-specific {{ checkmk_agent_edition | upper }} Agent."
  ansible.windows.win_copy:
    src: "{{ checkmk_agent_delegate_download_path }}/{{ inventory_hostname }}-check-mk-agent_{{ checkmk_agent_version }}.msi"
    dest: "{{ __checkmk_agent_agent.file.host }}"
  when: |
    checkmk_agent_edition | lower != "cre"
    and checkmk_agent_host_specific | bool
  tags:
    - download-package

- name: "{{ ansible_system }}: Transfer GENERIC or folder-specific {{ checkmk_agent_edition | upper }} Agent."
  ansible.windows.win_copy:
    src: "{{ checkmk_agent_delegate_download_path }}/check-mk-agent_{{ checkmk_agent_version }}.msi"
    dest: "{{ __checkmk_agent_agent.file.cee }}"
  when: |
    checkmk_agent_edition | lower != "cre"
    and not checkmk_agent_host_specific | bool
  tags:
    - download-package

- name: "{{ ansible_system }}: Transfer Vanilla {{ checkmk_agent_edition | upper }} Agent."
  ansible.windows.win_copy:
    src: "{{ checkmk_agent_delegate_download_path }}/check-mk-agent_{{ checkmk_agent_version }}.msi"
    dest: "{{ __checkmk_agent_agent.file.cre }}"
  when: |
    checkmk_agent_edition | lower == "cre"
  tags:
    - download-package

```

aewtemp:
Fixes #741

The `win_copy` and `copy` modules always expect their source file to be located either on the Ansible controller, or on the target host itself when `remote_src: true` is set. They cannot transparently bridge a file between two different remote hosts.

Previously, a single path variable (`__checkmk_agent_agent.file.*`) was used both as the destination for the initial package download (which may run on a delegated host, e.g. `localhost`, depending on `checkmk_agent_delegate_download`) and as the source/destination for the subsequent copy task onto the actual target host. When the download was delegated to a host different from the target, this caused the copy task to fail, since the file did not exist at the expected location on the target.

This change splits the previously shared path into two explicit, independent path sets:

- `__checkmk_agent_agent.file.download.*`: where the package is downloaded to (on the delegated/download host).
- `__checkmk_agent_agent.file.upload.*`: where the package is copied/uploaded to on the target host, and subsequently installed from.

All download tasks (`get_url`/`win_get_url` via
`checkmk_agent_delegate_download`) now write to `file.download.*`, while the `copy`/`win_copy` tasks explicitly use `file.download.*` as `src` and `file.upload.*` as `dest`. Installation tasks (`package`, `zypper`, `win_package`) now consistently reference `file.upload.*`, since that is the final location on the target host.

Additionally:
- `checkmk_agent_delegate_download` is changed from a hostname-based variable (defaulting to `inventory_hostname`) to a boolean, making the delegation intent explicit rather than implicit through hostname comparison.
- ~~`__checkmk_agent_host_tmp_dir` (used for the download destination) is now defined centrally in `vars/main.yml`, while OS-specific `__checkmk_agent_lin_tmp_dir` / `__checkmk_agent_win_tmp_dir` variables are introduced for the upload destination on Linux and Windows targets respectively.~~
- `__checkmk_agent_host_tmp_dir` stays as is.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The role "agent" currently is build upon the belief that setting a target_name on "checkmk_agent_delegate_download" would delegate the download and copy transaction of the agent to that host. But the module "windows.win_copy" and "builtin.copy" expect the file to be copied either from the controller or the target host.
Issue Number: #741 

## What is the new behavior?
The changes introduced by this request allow the download to either be delegated to the controller or to the target host.
The "checkmk_agent_delegate_download" variable was changed to a boolean. If true, the download of the file will be handled by the controller. Otherwise the download will be done on the target.

-
-
-

---
<a href="https://gitkraken.dev/review/github/Checkmk/ansible-collection-checkmk.general/pull/1166?source=pr_review_chip">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://gitkraken.dev/images/figures/gitkraken-review-badge-dark.svg">
    Open with GitKraken
  </picture>
</a>


roles/agent/vars/main.yml
```diff
@@ -37,3 +37,7 @@ __checkmk_agent_files_mapping:
   RedHat: linux_rpm
   Suse: linux_rpm
   Windows: windows_msi
+
+# Temporary directory to download the agent to (Win-Path if controller runs on Windows, Lin-Path if controller runs on Linux)
+# Hast to include the suffix directory delimiter (Win = \\, Lin = /)
+__checkmk_agent_host_tmp_dir: "/tmp/"
```
robin-checkmk: There is a newline missing at the end of this file. YAML syntax requires a file to always end with an empty line.

roles/agent/defaults/main.yml
```diff
@@ -49,7 +49,7 @@ checkmk_agent_configure_firewall: true
 
 ## Delegation
 checkmk_agent_delegate_api_calls: 'localhost'
-checkmk_agent_delegate_download: "{{ inventory_hostname }}"
```
robin-checkmk, line 52: This is a breaking change. Can we somehow avoid it?
Coming back to this after reviewing further: I think we can avoid this, as the conditionals you introduced further down are unnecessary, if you keep this filed reserved for the hostname to delegate to.

roles/agent/tasks/Linux-files.yml
```diff
@@ -1,123 +1,245 @@
...
-    checkmk_agent_edition | lower != "community"
-  become: false
-  register: __checkmk_agent_host_download_state
-  # This task may fail, as we fall back to the vanilla agent later
-  failed_when: false
-  changed_when: __checkmk_agent_host_download_state.status is defined and __checkmk_agent_host_download_state.status == 200
-  delegate_to: "{{ checkmk_agent_delegate_download }}"
-  tags:
-    - download-package
+- name: Copy block for Linux target and Linux controller
```
robin-checkmk: Please follow the naming pattern of existing tasks: `"{{ ansible_facts['system'] }}: `
Also may I suggest a more telling name like `"{{ ansible_facts['system'] }}: File management for Linux tagets`

roles/agent/tasks/Linux-files.yml
```diff
@@ -1,123 +1,245 @@
...
+        dest: "{{ __checkmk_agent_agent.file.upload.community }}"
+        mode: "0644"
+      when: |
+        checkmk_agent_delegate_download != inventory_hostname
+        and (checkmk_agent_edition | lower == "community"
+        or not (checkmk_agent_host_specific | bool or checkmk_agent_folder_specific | bool))
+      tags:
+        - download-package
+
+- name: Copy block for Windows target and Linux controller
```
robin-checkmk: See my comment above on the Linux block.

robin-checkmk:
@aewtemp the name/email you made your commits with, does not match the one in your GitHub account. Can you either add the name or make the commit with a name, that matches yours? Worst case you need to copy the files and create a new commit and PR. Let me know, if I can help with that.

We do appreciate your contribution though! I am currently reviewing, so if you need to re-create the PR, maybe wait for my feedback to land.

roles/agent/tasks/Linux.yml
```diff
@@ -90,7 +90,7 @@
   tags:
     - manage-host
 
-- name: "Ensure registration readiness."  # noqa no-handler
+- name: "Ensure registration readiness." # noqa no-handler
```
robin-checkmk: The double-blank is intended here, please do not remove it.

roles/agent/tasks/Linux.yml
```diff
@@ -268,7 +268,7 @@
   tags:
     - register-agent
 
-- name: "{{ ansible_facts['system'] }}: Initial push of data for push agent."  # noqa no-changed-when
```
robin-checkmk, line 271: The double-blank is intended here, please do not remove it.

roles/agent/tasks/main.yml
```diff
@@ -27,8 +27,30 @@
...
+- name: "{{ ansible_facts['system'] }}: Include agent package download tasks (Linux)."
+  ansible.builtin.include_tasks: "Linux-files.yml"
+  when: not __checkmk_agent_delegate_is_windows
+  tags:
+    - include-system-tasks
+    - download-package
+
+- name: "{{ ansible_facts['system'] }}: Include agent package download tasks (Windows)."
+  ansible.builtin.include_tasks: "Win32NT-files.yml"
+  when: __checkmk_agent_delegate_is_windows
```
robin-checkmk, line 53: Why is this necessary? This feels awefully complex and seems to do the same as the original code.

roles/agent/tasks/Linux-files.yml
```diff
@@ -1,123 +1,245 @@
...
+        mode: "0644"
+        timeout: "{{ checkmk_agent_download_timeout | default(omit) }}"
+      when: |
+        checkmk_agent_edition | lower != "community"
+      become: false
+      register: __checkmk_agent_host_download_state
+      # This task may fail, as we fall back to the vanilla agent later
+      failed_when: false
+      changed_when: __checkmk_agent_host_download_state.status is defined and __checkmk_agent_host_download_state.status == 200
+      delegate_to: "{{ 'localhost' if checkmk_agent_delegate_download else 'inventory_hostname' }}"
```
robin-checkmk: `'inventory_hostname'` is the literal string, but you intend to use the variable here, right? Search for the string, this happened in several places.

robin-checkmk (changes requested):
I like the idea of splitting the file path into two sets:

>     __checkmk_agent_agent.file.download.*: where the package is downloaded to (on the delegated/download host).
>     __checkmk_agent_agent.file.upload.*: where the package is copied/uploaded to on the target host, and subsequently installed from.

What I have doubts and concerns about:

>     checkmk_agent_delegate_download is changed from a hostname-based variable (defaulting to inventory_hostname) to a boolean, making the delegation intent explicit rather than implicit through hostname comparison.

This is a breaking change, which will affect every single user, who uses the delegation to this day. And while I understand the idea, I would ask you to revert to the previous behavior. Also consider my more detailed comments on this matter.

>     __checkmk_agent_host_tmp_dir (used for the download destination) is now defined centrally in vars/main.yml, while OS-specific __checkmk_agent_lin_tmp_dir / __checkmk_agent_win_tmp_dir variables are introduced for the upload destination on Linux and Windows targets respectively.

I do not see merrit in this change, why is it necessary?

roles/agent/defaults/main.yml
aewtemp: The problem here is, that this configuration option leaves the impression one can randomly choose a delegate to have the files downloaded and then copied from to the target host.

This violates ansibles own assumption for the modules `ansible.builtin.copy` and `ansible.windows.win_copy`, whose expectations are that the file to copy has to reside either on the controller or on the target host.
One can only download files directly to the target host, or the controller, to then be able to copy the files to their destination.

So I don't think it would be good to leave the logic behind the delegation as is.

roles/agent/tasks/Linux-files.yml
aewtemp: Really sorry for that.
My bad.

roles/agent/vars/main.yml
aewtemp: Will be fixed with the next commit.

roles/agent/tasks/Linux-files.yml
aewtemp: Will be fixed with the next commit.

roles/agent/tasks/Linux.yml
aewtemp: Will be fixed with the next commit.

roles/agent/tasks/Linux-files.yml
aewtemp: Will be fixed with the next commit.

roles/agent/tasks/Linux.yml
aewtemp: Will be fixed with the next commit.

roles/agent/defaults/main.yml
robin-checkmk: I think I see your point and I remember considering the option to actually delegate things to a third party host, so it can act as an intermediary.  But I am still very hesitant to introduce such a breaking change. Especially, as the issue you describe has not surfaced yet. I would rather extend on documentation, instead of an unnecessary breaking change.

roles/agent/tasks/main.yml
aewtemp: I just came to notice that I left the original inclusion of `os_family` specific variables above this block.
I'll check if it works without explicitly loading the needed variables for the controller host.

roles/agent/defaults/main.yml
aewtemp: I fully understand your point.
But hear me out: Just because nobody has encountered the error yet, doesn't mean it's not there. And I am the first person then to have hit that problem.
I am targeting 1700+ hosts with this Playbook. Windows and Linux hosts. I tried delegating the operations for Windows hosts to a third (not controller nor target) Windows machine. It fails.

This is my approach to fix this misconception and the resulting error.

To me this wont break any setup already using this collection in their setups. Because they have not had this error yet. 

There is no way this would be working if you choose a third host as delegate for the download and copy operation. Try it yourself.
Besides that Problem, if the controller is a Linux host, you only will be able to transfer files to Linux hosts. The current state does not work when cross-transferring files from Linux to Windows and vice versa.
This will also be fixed with my approach.

I'll fix and commit the outstanding things.

I would appreciate it if you could reconsider this once more.

aewtemp:
I did not find the time to check the `os_family` specific imports, but it's noted on my ToDos.
Another commit will be following once I have verified that.

robin-checkmk:
Please also address the CLA Assistant issue, as without a proper signature I cannot merge this.

aewtemp:

roles/agent/tasks/main.yml
aewtemp: Checked and streamlined.

roles/agent/defaults/main.yml
robin-checkmk: > To me this wont break any setup already using this collection in their setups. Because they have not had this error yet.

That is not the point. The point is: If this change goes through, it is a breaking change, as the default value goes from `boolean` to `string`, meaning every user must update their configuration. Otherwise they will hit this error revealed by `molecule`:

```yaml
fatal: [debian12 -> localhost]: FAILED! => {"changed": false, "msg": "Task failed: Conditional result (True) was derived from value of type 'str' at '/home/myuser/.ansible/tmp/molecule.1PC8.default/inventory/ansible_inventory.yml:6:5'. Conditionals must have a boolean result."}
```

Maybe we can compromise? Remove the breaking change for this PR and we can revisit the redesign of the mechanism in a dedicated PR, which is not fixing a bug but introducing a new feature? The same ask would be for the other "Additional" topic from the initial PR description, wich changes the temporary directory variables. Or is that change absolutetly necessary for the initial fix to work?

roles/agent/tasks/Win32NT-files.yml
```diff
@@ -1,118 +1,248 @@
...
+      ansible.windows.win_copy:
+        src: "{{ __checkmk_agent_agent.file.download.community }}"
+        dest: "{{ __checkmk_agent_agent.file.upload.community }}"
+        remote_src: "{{ not checkmk_agent_delegate_download }}"
+      when: |
+        checkmk_agent_delegate_download != inventory_hostname
+        and (checkmk_agent_edition | lower == "community"
+        or not (checkmk_agent_host_specific | bool or checkmk_agent_folder_specific | bool))
+      tags:
+        - download-package
```
robin-checkmk: I might be reading this wrong, but isn't this the wrong way around? Using the Linux modules on a Windows target?

roles/agent/tasks/Linux-files.yml
```diff
@@ -1,123 +1,248 @@
...
+    - name: "{{ ansible_facts['system'] }}: Transfer Vanilla {{ checkmk_agent_edition | upper }} Agent."
+      ansible.windows.win_copy:
+        src: "{{ __checkmk_agent_agent.file.download.community }}"
+        dest: "{{ __checkmk_agent_agent.file.upload.community }}"
+      when: |
+        checkmk_agent_delegate_download != inventory_hostname
+        and (checkmk_agent_edition | lower == "community"
+        or not (checkmk_agent_host_specific | bool or checkmk_agent_folder_specific | bool))
+      tags:
+        - download-package
```
robin-checkmk: I might be reading this wrong, but isn't this the wrong way around? Using the Windows modules on a Linux target?

roles/agent/defaults/main.yml
robin-checkmk: P.S.: This breaking change now also renders all occurences of `checkmk_agent_delegate_download != inventory_hostname` to always be true.

roles/agent/tasks/main.yml
```diff
@@ -20,6 +20,31 @@
   tags:
     - always
 
+- name: "Preflight - Make sure facts for the Ansible controller are present."
+  ansible.builtin.setup:
+  when: checkmk_agent_delegate_download
+  delegate_to: localhost
+  delegate_facts: true
+  tags:
+    - download-package
```
robin-checkmk: I do not think the role should handle fact gathering explicitely. The user should be able to decide themselfes if and how they gather facts. This might actually break elaborate use cases.

roles/agent/tasks/main.yml
```diff
@@ -20,6 +20,31 @@
...
+        (checkmk_agent_delegate_download and hostvars['localhost']['ansible_facts']['os_family'] == 'Windows') or
+        (not checkmk_agent_delegate_download and ansible_facts['os_family'] == 'Windows')
+      }}
+  tags:
+    - download-package
+
+- name: "Preflight - Set facts for handling of the agent download"
+  ansible.builtin.set_fact:
+    __checkmk_agent_host_tmp_dir: >-
+      {{ '/tmp/' if not __checkmk_agent_download_to_windows else 'C:\\Windows\\Temp\\' }}
```
robin-checkmk: This conditional is hard to understand, what are you trying to achieve here?

roles/agent/vars/RedHat.yml
```diff
@@ -1,11 +1,16 @@
...
     community: "{{ __checkmk_agent_site_url }}/check_mk/agents/check-mk-agent-{{ checkmk_agent_version }}-1.noarch.rpm"
     pro: "{{ __checkmk_agent_site_url }}/check_mk/api/1.0/domain-types/agent/actions/download_by_host/invoke"
   file:
-    community: "{{ __checkmk_agent_host_tmp_dir }}/check-mk-agent-{{ checkmk_agent_version }}-1.noarch-vanilla.rpm"
-    pro: "{{ __checkmk_agent_host_tmp_dir }}/check-mk-agent-{{ checkmk_agent_version }}-1.noarch-generic.rpm"
-    host: "{{ __checkmk_agent_host_tmp_dir }}/check-mk-agent-{{ checkmk_agent_version }}-1.noarch-{{ inventory_hostname }}.rpm"
+    download:
+      community: "{{ __checkmk_agent_host_tmp_dir }}check-mk-agent-{{ checkmk_agent_version }}-1.noarch-vanilla.rpm"
+      pro: "{{ __checkmk_agent_host_tmp_dir }}check-mk-agent-{{ checkmk_agent_version }}-1.noarch-generic.rpm"
+      host: "{{ __checkmk_agent_host_tmp_dir }}check-mk-agent-{{ checkmk_agent_version }}-1.noarch-{{ inventory_hostname }}.rpm"
```
robin-checkmk: This should use `__checkmk_agent_lin_tmp_dir` as well, right?

roles/agent/vars/Debian.yml
```diff
@@ -1,11 +1,16 @@
 ---
-__checkmk_agent_host_tmp_dir: "/tmp"
```
robin-checkmk, line 2: In the PR description you said:

> __checkmk_agent_host_tmp_dir (used for the download destination) is now defined centrally in vars/main.yml, while OS-specific __checkmk_agent_lin_tmp_dir / __checkmk_agent_win_tmp_dir variables are introduced for the upload destination on Linux and Windows targets respectively.

But the variable is missing in `vars/main.yml` which would leave us with an undefined variable.

robin-checkmk (changes requested):
@aewtemp I am really struggling with this PR. On the one hand I am very happy and thankful, that you tackle this issue. On the other hand, the changes here are massive and I am afraid to break more that we fix. So while I do not want to frustrate you, I still have to stay diligent and I hope you can bear with me.

One thing I realized, but did not say earlier is, that you did not touch the molecule tests. I assume you are not able to run them locally, or you just never worked with them. I can support with that, but might be worth looking into to understand, if these changes break anything.

And lastly: You probably want to rebase your PR on the current `devel` branch, to have an up-to-date basis and to fix the conflict mentioned by GitHub.

roles/agent/tasks/Win32NT-files.yml
aewtemp: The module to be used has to match the operating environment of the machine (controller or target) doing the transfer.
So:
|OS|Module|
|:---|:---:|
|Linux|`ansible.builtin.copy`|
|Windows|`ansible.windows.win_copy`|

That means, if we delegate downloading and copying to a controller operating on a Linux based operating system, we have to use Linux compatible modules to successfully execute the file transfer after downloading and vice versa (windows).

roles/agent/tasks/Linux-files.yml
aewtemp: Please see above comment.

roles/agent/tasks/main.yml
aewtemp: :open_mouth:  The first catch is a good one. Doesn't break the logic behind the commited code but still is a nice blunder of mine.
Will be fixed with the next commit.

The second one is needed, as far as I read. Ansible doesn't automatically gather facts for the controller. Further, as far as I understood, are Modules executed on the controller still done in the target_hosts context. Hence the need to gather facts about the controller to have the correct environment and variables. This is crucial to build the correct file paths, for example.

I will look into having the facts to be gathered reduced.

roles/agent/tasks/main.yml
aewtemp: I am happy to be given the chance :smile: 

To have the `os_family` based variables change as little as possible and to not break their dependency to the operating system family type, this logic sets `__checkmk_agent_host_tmp_dir` to either a Linux based prefix or a Windows based prefix.

This leaves the vars blocks under `vars/Windows.yml` and `vars/{Debian,RedHat,Suse}.yml` relatively unchanged. I just had to remove the folder delimiter and move it to the variable `__checkmk_agent_host_tmp_dir`.

In order to set this variable, the task "Preflight - Set facts for the downloading Host" above "Preflight - Set facts for handling of the agent download" defines if the host downloading the agent is windows or linux based.

roles/agent/vars/RedHat.yml
aewtemp: Another blunder...
But not those lines (10-12) but those below (14-16)

roles/agent/vars/Debian.yml
aewtemp: As I iterated over my edits I came to the conclusion that there is no need to manually define this variable anymore.

aewtemp:
I'll take a look at the molecule stuff later.

pyproject.toml
```diff
@@ -14,7 +14,8 @@ dependencies = [
 
 [dependency-groups]
 dev = [
-    "black==26.5.1",
+    "black==26.3.1",
+    "click==8.3.1",
```
robin-checkmk: This is obviously not an intended change, please revert it.

roles/agent/vars/Debian.yml
robin-checkmk: So you will revert to using `__checkmk_agent_host_tmp_dir` or what are you saying?

roles/agent/tasks/Linux-files.yml
```diff
@@ -1,123 +1,242 @@
...
+        dest: "{{ __checkmk_agent_agent.file.upload.community }}"
+        mode: "0644"
+        remote_src: "{{ not checkmk_agent_delegate_download }}"
+      when: |
+        (checkmk_agent_edition | lower == "community"
+        or not (checkmk_agent_host_specific | bool or checkmk_agent_folder_specific | bool))
+      tags:
+        - download-package
+
+- name: "{{ ansible_facts['system'] }}: File handling for Linux target and Windows controller"
```
robin-checkmk: This is impossible, Ansible does not support Windows.

roles/agent/tasks/main.yml
robin-checkmk: But as Ansible cannot use Windows as a control node, your changes to the download variables should already be enough to populate the necessary path variables.

roles/agent/tasks/main.yml
robin-checkmk: See my comment about this on the code below: I think this becomes irrelevant, once you consider, that Windows cannot be an Ansible control node.

roles/agent/tasks/Linux-files.yml
robin-checkmk: I just relalized, that the "File handling for Linux target and Windows controller" block can go completely, as Ansible does not support Windows as a controller.

roles/agent/tasks/Win32NT-files.yml
robin-checkmk: The module has to match the target, not the controller.
But the more pressing matter here is, that the whole "File handling for Windows target and Windows controller" block can go, because Ansible does not support Windows as a controller.

robin-checkmk (changes requested):
I just reviewed the current state, but I sense there are still changes pending on your end. Please also revisit the comments about the breaking change, as that will be the decisive factor for this PR. I will commit to merging this, if we can focus on the path-split change without a breaking change.

pyproject.toml
aewtemp: This came from rebasing on your dev-branch.

Your "dependabot" made this change.
If you want me to, I can revert it.

aewtemp:
I'll rework the process based on what I've learned. I can not tell where I got the impression windows could be used as a controller.

Besides, the workflow must still respect, based on the target (windows/linux) where the file should be downloaded to.
The breaking part then was, that if you delegate the downloading to the controller, variables for the temporary download directory would be windows based.
I am on it.

Thanks for your time and patience.

aewtemp:
Shall I rebase on the main or the dev branch?

robin-checkmk:
> Shall I rebase on the main or the dev branch?

Use `devel`. I think you got really unlucky with the last rebase, as you got the faulty version bump, which was only there for some minutes. But generally `devel` should be the base branch for contributions.

pyproject.toml
robin-checkmk: The new rebase on `devel` should do it.

PR #1176, opened 2026-08-25
Unify exit_module and RETURN block.

robin-checkmk:

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

- All modules - `http_code`, `content` and `etag` are now returned by every module that talks to the REST API. Existing return values are unchanged, the keys are only added.
- All modules - `result_as_dict()` was removed from `module_utils.utils`. It returned a subset of the result and was what dropped the keys above. Every module exits through `exit_module()` now. This only affects code importing the helper directly from `module_utils`, not playbooks or roles.


lgetwan (approved):
I can hardly understand how this unit test works, but the actual changes in the modules look good to me!

lgetwan (approved):
I foxed myself into the unit tests, and they look good to me!

